PR #7, opened 2020-07-02
add link to all referenced KIPs

ickymettle:
Adds links back to the KIP documents listed in the Supported KIPs table.


twmb:
Thanks!

PR #11, opened 2020-10-05
Code examples demonstrating the connection options

weeco:


examples/connect/sasl_ssl_plain.go
```diff
@@ -0,0 +1,49 @@
...
+				User: user,
+				Pass: password,
+			}, nil
+		})),
+
+		// Configure TLS. Uses SystemCertPool for RootCAs by default.
+		kgo.Dialer(func(ctx context.Context, host string) (net.Conn, error) {
+			return (tls.Dialer{
+				NetDialer: &net.Dialer{Timeout: 10 * time.Second},
+			}).NetDialer.DialContext(ctx, "tcp", host)
```
twmb, line 37: This line should be `}).DialContext(ctx, "tcp", host)`, so as to use the tls dialer, not the inner standard dialer.

twmb:
Minor nit but can the directory be renamed to `connecting`? I'd like to avoid any potential confusion with Kafka Connect.

weeco:
Makes sense, haven't even thought about Kafka connect. Addresses both feedbacks 👍

examples/connecting/sasl_ssl_plain.go
```diff
@@ -0,0 +1,49 @@
...
+		kgo.SASL(plain.Plain(func(context.Context) (plain.Auth, error) {
+			return plain.Auth{
+				User: user,
+				Pass: password,
+			}, nil
+		})),
+
+		// Configure TLS. Uses SystemCertPool for RootCAs by default.
+		kgo.Dialer(func(ctx context.Context, host string) (net.Conn, error) {
+			return (tls.Dialer{
```
twmb, line 35: Can this have an `&` before the tls.Dialer? `DialContext` requires a pointer receiver, so without the `&` this fails to compile if one just blindly copies and pastes this.

twmb: Alternatively, this dialer could be defined above the options; as it currently exists, this does allocate a tls.Dialer and net.Dialer every dial. I think this is showing me that it's worth it to have a direct `kgo.TLSDialer` option, but I'm not sure what the args for that should be (maybe a tls.Config? maybe also a net.Dialer?)

twmb: I've pushed a breaking API change for the Dialer, this new change should make this API slightly easier to use, and mirrors other places in the go stdlib that take dialers.

examples/connecting/sasl_ssl_plain.go
```diff
@@ -0,0 +1,50 @@
...
+		// SASL Options
+		kgo.SASL(plain.Plain(func(context.Context) (plain.Auth, error) {
+			return plain.Auth{
+				User: user,
+				Pass: password,
+			}, nil
+		})),
+
+		// Configure TLS. Uses SystemCertPool for RootCAs by default.
+		dialer := &tls.Dialer{
```
twmb, line 34: This dialer needs to be above the `kgo.Opt[]`, but alternatively, you could to
```go
kgo.Dialer((&tls.Dialer{
  NetDialer: &net.Dialer{Timeout: 10*time.Second},
}).DialContext)
```

If you instead move the dialer above `kgo.Opt`, this can be
```go
kgo.Dialer(dialer.DialContext)
```

examples/connecting/sasl_ssl_plain.go
```diff
@@ -0,0 +1,46 @@
...
+	opts := []kgo.Opt{
+		kgo.SeedBrokers(seeds...),
+
+		// SASL Options
+		kgo.SASL(plain.Plain(func(context.Context) (plain.Auth, error) {
+			return plain.Auth{
+				User: user,
+				Pass: password,
+			}, nil
+		})),
```
twmb, line 32:
```suggestion
		kgo.SASL(plain.Auth{
			User: user,
			Pass: password,
		}.AsMechanism()),
```

twmb (approved):
Thanks!

PR #14, opened 2020-11-15
Redesign readme

weeco:
This PR aims to redesign the README, which is obviously highly opinionated. My main motivation is to only put the most important parts into the README that you expect your first time users to read. In my opinion the previous read me proivided to much detailed information while it was lacking a getting started section. 

Thus I moved lots of the information into their separate documents and linked them instead. On top of that I added some badges.


README.md
```diff
@@ -2,48 +2,59 @@ franz-go
...
 
-kgo has support for transactions, regex topic consuming, the latest partitioning
-strategies, data loss detection, closest replica fetching, and more. If a client
-[KIP][1] exists, this library aims to support it.
+ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+defer cancel()
 
-[1]: https://cwiki.apache.org/confluence/display/KAFKA/Kafka+Improvement+Proposals
+// Construct message request and send it to Kafka
+req := kmsg.MetadataRequest{
```
twmb, line 33: Can this use `kmsg.NewMetadataRequest`? And for the append below, `kmsg.NewMetadataRequestTopic()`?

README.md
```diff
@@ -2,48 +2,58 @@ franz-go
...
+## Getting started
 
-franz-go is an all-encompassing Kafka consumer api in Go. The main package is kgo,
-but kmsg also exists containing autogenerated code for every request and
-response type, and a few other miscellaneous packages exist.
+Basic usage looks like this:
 
-The main package kgo is a high performance, pure Go library for interacting
-with Kafka. This library aims to provide every Kafka feature from 0.8.0+.
+```go
```
twmb, line 21: I think the most common / most expected usage of this package would be for producing or consuming, so it may be better to have one message produced here (maybe the callback can use a wait group just to show), and then consuming that same message.

Maybe something like

```go
ctx := context.Background()

// All record production goes through Produce, and the callback can be used
// to allow for syncronous or asyncronous production.
var wg sync.WaitGroup
wg.Add(1)
record := &kgo.Record{Topic: "foo", Value: []byte("bar")}
err := client.Produce(ctx, record, func(_ *Record, err error) {
        defer wg.Done()
        if err != nil {
                fmt.Printf("record had a produce error: %v\n", err)
        }

}
if err != nil {
        panic("we are unable to produce if the context is canceled, we have hit max buffered," +
                "or if we are transactional and not in a transaction")
}
wg.Wait()

// Consuming can either be direct (no group), or through a group.
// Below, we use a group.
client.AssignGroup("group",
        kgo.GroupTopics("foo"))

for {
        fetches := client.PollFetches(ctx)
        // process fetches
}
```

PR #18, opened 2020-12-05
Add BrokerThrottleHook

akesle:
Innocuous change that has been smoke-tested with producer and consumer client instances running against a `5.5.2` cluster with default producer and consumer client throttling configs for over 1h.


twmb:
I changed how throttling can be accessed, removing the need for reflect and removing the need for the user to determine whether the throttle was applied before or after. I've taken this commit and modified it a bit in https://github.com/twmb/franz-go/commit/303186a3a720c45a2e71333a6310b06ff4aa097d. This removes the version from being passed to the hook, but adds whether the throttle was already applied (which is what the version would likely be used for). Please reopen if this is not sufficient, thanks!

PR #22, opened 2021-02-01
Add examples for transactional producer and consumer

dcrodman:
Hi @twmb! I'm implementing a new producer with this library since it seems to have great support for producing messages in transactions and I need exactly-once semantics. I ended up crafting a couple of examples to test things out and figured I might put them out there for the sake of consolidating some of the transaction documentation.

I know the other examples don't use a main package but I found it convenient to be able to run the example as-is so I left it in there. More than happy to make changes to that or any other part of this.

As a side note, if I'm doing anything wrong here I'd really appreciate any feedback or guidance here.

Example output (repeats every 10s):
```
produced message: batch 0 record 0
produced message: batch 0 record 1
produced message: batch 0 record 2
produced message: batch 0 record 3
produced message: batch 0 record 4
produced message: batch 0 record 5
produced message: batch 0 record 6
produced message: batch 0 record 7
produced message: batch 0 record 8
produced message: batch 0 record 9
consumed record with message: batch 0 record 0
consumed record with message: batch 0 record 1
consumed record with message: batch 0 record 2
consumed record with message: batch 0 record 3
consumed record with message: batch 0 record 4
consumed record with message: batch 0 record 5
consumed record with message: batch 0 record 6
consumed record with message: batch 0 record 7
consumed record with message: batch 0 record 8
consumed record with message: batch 0 record 9
```

### DIFF
--- a/examples/transactions/transactional_consumer.go
+++ b/examples/transactions/transactional_consumer.go
@@ -1,0 +1,47 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/twmb/franz-go/pkg/kgo"
+)
+
+func startConsuming(ctx context.Context, kafkaBrokers string, topic string) {
+	client, err := kgo.NewClient(
+		kgo.SeedBrokers(strings.Split(kafkaBrokers, ",")...),
+		// Only read messages that have been written as part of committed transactions.
+		kgo.FetchIsolationLevel(kgo.ReadCommitted()),
+	)
+	if err != nil {
+		fmt.Printf("error initializing Kafka consumer: %v\n", err)
+		return
+	}
+
+	client.AssignGroupTransactSession(
+		"my-consumer-group",
+		kgo.GroupTopics(topic),
+	)
+
+consumerLoop:
+	for {
+		fetches := client.PollFetches(ctx)
+		iter := fetches.RecordIter()
+
+		for _, fetchErr := range fetches.Errors() {
+			fmt.Printf(
+				"error consuming from topic: topic=%s, partition=%d, err=%v",
+				fetchErr.Topic, fetchErr.Partition, fetchErr.Err,
+			)
+			break consumerLoop
+		}
+
+		for !iter.Done() {
+			record := iter.Next()
+			fmt.Printf("consumed record with message: %v", string(record.Value))
+		}
+	}
+
+	fmt.Println("consumer exited")
+}
--- a/examples/transactions/transactional_consumer.go
+++ b/examples/transactions/transactional_consumer.go
@@ -20,8 +20,7 @@ func startConsuming(ctx context.Context, kafkaBrokers string, topic string) {
 	}
 
 	client.AssignGroup("my-consumer-group", kgo.GroupTopics(topic))
-	// Leave the consumer group when the consumer exits.
-	defer client.AssignGroup("")
+	defer client.Close()
 
 consumerLoop:
 	for {
--- a/examples/transactions/transactional_consumer.go
+++ b/examples/transactions/transactional_consumer.go
@@ -19,10 +19,9 @@ func startConsuming(ctx context.Context, kafkaBrokers string, topic string) {
 		return
 	}
 
-	client.AssignGroupTransactSession(
-		"my-consumer-group",
-		kgo.GroupTopics(topic),
-	)
+	client.AssignGroup("my-consumer-group", kgo.GroupTopics(topic))
+	// Leave the consumer group when the consumer exits.
+	defer client.AssignGroup("")
 
 consumerLoop:
 	for {
--- a/examples/transactions/transactional_producer.go
+++ b/examples/transactions/transactional_producer.go
@@ -67,6 +67,9 @@ func startProducing(ctx context.Context, kafkaBrokers string, topic string) {
 func produceRecords(ctx context.Context, client *kgo.Client, topic string, batch int) error {
 	errChan := make(chan error)
 
+	// Records are produced sequentially in order to demonstrate that a consumer
+	// using the ReadCommitted isolation level will not consume any records until
+	// the transaction is committed.
 	for i := 0; i < 10; i++ {
 		message := fmt.Sprintf("batch %d record %d\n", batch, i)
 		r := &kgo.Record{
--- a/examples/transactions/transactional_producer.go
+++ b/examples/transactions/transactional_producer.go
@@ -1,0 +1,103 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/twmb/franz-go/pkg/kgo"
+)
+
+func startProducing(ctx context.Context, kafkaBrokers string, topic string) {
+	producerId := strconv.FormatInt(int64(os.Getpid()), 10)
+	client, err := kgo.NewClient(
+		kgo.SeedBrokers(strings.Split(kafkaBrokers, ",")...),
+		kgo.TransactionalID(producerId),
+	)
+	if err != nil {
+		fmt.Printf("error initializing Kafka producer client: %v", err)
+		return
+	}
+
+	defer client.Close()
+
+	batch := 0
+	for {
+		if err := client.BeginTransaction(); err != nil {
+			fmt.Printf("error beginning transaction: %v\n", err)
+			break
+		}
+
+		// Write some messages in the transaction.
+		if err := produceRecords(ctx, client, topic, batch); err != nil {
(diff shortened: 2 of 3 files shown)


examples/transactions/transactional_consumer.go
```diff
@@ -0,0 +1,47 @@
...
+		kgo.SeedBrokers(strings.Split(kafkaBrokers, ",")...),
+		// Only read messages that have been written as part of committed transactions.
+		kgo.FetchIsolationLevel(kgo.ReadCommitted()),
+	)
+	if err != nil {
+		fmt.Printf("error initializing Kafka consumer: %v\n", err)
+		return
+	}
+
+	client.AssignGroupTransactSession(
```
twmb, line 22: A group transaction session is generally meant for a consumer-modify-produce type of loop, but the below code does not produce back its modifications. When assigning a transact session, you'd definitely want to use the `GroupTransactSession` return to ensure that things are done properly (commit vs. abort, stuff like that).

In this case, you could just AssignGroup.

examples/transactions/transactional_producer.go
```diff
@@ -0,0 +1,103 @@
...
+
+		err := client.Produce(ctx, r, func(r *kgo.Record, e error) {
+			fmt.Printf("produced message: %s", message)
+			errChan <- e
+		})
+		if err != nil {
+			return err
+		}
+
+		select {
```
twmb, line 85: This select forces the loop to wait until a `ProduceRequest` is actually issued, meaning each produce is sequential and blocking. What you likely want to do is to produce all 10 records as fast as possible.

In this mini example, you may want to pass in the `errChan` to `produceRecords` and in `startProducing`, have a goroutine draining `errChan` and reacting to what it receives.

examples/transactions/transactional_consumer.go
twmb: (if you do want to use a GroupTransactSession, an example is here: https://github.com/twmb/kcl/blob/master/commands/transact/transact.go#L281-L307)

examples/transactions/transactional_consumer.go
```diff
@@ -0,0 +1,47 @@
...
+		"my-consumer-group",
+		kgo.GroupTopics(topic),
+	)
```
twmb, line 25: You likely want to `defer client.AssignGroup("")` to leave the group, rather than shutting down without leaving (which leaves the responsibility to Kafka to eventually notice the heartbeat interval has lapsed and the group member is missing).

examples/transactions/transactional_producer.go
```diff
@@ -0,0 +1,103 @@
...
+			return ctx.Err()
+		}
+	}
+
+	return nil
+}
+
+func rollback(ctx context.Context, client *kgo.Client) {
+	if err := client.EndTransaction(ctx, kgo.TryAbort); err != nil {
+		fmt.Printf("error rolling back transaction: %v\n", err)
```
twmb, line 100: probably a return under here so as to not fall into the successful println below

examples/transactions/transactional_producer.go
```diff
@@ -0,0 +1,103 @@
...
+
+		// Flush all of the buffered messages.
+		if err := client.Flush(ctx); err != nil {
+			if err != context.Canceled {
+				fmt.Printf("error flushing messages: %v\n", err)
+			}
+
+			rollback(ctx, client)
+			continue
+		}
```
twmb, line 49: Regarding that errChan comment I made below, above produceRecords you could create the goroutine and drain it until you close the channel; if any errors are received, set an abort bool to rollback.

Then, after the Flush, check the abort bool and, if necessary, rollback here before EndTransaction

twmb:
Hey there, thanks for the example! I've added a few small comments to clarify some underlying behaviors, but otherwise this generally looks good!

examples/transactions/transactional_producer.go
twmb: Or is this meant to demonstrate that the consumer doesn't consume these messages until the transaction is done?

dcrodman: That's exactly what I had in mind, I was trying to demonstrate that "batches" of messages are committed at once and the consumer will not read them until the batch is ready. However I'm happy to change this if you think it's clearer

examples/transactions/transactional_consumer.go
dcrodman: As one potential suggestion/point of feedback, would it be worth considering adding a helper method like `client.LeaveGroup()` that does this call with the empty string?

examples/transactions/transactional_producer.go
dcrodman: Am I right in thinking that it's actually unnecessary to check `ctx.Done` explicitly since `Produce` will return the context cancellation error if that occurs?

examples/transactions/transactional_consumer.go
twmb: With Sarama, you have to call `Close` on the `ConsumerGroup` interface you have. I'm open to adding a `LeaveGroup()`. I'm not 100% satisfied with the current commit offsets functions, so there's definitely room for improvement in the broader API as a whole (so if you have even more ideas I'm open to them). Unfortunately this would likely require a breaking change in the future if a much better API is found (before stabilizing on 1.0.0)

examples/transactions/transactional_producer.go
twmb: `Produce` itself only returns an error in a very few select cases, so if you know what you're doing (not trying to produce when not in a transaction, not buffering max records and then canceling the context), you don't really _need_ to inspect the error.

But yes, you're right that it's not necessary to check `ctx.Done` here. The context is only used if the client is currently at the max buffered records, if it is, it waits for some records to finish or for the context to be canceled.

twmb: > However I'm happy to change this if you think it's clearer
I think that since the point is to show the consumer doesn't consume when using ReadCommitted until the transaction is done, this is good. Maybe add some preceding documentation to explain that purpose?
```go
// We wait sequentially for every record to be produced,
// demonstrating that the consumer does not consume
// with ReadCommitted until the transaction is ended.
```
or something

examples/transactions/transactional_consumer.go
twmb: Actually, I've remembered that `client.Close()` leaves the group as well (unless you are using instance IDs). Do you think an explicit `LeaveGroup` would still be beneficial?

dcrodman: Yep, you're right - from `client.go`:
```
// Close leaves any group and closes all connections and goroutines.
func (cl *Client) Close() {
```
It would be nice to have if the expected way to leave a group was the explicit `client.AssignGroup("")` but given the `Close()` behavior I don't think the `client.LeaveGroup()` is necessary

twmb: I'm open to adding an explicit function still, it may be helpful.

dcrodman: If you don't mind making the API change then I certainly won't argue with adding it. Even though it's _technically_ a little redundant its usage would be clearer and might highlight the need to do this, whether it's via `client.Close()` or the explicit call to `LeaveGroup()` (or whatever)

examples/transactions/transactional_producer.go
```diff
@@ -0,0 +1,107 @@
...
+		err := client.Produce(ctx, r, func(r *kgo.Record, e error) {
+			fmt.Printf("produced message: %s", message)
+			errChan <- e
+		})
+		if err != nil {
+			return err
+		}
+
+		select {
+		case err := <-errChan:
```
twmb, line 89: Since you pointed out the select on `ctx.Done` here is unnecessary, maybe this should be:
```go
if err := <-errChan; err != nil {
    return err
}
```
?

dcrodman: You're totally right, let me clean that up

examples/transactions/transactional_consumer.go
twmb: FWIW I've added `LeaveGroup`, it's on the latest untagged commits -- but I think your update to use `client.Close` is great here.

twmb:
Looks great, thank you!

dcrodman:
Thanks for the suggestions and feedback!

examples/transactions/transactional_consumer.go
twmb: (forgot to push, now it's pushed -- I plan to tag tomorrow or the next day)

twmb: I've cut v0.6.4 after running race tests for a bit -- turns out that there were a few small bugs, two that were tagged, one not yet tagged.

PR #35, opened 2021-05-06
Improve formatting and fix a typo

kalmant:
A small improvement to the formatting of the docs README.md.
I believe this is the desired layout based on the original markdown.


twmb:
Hi there, thanks!

This is my original desire, actually, so this is great! I forgot that you could just put html links into pre tags; I originally just tried triple ticks for code blocks but I couldn't get links in them.

This looks good here. I'm working on some other doc changes locally and I'll copy what you did back into the top-level README (where this is duplicated).

PR #43, opened 2021-06-24
bench: fix double goroutine invoke for printRate()

Neal:
`printRate()` is already called with `go printRate()` so remove another `go func() {}()` inside `printRate()` methods.


twmb:
Ah good catch, thank you!

PR #44, opened 2021-06-24
CHANGELOG: fix commit links

Neal:


twmb (approved):
Thank you for this as well! Not sure how I missed it in v0.8.2, but I likely c&p'd that to v0.8.3 and v0.8.4.

PR #47, opened 2021-07-05
kprom: use Registerer and Gatherer interface instead of concreate Registry type

simon0191:
Use Registerer and Gatherer interface instead of concreate Registry type

This allows the user of the plugin to:
- pass prometheus.DefaultRegisterer and prometheus.Defaultgatherer
- pass a testing a custom Registerer implementation (e.g. for testing purposes)

The change is backwards compatible.

### DIFF
--- a/plugin/kprom/kprom.go
+++ b/plugin/kprom/kprom.go
@@ -70,20 +70,26 @@ type Metrics struct {
 //
 // This is useful if you want the Metrics type to create its own registry for
 // you to add additional metrics to.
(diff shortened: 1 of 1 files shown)


twmb (approved):
Looks good! The only mildly breaking bit is if somebody stored the return from `Registry` into a `*prometheus.Registry` type directly, but I doubt that is common, and this package is still early enough to make these types of changes.

twmb:
I'll tag this as plugin/kprom/v0.2.0

twmb:
Tagged

PR #65, opened 2021-08-22
when CommitRecords commit next offset to move forward

vtolstov:
Signed-off-by: Vasiliy Tolstov <v.tolstov@unistack.org>


twmb:
I can't believe I missed this, great fix, thank you!

PR #94, opened 2021-10-12
Fix createTopics admin method

turkenh:
Admin client method `createTopics` always sets `req.ValidateOnly` as `true` which causes Topics never created.

This PR fixes the issue by using already existing `dry` parameter.

### DIFF
--- a/pkg/kadm/topics.go
+++ b/pkg/kadm/topics.go
@@ -75,7 +75,7 @@ func (cl *Client) createTopics(ctx context.Context, dry bool, p int32, rf int16,
 	}
 
 	req := kmsg.NewCreateTopicsRequest()
-	req.ValidateOnly = true
+	req.ValidateOnly = dry
 	for _, t := range topics {
 		rt := kmsg.NewCreateTopicsRequestTopic()
 		rt.Topic = t


turkenh:
> Great fix, thank you! Also, I thought I'd be the only one using this new package for a bit, very happy to have some external use as well :)

We wanted to use the package in the [Crossplane Kafka provider](https://github.com/crossplane-contrib/provider-kafka). I liked this library supporting full feature set of Kafka and preferred to use this instead of [more](https://github.com/Shopify/sarama), [popular](https://github.com/segmentio/kafka-go), [options](https://github.com/confluentinc/confluent-kafka-go), [out](https://github.com/lovoo/goka), there :) Great job 👍 

A comparison with other libraries in Readme would really be helpful.

> Also if you have any API suggestions on that package, I very much welcome it. I'm somewhat thinking to return named types for any slices at the moment, so that the types can add helper functions in the future. Not 100% sure yet.

One minor difficulty I had was, in the case of being interested in a single topic, I [needed](https://github.com/crossplane-contrib/provider-kafka/pull/1/files#diff-42dc40192aa2e1f87f7af11d2e9a0a87b2cf9e29cababfb472e9f46c4d8501c6R204) to first check if returned response array has something inside and check the first one with the assumption of I will find what I am looking for, as the first element of the array. This assumption bothers me a bit and some sort of helper function that could simply return the response for a specific topic that I am interested in would be helpful. Other than that, looking good so far.

twmb:
Cool, sounds good. I'll make that change and push it (will be a breaking change, but kadm is a separate module for a bit specifically for this reason :) )

twmb:
Although the error message might need improving (currently if the name does not exist, that'd return `missing "yourname"`, no further details). Open to a PR for that.

PR #113, opened 2021-11-30
Run go tidy on examples

travisdowns:
Also check in script to make this easier.


travisdowns:
I also added a script `examples/tidy-examples.sh` to automate the tidying. If you don't want that lmk and I'll remove it.

twmb:
This looks good, thank you, and the script is better than the one liner I did in the past :)

PR #115, opened 2021-12-20
Add enum for quotas match types

weeco:


generate/definitions/enums
```diff
@@ -119,6 +119,19 @@ TransactionState int8 camelcase (
   7: PrepareEpochFence
 )
 
+////////////
+// QUOTAS //
+////////////
```
twmb, line 124: Needs a newline after so that this section is not a part of the kmsg documentation.

generate/definitions/enums
```diff
@@ -119,6 +119,19 @@ TransactionState int8 camelcase (
...
+// QuotasMatchType specifies how to match a Quota entity as part of the DescribeClientQuotasRequestComponent
```
twmb, line 125: Can this have a trailing period? Same on the `//` doc comments below.

generate/definitions/enums
```diff
@@ -119,6 +119,19 @@ TransactionState int8 camelcase (
...
+QuotasMatchType int8 (
+  // Match on the name exactly
+  0: EXACT
+  // Match on the EntityType's default name
+  1: DEFAULT
+  // Match any specified name
```
twmb, line 131: Perhaps this should be `ANY`? And maybe this doc should be something like,
```
// Returns all named quotas and default quotas for the given EntityType.
```
? (if that's the correct behavior)

generate/definitions/enums
```diff
@@ -119,6 +119,19 @@ TransactionState int8 camelcase (
...
 )
 
+////////////
+// QUOTAS //
+////////////
+// QuotasMatchType specifies how to match a Quota entity as part of the DescribeClientQuotasRequestComponent
+QuotasMatchType int8 (
+  // Match on the name exactly
+  0: EXACT
+  // Match on the EntityType's default name
```
twmb, line 129: Maybe `Returns the default for the given EntityType.`

generate/definitions/enums
```diff
@@ -119,6 +119,19 @@ TransactionState int8 camelcase (
   7: PrepareEpochFence
 )
 
+////////////
+// QUOTAS //
+////////////
+// QuotasMatchType specifies how to match a Quota entity as part of the DescribeClientQuotasRequestComponent
+QuotasMatchType int8 (
+  // Match on the name exactly
```
twmb, line 127: Maybe
```
// Returns all quotas for the given EntityType with names equal to the Match field.
```
?

PR #120, opened 2022-01-01
Support delegation of offset management before consumer group topic consumption begins

michaelwilner:
One of the features preventing full adoption of `franz-go` in my application was the ability for consumer groups to swizzle offsets upon assignment of topics & partitions before consumption on those topics begins. In `confluent-kafka-go` this is illustrated via [this example](https://gist.github.com/edenhill/9dfc019b980a5eb3365c84524a1f12b0).

This PR offers a means via callback function (set with a GroupOpt) to intercept those assignments and change them before cursors are created.

### DIFF
--- a/pkg/kgo/config.go
+++ b/pkg/kgo/config.go
@@ -1319,20 +1319,19 @@ func RequireStableFetchOffsets() GroupOpt {
 	return groupOpt{func(cfg *cfg) { cfg.requireStable = true }}
 }
 
-// AdjustOffsetsBeforeAssign sets the function to be called when a group is joined
+// AdjustFetchOffsetsFn sets the function to be called when a group is joined
 // after offsets are fetched for those partitions so that a user can adjust them
 // before consumption begins.
 //
-// This function combined with OnPartitionsRevoked should not exceed the
-// rebalance interval. It is possible for the group, immediately after
-// finishing a balance, to re-enter a new balancing session.
+// This function combined should not exceed the rebalance interval. It is possible
+// for the group, immediately after finishing a balance, to re-enter a new balancing
+// session. This function is passed a context that is canceled if the current group
+// session finishes (i.e., after revoking).
 //
-// The AdjustOffsetsBeforeAssign function is passed the client's context, which is
-// only canceled if the client is closed.
-//
-// This function is not called concurrent with any other On callback, and this
-// function is given a new map that the user is free to modify.
-func AdjustOffsetsBeforeAssign(adjustOffsetsBeforeAssign func(context.Context, map[string]map[int32]Offset) (map[string]map[int32]Offset, error)) GroupOpt {
+// If you are resetting the position of the offset, you may want to clear any existing
+// "epoch" with WithEpoch(-1). If the epoch is non-negative, the client performs
+// data loss detection, which may result in errors and unexpected consuming offsets.
+func AdjustFetchOffsetsFn(adjustOffsetsBeforeAssign func(context.Context, map[string]map[int32]Offset) (map[string]map[int32]Offset, error)) GroupOpt {
 	return groupOpt{func(cfg *cfg) { cfg.adjustOffsetsBeforeAssign = adjustOffsetsBeforeAssign }}
 }
 
--- a/pkg/kgo/config.go
+++ b/pkg/kgo/config.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
+	"github.com/twmb/franz-go/pkg/kmsg"
 	"math"
 	"math/rand"
 	"net"
@@ -13,7 +14,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/twmb/franz-go/pkg/kmsg"
 	"github.com/twmb/franz-go/pkg/kversion"
 	"github.com/twmb/franz-go/pkg/sasl"
 )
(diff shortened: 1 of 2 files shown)


twmb:
This looks good, I think I may have a comment around some doc string aspect, but I plan to look more either tomorrow or Tuesday. Thanks for the PR, interesting idea! I also need to look into if relative offsets are supported in this code path (I think they may be, but just need to dive in).

michaelwilner:
@twmb thanks for taking a look! As far as I can tell, once we hit `assignPartitions` right after this callback, relative offsets are indeed calculated: https://github.com/twmb/franz-go/blob/bc391a31c0eb2ad202f6636e37d9326475b6e706/pkg/kgo/consumer.go#L647

michaelwilner:
@twmb anything noteworthy I should be aware of?

pkg/kgo/config.go
```diff
@@ -1317,6 +1319,23 @@ func RequireStableFetchOffsets() GroupOpt {
...
+//
+// This function combined with OnPartitionsRevoked should not exceed the
+// rebalance interval. It is possible for the group, immediately after
+// finishing a balance, to re-enter a new balancing session.
+//
+// The AdjustOffsetsBeforeAssign function is passed the client's context, which is
+// only canceled if the client is closed.
+//
+// This function is not called concurrent with any other On callback, and this
+// function is given a new map that the user is free to modify.
```
twmb, line 1334: This currently races with the callback for `OnPartitionsAssigned`, since fetching offsets is done in a separate goroutine. However, the purpose of this function isn't to modify anything that is passed to `OnPartitionsAssigned`, it's to modify the actual offsets that a consumer is about to consume. As well, this is passed a pretty distinct context: the context used here is canceled after OnRevoked finishes (after the heartbeat loop returns). As well, this doesn't duplicate the input map, but I think that's fine in this case. Also there's one more footgun that a user has to be aware of: epoch validation.

I think the doc bit above can be updated to:
```go
// AdjustOffsetsBeforeAssign sets the function to be called when a group is joined
// after offsets are fetched for those partitions so that a user can adjust them
// before consumption begins.
//
// This function combined should not exceed the rebalance interval. It is possible
// for the group, immediately after finishing a balance, to re-enter a new balancing
// session. This function is passed a context that is canceled if the current group
// session finishes (i.e., after revoking).
//
// If you are resetting the position of the offset, you may want to clear any existing
// "epoch" with WithEpoch(-1). If the epoch is non-negative, the client performs
// data loss detection, which may result in errors and unexpected consuming offsets.
```

Also, what do you think about `AdjustOffsetsBeforeAssignFn`? It's a bit of an ugly name but it somewhat fits a pattern: minus `Dialer`, the existing options that accept functions are all either `On`, `Callback`, or `Fn`:
```go
func Dialer(fn func(ctx context.Context, network, host string) (net.Conn, error)) Opt
func RetryBackoffFn(backoff func(int) time.Duration) Opt
func RetryTimeoutFn(t func(int16) time.Duration) Opt 
func ProducerOnDataLossDetected(fn func(string, int32)) ProducerOpt
func AutoCommitCallback(fn func(*Client, *kmsg.OffsetCommitRequest, *kmsg.OffsetCommitResponse, error)) GroupOpt
func OnPartitionsAssigned(onAssigned func(context.Context, *Client, map[string][]int32)) GroupOpt
func OnPartitionsLost(onLost func(context.Context, *Client, map[string][]int32)) GroupOpt
func OnPartitionsRevoked(onRevoked func(context.Context, *Client, map[string][]int32)) GroupOpt
```
"callback" could also work here but "fn" feels more like what this is doing. Maybe `AdjustFetchOffsetsFn`? I wonder if the current name could be confused with existing OnPartitionsAssigned semantics, when this is quite separate.

michaelwilner: Works for me - I’ll update the PR when I’m back at my keyboard 👍

michaelwilner: https://github.com/twmb/franz-go/pull/120/commits/2d6745063fbc8fe9cf678b21d0d0a7c8e8f4d93f

michaelwilner: Went with AdjustFetchOffsetsFn for brevity

pkg/kgo/config.go
```diff
@@ -1317,6 +1319,22 @@ func RequireStableFetchOffsets() GroupOpt {
...
+// before consumption begins.
+//
+// This function combined should not exceed the rebalance interval. It is possible
+// for the group, immediately after finishing a balance, to re-enter a new balancing
+// session. This function is passed a context that is canceled if the current group
+// session finishes (i.e., after revoking).
+//
+// If you are resetting the position of the offset, you may want to clear any existing
+// "epoch" with WithEpoch(-1). If the epoch is non-negative, the client performs
+// data loss detection, which may result in errors and unexpected consuming offsets.
```
twmb, line 1333: `unexpected consume offsets`? Probably another typo of mine; I wrote that doc up while cross referencing some of the code, too much window flipping.

michaelwilner: https://github.com/twmb/franz-go/pull/120/commits/f5414b2e8e0bc08109d6763fb6373052176c6ba0

pkg/kgo/config.go
```diff
@@ -1317,6 +1319,22 @@ func RequireStableFetchOffsets() GroupOpt {
 	return groupOpt{func(cfg *cfg) { cfg.requireStable = true }}
 }
 
+// AdjustFetchOffsetsFn sets the function to be called when a group is joined
+// after offsets are fetched for those partitions so that a user can adjust them
+// before consumption begins.
+//
+// This function combined should not exceed the rebalance interval. It is possible
```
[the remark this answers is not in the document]
michaelwilner: https://github.com/twmb/franz-go/pull/120/commits/f5414b2e8e0bc08109d6763fb6373052176c6ba0

twmb:
Looks good, thanks! I'll issue a tag shortly, I have some changes locally I'm working on as well that I aim to finish tonight.

twmb:
Tagged in v1.3.0.

PR #131, opened 2022-02-10
txns: don't ignore error in doWithConcurrentTransactions

eduard-netsajev:


twmb (approved):
This is a really good find and a great fix. I'll cut a patch release with this tonight as well after some integration testing. The ramifications for this are not the clearest: afaict, missing the error here could at worse lead to data loss, but more likely would lead to a stalled pipeline and would be pretty noticeable in graphs.

PR #133, opened 2022-02-10
config: add producer option to configure transactions backoff duration

eduard-netsajev:
And lower the default from 100ms to 20ms in line with Java Kafka client

### DIFF
--- a/pkg/kgo/config.go
+++ b/pkg/kgo/config.go
@@ -782,6 +782,19 @@ func WithHooks(hooks ...Hook) Opt {
 	return clientOpt{func(cfg *cfg) { cfg.hooks = append(cfg.hooks, hooks...) }}
 }
 
+// ConcurrentTransactionBackoff sets the backoff interval to use during
+// transactional requests in case we encounter CONCURRENT_TRANSACTIONS error,
+// overriding the default 20ms.
+//
+// Sometimes, when a client begins a transaction quickly enough after finishing
+// a previous one, Kafka will return a CONCURRENT_TRANSACTIONS error. Clients
+// are expected to backoff slightly and retry the operation. Lower backoffs may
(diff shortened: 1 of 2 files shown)


pkg/kgo/config.go
```diff
@@ -1037,6 +1039,12 @@ func TransactionTimeout(timeout time.Duration) ProducerOpt {
 	return producerOpt{func(cfg *cfg) { cfg.txnTimeout = timeout }}
 }
 
+// ConcurrentTransactionBackoff sets the backoff interval to use during
+// transactional requests in case we encounter CONCURRENT_TRANSACTIONS error.
```
twmb, line 1043: Can this documentation be expanded to,
```go
// ConcurrentTransactionBackoff sets the backoff interval to use during
// transactional requests in case we encounter CONCURRENT_TRANSACTIONS error,
// overriding the default 20ms.
//
// Sometimes, when a client begins a transaction quickly enough after finishing
// a previous one, Kafka will return a CONCURRENT_TRANSACTIONS error. Clients
// are expected to backoff slightly and retry the operation. Lower backoffs may
// increase load on the brokers, while higher backoffs may increase transaction
// latency in clients.
```

pkg/kgo/config.go
```diff
@@ -1037,6 +1039,12 @@ func TransactionTimeout(timeout time.Duration) ProducerOpt {
...
+func ConcurrentTransactionBackoff(backoff time.Duration) ProducerOpt {
```
twmb, line 1044: Can this be moved below into a ConsumerOpt? Or perhaps above to just a plain Opt, since transactions kinda encompass both producing & consuming.

eduard-netsajev: Done

twmb:
Looks great, thank you!

PR #148, opened 2022-03-18
kgo/config: change default maxPartBytes to 1MB

jcsp:
This is the same as the Apache Kafka default.  There
are two motivations:
- 10MB is a rather large default: results in a client
  requesting up to a gig of data for a 100 partition
  topic, which hits server side limits and gets clamped.
- Simplify support by having default client request
  sizes not depend on which of the popular clients
  the user is on.


twmb:
Broker's should be obeying the FetchMaxBytes, but I'm all for matching The Apache Kafka default, thanks! I'll update the doc on the config in a followup commit.

PR #170, opened 2022-06-08
Fix kadm Sorted: TopicsSet -> TopicsList function

larry-cdn77:
Just a quick fix to one of the kadm convenience representations

Tested with:

```go
	dg, err := adm.DescribeGroups(context.Background(), group)
	ts := dg.AssignedPartitions()

	tl := ts.Sorted()

	fmt.Println(tl[0].Topic)
	fmt.Println(len(ts[tl[0].Topic]), len(tl[0].Partitions))
```

to see if partitions in the output TopicsList match those in the input TopicsSet

As a side note, I used librdkafka previously and recall seeing a 'top-par' type often. In kadm it is equivalent to the Partition Type and it can be a handy abstraction (not always). I wondered if some other kadm structs could embed Partition. Perhaps Offset, ListedOffset, DescribedLogDirPartition or DeleteRecordsResponse? I am happy to submit a suggestion, if it could be useful.

### DIFF
--- a/pkg/kadm/kadm.go
+++ b/pkg/kadm/kadm.go
@@ -452,6 +452,7 @@ func (s TopicsSet) Sorted() TopicsList {
 			tps.Partitions = append(tps.Partitions, p)
 		}
 		tps.Partitions = int32s(tps.Partitions)
+		l = append(l, tps)
 	}
 	sort.Slice(l, func(i, j int) bool { return l[i].Topic < l[j].Topic })
 	return l


twmb:
Also ty for the pr :)

This should require a patch release, but I'll wait for your reply to the above^ so that we can potentially squeeze in a small bit of other work first.

larry-cdn77:
> It's too late to change the types on `Offset`, `ListedOffset`, etc. since kadm was stabilized ~1w ago. However, all of these types should have the Topic and Partition fields present already. What would be the benefit for embedding the partition type? Perhaps an alternative would be to add helper functions to convert these types to partitions.

Thanks for your comments. I can't think of a benefit other than fewer types with the downside of more nesting. It might as well be left as it is 😄

twmb:
I have code locally that implements the AlterPartitionAssignments / ListPartitionReassignments APIs, I'll vet that and tag probably 1.1.0 this weekend

larry-cdn77:
Thank you. No rush, I am not blocked by the lack of TopicsSet Sorted.

twmb:
@larry-cdn77 I've pushed this in kadm release v1.1.0:

https://github.com/twmb/franz-go/releases/tag/pkg%2Fkadm%2Fv1.1.0

Thank you again for the fix!

twmb:
As it turns out, `git stash` to check a few things has betrayed me, so this is _actually_ working in v1.1.1:

https://github.com/twmb/franz-go/releases/tag/pkg%2Fkadm%2Fv1.1.1

PR #171, opened 2022-06-09
Fix sr.URLs handling of https

kklipsch:
Currently sr.URLs will prepend http:// onto urls with https:// addresses making it unusable with https based schema registries.


twmb (approved):
Ah my mistake for this obvious oversight, thanks!

Issue #179, opened 2022-06-29
Prometheus metrics instrumentation is fairly unergonomic


kklipsch:
I've run into a few issues when using the prometheus metrics plugin.  Particularly:
- No example of how to hook up the plugin with the default registry.
- If you do hookup the plugin with the default registry you cannot have more than one kafka client because the same metric gets registered twice on client creation, I haven't encountered anything else that requires this.  Its not too onerous except in tests.
- Even if you do use different registries, its not obvious how to hook them up to your metrics endpoint.

twmb:
What's the recommendation? Alternatively, PR?

kklipsch:
Let me noodle on one.  In the meantime can you put an example in the docs of how you think the plugin should be used?

twmb:
There are already two examples linked in the metrics & logging section of the readme:
https://github.com/twmb/franz-go/#metrics--logging
https://github.com/twmb/franz-go/tree/master/examples/hooks_and_logging/expansive_prometheus
https://github.com/twmb/franz-go/tree/master/examples/hooks_and_logging/plugin_prometheus

as well as another use case in the bench code (another example, but a useful utility)

kklipsch:
Ah sorry, didn't see those.  I was looking in the godoc.  I'll put up a PR to link to them there.

I think the expansive example actually shows a way around the problems I mention above at the cost of  the interface checks would have to be duplicated to client code and kept up to date with the client library at run/test time instead of compile time.

It gives me more to work with though so I'll see if I can figure out a way to make them more ergonomic.

kklipsch:
https://github.com/twmb/franz-go/pull/181 is the documentation change.

twmb:
I'm going to go ahead and close this for now, I'm not sure of a better way to do this, and I think the documentation should help :)

kklipsch:
Yeah sorry, I got distracted.  I'll open a PR if i figure out a better solution.

PR #181, opened 2022-07-11
Add a link to the prometheus examples to the godoc

kklipsch:
https://github.com/twmb/franz-go/issues/179


twmb (approved):
awesome, ty.

Unfortunately, to actually release this doc line, I'll need to tag kprom itself with a specific release, and there's no code change to actually release. It might be worth it to tag this in a patch release, but I'm not sure. wdyt?

Issue #190, opened 2022-08-09
panic in updateMetadataLoop : runtime error: index out of range [-1]


jcsp:
franz-go v1.6.0

I've just seen this for the first time:
```
panic: runtime error: index out of range [-1]

goroutine 1678 [running]:
github.com/twmb/franz-go/pkg/kgo.(*sink).removeRecBuf(0x7f2360?, 0xc00e80b700?)
        /root/go/pkg/mod/github.com/twmb/franz-go@v1.6.0/pkg/kgo/sink.go:978 +0x1af
github.com/twmb/franz-go/pkg/kgo.(*topicPartition).migrateProductionTo(0xc2380356e0, 0xc36611c9c0)
        /root/go/pkg/mod/github.com/twmb/franz-go@v1.6.0/pkg/kgo/topics_and_partitions.go:472 +0x35
github.com/twmb/franz-go/pkg/kgo.(*Client).mergeTopicPartitions(0xc00df57800, {0x7fff75e79cd6, 0xc}, 0xc8383be810, 0xce96771630, 0x1, 0x40e725?, 0x7f0c128963e0?, 0x0?)
        /root/go/pkg/mod/github.com/twmb/franz-go@v1.6.0/pkg/kgo/metadata.go:691 +0x1053
github.com/twmb/franz-go/pkg/kgo.(*Client).updateMetadata(0xc00df57800)
        /root/go/pkg/mod/github.com/twmb/franz-go@v1.6.0/pkg/kgo/metadata.go:354 +0x7ee
github.com/twmb/franz-go/pkg/kgo.(*Client).updateMetadataLoop(0xc00df57800)
        /root/go/pkg/mod/github.com/twmb/franz-go@v1.6.0/pkg/kgo/metadata.go:184 +0x5a6
created by github.com/twmb/franz-go/pkg/kgo.NewClient
        /root/go/pkg/mod/github.com/twmb/franz-go@v1.6.0/pkg/kgo/client.go:213 +0x9bc
```

The exact code in use is https://github.com/jcsp/kgo-verifier/tree/c6540c4b68d6a81a9677fa66ebd9e4576aecc1d9

It's a traffic generator that runs a bunch of threads with their own client, a producer+consumer in each one, producing one message for each consumed.

The cluster under test is redpanda nightly 0.0.0~20220809gitfbdafab-1 , running with 9 nodes and 67490 partitions in 1 topic.

twmb:
This is perplexing

[...]

PR #193, opened 2022-08-16
Use correct error in debug log for request failure

keitwb:
It was using plain `err` twice and never logging the `retryErr`.


twmb (approved):
Nice find, thank you. Were you looking into a problem to find this, or just browsing the code?

twmb:
This will be out in the next release, either a patch release if I fix only #190 and then release, or a feature release if I address everything else right after it. #190 is tricky.

Issue #195, opened 2022-08-20
[Question] Dynamically leave and join to consumer group


dwagin:
Please help me figure out how to dynamically leave and join to consumer group?
I only found LeaveGroup(). Is there no possibility to join back?

twmb:
The client doesn't implement the option to rejoin after leaving. What's the use case for this?

twmb:
I'd recommend recreating the client, it's cheap to do so and would have roughly the same effect. If possible, don't leave the because leaving and rejoining would cause rebalance storms. lmk if you have a feature proposal here, closing this issue for now

dwagin:
This requires the ability to validate configuration from the outside. I will prepare a pull request about it.

PR #197, opened 2022-08-23
client: added the ability to validate the configuration

dwagin:
For recreating the client, need the ability to check the configuration separately.

in continuation #195.

### DIFF
--- a/pkg/kgo/client.go
+++ b/pkg/kgo/client.go
@@ -217,6 +217,16 @@ func NewClient(opts ...Opt) (*Client, error) {
 	return cl, nil
 }
 
+// Validate returns an error if the options are invalid.
+func Validate(opts ...Opt) error {
+	cfg := defaultCfg()
+	for _, opt := range opts {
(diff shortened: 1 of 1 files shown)


twmb (commented):
I think that since `NewClient` does more initialization and validation than this, the `Validate` function here isn't comprehensive enough to all the errors that could be returned. Also, I think the name needs to be a bit more specific, either `ValidateOpts` or `CheckNewClient` or `ValidateNewClient` or something.

I'd split the two goroutines started in `NewClient` into something like `func (*Client) start()`, add `func (*Client) init(...Opt) error`, and then have `NewClient` be `init(opt...); errcheck; start()` and `ValidateOpts` to be just `return init()`.

Although this runs the risk someday of accidentally adding a goroutine in the middle of NewClient, so then add a doc line warning maintainers in a year not to add goroutines.

Also, since 1.7 has just released, this will probably wait a bit until a few more features for 1.8 stack up, perhaps a month or three in the worst case.

PR #203, opened 2022-09-12
chore: upgrade go to 1.15

robsonpeixoto:
Some packages are using go 1.16 and another usine 1.15.
This PR will update all `go 1.15` to `go 1.16`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/twmb/franz-go
 
-go 1.15
+go 1.16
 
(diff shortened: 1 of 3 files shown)


twmb:
I don't think there's a need for this, it's alright if different packages within franz-go use different Go versions. I _think_ the current deps pinned by franz-go are all 1.15. That said, compress recently upgraded to require 1.17, so I may bump to that at some point...

PR #206, opened 2022-09-22
Record TimestampType functionality match comment

JacobSMoller:
Now returns 1 if bit 4 is set as stated in the comment

### DIFF
--- a/pkg/kgo/record_and_fetch.go
+++ b/pkg/kgo/record_and_fetch.go
@@ -44,7 +44,7 @@ func (a RecordAttrs) TimestampType() int8 {
 	if a.attrs&0b1000_0000 != 0 {
 		return -1
 	}
-	return int8(a.attrs & 0b0000_1000)
+	return int8(a.attrs&0b0000_1000) >> 3
 }
 
 // CompressionType signifies with which algorithm this record was compressed.


pkg/kgo/record_and_fetch.go
```diff
@@ -44,7 +44,10 @@ func (a RecordAttrs) TimestampType() int8 {
 	if a.attrs&0b1000_0000 != 0 {
 		return -1
 	}
-	return int8(a.attrs & 0b0000_1000)
+	if a.attrs&0b0000_1000 == 8 {
```
twmb, line 47: What do you think about changing this to `return (a.attrs&0b0000_1000) >> 4 == 1`?

JacobSMoller: Hmm im fine with either but they both return a bool not an int8. and it should be >> 3 right?

twmb: Ah right, it can be `return int8((a.attrs&0b0000_1000) >> 4)`

brunsgaard: smart 👍

JacobSMoller: Nice. But still it should be >> 3. Right

`int8((8 & 0b0000_1000) >> 4) == 0` 
`int8((8 & 0b0000_1000) >> 3) == 1`

pkg/kgo/record_and_fetch.go
```diff
@@ -44,7 +44,7 @@ func (a RecordAttrs) TimestampType() int8 {
 	if a.attrs&0b1000_0000 != 0 {
 		return -1
 	}
-	return int8(a.attrs & 0b0000_1000)
+	return int8((a.attrs & 0b0000_1000)) >> 3
```
twmb, line 47: Unnecessary double parens, mind removing and squashing into one commit? I could too if you don't get to it before I do (though I'm working my way through #210 so I won't be getting to this _that_ fast)

JacobSMoller: Squashed. :)

Issue #218, opened 2022-10-11
Build failure on ARM


mihaitodor:
Hey @twmb, thanks for this awesome package! I noticed that commit 31f3f5f changed `p.waitBuffer = make(chan struct{}, 32)` to `p.waitBuffer = make(chan struct{}, math.MaxInt64)` [here](https://github.com/twmb/franz-go/blob/7b8d404ce83236a1d1b529ccd8b3b312f8dc0197/pkg/kgo/producer.go#L100), which seems to upset the compiler on 32-bit platforms. I noticed this after franz-go was updated to v1.8.0 in Benthos [here](https://github.com/benthosdev/benthos/actions/runs/3226927155/jobs/5281057979).

twmb:
Interesting -- the `sync/atomic` notes at the bottom indicate that 64 bit atomic operations are not guaranteed to be aligned on ARM. I don't have access to a 32 bit system to check alignment of all the atomics currently used, but they're almost certainly not aligned.

Also looking at the Benthos source, I'm pretty sure _its_ atomics are not aligned either, example:
* Three `sync.Map` fields used here: https://github.com/benthosdev/benthos/blob/f7dac91e4e568848b406142734dc0fa6bf45e6d6/internal/bundle/tracing/events.go#L29
* `sync.Map` ends in an `int`, which should be 32 bit aligned on a 32 bit system: https://cs.opensource.google/go/go/+/refs/tags/go1.19.2:src/sync/map.go;l=66
* Meaning the second `sync.Map` is not aligned and the atomic ops within `sync.Map` are likely invalid (specifically within the `read atomic.Value`

I also highly doubt that the many many imported modules are all aligning atomics properly.

I'll see about setting up a 32 bit `go vet` with github actions, but it might be worth dropping arm support from Benthos' goreleaser.

I'll also change the `make` there to use math.MaxInt32 ... I'm pretty sure that should be safe, I'll stare at the code again. This is _probably_ not important enough for a patch release, so I'd like to wait ~1day for a few other features and roll this into v1.9. I wasn't expecting to add a few things so quickly after v1.8, but we're here now.

twmb:
In the above comment, I realize that the three `sync.Map` field internals are aligned, because I forgot the latter half of the sync/atomic note while writing: the first field of any struct is guaranteed to be 64 bit aligned.

So, benthos is probably fine. Not sure about the dependencies, though.

mihaitodor:
[...]

PR #223, opened 2022-10-16
Fix integration test action

mihaitodor:
Hey @twmb, following up on #218, I think this should do the trick using the `bitnami/kafka` container in Zookeeperless mode. I had to adjust the tests a bit so the `KGO_SEEDS` env var is read by each test which needs to connect to Kafka.

I also tried to use Redpanda, but there are some issues with it. Could you please let me know how you got the tests to pass against it? It may just be that I'm not configuring it correctly (maybe it doesn't like the fact that I'm running it as a single node?), but my Kafka knowledge is too limited to troubleshoot it. Here's what I tried:

```shell
> docker run --rm -p9092:9092 docker.vectorized.io/vectorized/redpanda:latest start --default-log-level=debug --smp 1 --reserve-memory 0M --overprovisioned --set redpanda.enable_transactions=true --kafka-addr 0.0.0.0:9092 --advertise-kafka-addr 127.0.0.1:9092
> # Wait for container to start
> KGO_TEST_RF=1 KGO_TEST_RECORDS=50000 go test -v -count 1 ./...
```

I pulled the current `latest` tag of the Redpanda container (`sha256:84c8cd98ed7a50e4f378af55600c92c4fb097afab4ec164c3ded9b0e884754c4`).

I guess this log entry might be relevant: [ERROR] fatal InitProducerID error, failing all buffered records; broker: 1, err: INVALID_PRODUCER_EPOCH: Producer attempted an operation with an old epoch.

Here are the full logs:

[redpanda.log](https://github.com/twmb/franz-go/files/9793352/redpanda.log)
[test.log](https://github.com/twmb/franz-go/files/9793354/test.log)

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -14,7 +14,7 @@ jobs:
   golangci:
     if: github.repository == 'twmb/franz-go'
     runs-on: ubuntu-latest
-    name: 'golangci-lint on amd64'
+    name: "golangci-lint on amd64"
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v3
@@ -28,7 +28,7 @@ jobs:
     if: github.repository == 'twmb/franz-go'
     needs: golangci
     runs-on: ubuntu-latest
-    name: 'vet on arm'
+    name: "vet on arm"
     steps:
       - uses: actions/checkout@v3
         with:
@@ -57,31 +57,34 @@ jobs:
             echo "staticcheck ./..."
             staticcheck -checks 'all,-ST1003,-SA1012,-ST1016,-SA1019,-SA2001' ./... # actually contains atomicalign check
 
-# TODO: fix
-#  integration-test:
-#    if: github.repository == 'twmb/franz-go'
-#    needs: golangci
-#    runs-on: ubuntu-latest
-#    name: 'integration test kafka'
-#    container: golang:1.19.2
-#    services:
-#      zk:
-#        image: bitnami/zookeeper:latest
-#        ports:
-#          - 2181:2181
-#        env:
-#          ALLOW_ANONYMOUS_LOGIN: yes
-#      kafka:
-#        image: bitnami/kafka:latest
-#        ports:
-#          - 9092:9092
-#        env:
-#          ALLOW_PLAINTEXT_LISTENER: yes
-#          KAFKA_CFG_ZOOKEEPER_CONNECT: zk:2181
-#    steps:
-#      - uses: actions/checkout@v3
-#      - run: go test ./...
-#        env:
-#          KGO_TEST_RF: 1
-#          KGO_SEEDS: kafka:9092
-#          KGO_TEST_RECORDS: 50000
+  integration-test:
+    if: github.repository == 'twmb/franz-go'
+    needs: golangci
+    runs-on: ubuntu-latest
+    name: "integration test kafka"
+    container: golang:1.19.2
+    services:
+      kafka:
+        image: bitnami/kafka:latest
+        ports:
+          - 9092:9092
+        env:
+          KAFKA_ENABLE_KRAFT: yes
+          KAFKA_CFG_PROCESS_ROLES: controller,broker
+          KAFKA_CFG_CONTROLLER_LISTENER_NAMES: CONTROLLER
+          KAFKA_CFG_LISTENERS: PLAINTEXT://:9092,CONTROLLER://:9093
+          KAFKA_CFG_LISTENER_SECURITY_PROTOCOL_MAP: CONTROLLER:PLAINTEXT,PLAINTEXT:PLAINTEXT
+          KAFKA_CFG_CONTROLLER_QUORUM_VOTERS: 1@127.0.0.1:9093
+          # Set this to "PLAINTEXT://127.0.0.1:9092" if you want to run this container on localhost via Docker
+          KAFKA_CFG_ADVERTISED_LISTENERS: PLAINTEXT://kafka:9092
+          KAFKA_CFG_BROKER_ID: 1
+          ALLOW_PLAINTEXT_LISTENER: yes
+          KAFKA_KRAFT_CLUSTER_ID: XkpGZQ27R3eTl3OdTm2LYA # 16 byte base64-encoded UUID
+          # BITNAMI_DEBUG: true # Enable this to get more info on startup failures
+    steps:
+      - uses: actions/checkout@v3
+      - run: go test ./...
+        env:
+          KGO_TEST_RF: 1
+          KGO_SEEDS: kafka:9092
+          KGO_TEST_RECORDS: 50000
--- a/pkg/kgo/group_test.go
+++ b/pkg/kgo/group_test.go
@@ -37,6 +37,7 @@ func TestGroupETL(t *testing.T) {
 
 	go func() {
 		cl, _ := NewClient(
+			getSeedBrokers(),
 			WithLogger(BasicLogger(os.Stderr, testLogLevel, nil)),
 			MaxBufferedRecords(10000),
 		)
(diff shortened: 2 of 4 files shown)


ssorren:
As far as I know, Redpanda does not yet support transactions. I encountered the same error myself and read their docs.

mihaitodor:
I thought it does if you use `--set redpanda.enable_transactions=true`, but maybe the implementation isn't complete yet.

ssorren:
I thought it was that the community edition did not support it, I may be mistaken

ssorren:
Ahhh: `Compacted topic doesn't support transactional processing` from the Redpanda logs. That's what is causing my issue. Not sure about the integration tests.

mihaitodor:
Oh, I see. Thank you for digging that up! ❤️ Dunno if the failures I'm seeing are related (maybe @twmb knows), but, hopefully, they'll catch up soon :)

twmb:
I thought I tested against Redpanda above, but I had a stray Kafka broker alive which allowed the tests to be successful. Redpanda's transactions are a bit WIP right now and _should_ be fixed in the next release. I'll test against it, but this repo should stick to Kafka for now.

mihaitodor:
OK, makes sense to stick with Kafka for now and maybe add Redpanda as an extra test in the future.

Not sure what to make of the failure... It worked great for me: https://github.com/mihaitodor/franz-go/actions/runs/3257547284/jobs/5348876552. Hope it's not a heisenbug.

twmb:
I'll see if I can recreate this locally -- a sequence number reset can happen in a few scenarios, but they shouldn't be happening in this test.

ssorren:
> Since Kafka is running in Zookeeperless mode, I wonder if it might have some rough corners. Hopefully not.

No, I don't think this has anything to do with Zookeeper. I'm able to use Redpanda fine so long as I don't us a transactional producer. But my use case calls for transactions on a compacted topic.

twmb:
It looks like Kafka isn't respecting its 5-wide idempotency window. I'm checking to see if something in Bitnami's image specifically changed it, or what. From local reproducing,

* send partition 13 sequence # start 0
* send partition 13 sequence # start 701
* failure, retry the above two
* receive OOOSN because Kafka is not de-duplicating the repeated messages

twmb:
I think this is bringing up an issue that has always been possible in Kafka, but is now visible with whatever changes they're making in Kraft. Apparently, the first produce request's sequence number does not need to start at 0. That's problematic:

* client creates topic foo, kafka replies "sure, created"
* client sends produce request A with 5 messages, sends produce request B with 5 messages _before_ a response for A is received (base sequence numbers 0 and 5)
* request A is rejected with NOT_LEADER_FOR_PARTITION because the topic is not fully initialized, request B is accepted
* Kafka now has 5 records locally, _starting_ at a base sequence of 5 and it is expecting 10
* client re-sends request A and B
* request A is rejected as out of order, since it has base sequence 0

I'll check the Kafka source to see if it truly does accept a non-zero sequence number for the first request.

It seems that Kraft replies NOT_LEADER_FOR_PARTITION for a bit, quite a bit more frequently.

twmb:
https://github.com/apache/kafka/blob/3e7eddecd6a63ea6a9793d3270bef6d0be5c9021/core/src/main/scala/kafka/log/ProducerStateManager.scala#L235-L236

mihaitodor:
Very interesting, thanks for sharing that insight!

Out of curiosity, I also tried to see if I can use Confluent's Kafka container in Zookeeperless mode, since they offer some setup example here https://github.com/confluentinc/cp-all-in-one/blob/7.2.1-post/cp-all-in-one-kraft/docker-compose.yml but I guess I'm doing smth wrong with the way I'm volume mounting that silly shell script: https://github.com/mihaitodor/franz-go/actions/runs/3269542233/jobs/5377158865. GitHub has [some bugs](https://stackoverflow.com/questions/58997638/docker-container-volume-does-not-get-mounted-in-github-actions) around volume mounting stuff in service containers and getting around them is very frustrating... Anyway, I guess it will behave the same as the Bitnami one based on your investigation above.

twmb:
For context (creating Kafka issue now), this is my docker compose:

```yaml
version: "3.7"
services:
  kafka:
    image: bitnami/kafka:latest
    network_mode: host
    environment:
      KAFKA_ENABLE_KRAFT: yes
      KAFKA_CFG_PROCESS_ROLES: controller,broker
      KAFKA_CFG_CONTROLLER_LISTENER_NAMES: CONTROLLER
      KAFKA_CFG_LISTENERS: PLAINTEXT://:9092,CONTROLLER://:9093
      KAFKA_CFG_LISTENER_SECURITY_PROTOCOL_MAP: CONTROLLER:PLAINTEXT,PLAINTEXT:PLAINTEXT
      KAFKA_CFG_CONTROLLER_QUORUM_VOTERS: 1@127.0.0.1:9093
      # Set this to "PLAINTEXT://127.0.0.1:9092" if you want to run this container on localhost via Docker
      KAFKA_CFG_ADVERTISED_LISTENERS: PLAINTEXT://127.0.0.1:9092
      KAFKA_CFG_BROKER_ID: 1
      ALLOW_PLAINTEXT_LISTENER: yes
      KAFKA_KRAFT_CLUSTER_ID: XkpGZQ27R3eTl3OdTm2LYA # 16 byte base64-encoded UUID
      BITNAMI_DEBUG: true # Enable this to get more info on startup failures
```

How I am running this:
```
export KGO_TEST_RF=1
export KGO_TEST_RECORDS=50000
export KGO_LOG_LEVEL=debug
```
and then
```
docker compose down; sleep 1; docker compose up -d; sleep 5; while go test -run '(Txn|Group)' > logs ; do echo "whoo"; docker compose down; sleep 1 ; docker compuse up -d ; sleep 5; done
```

twmb:
https://issues.apache.org/jira/browse/KAFKA-14312
https://issues.apache.org/jira/browse/KAFKA-14313

AFAICT, there is no solution to 14312.

One approach I could try is to sleep 1s after creating topics in the test.

twmb:
Another odd issue,
```
__transaction_state partition 33 offset 7 at [2022-10-18 11:15:37.821]
TxnMetadataKey(0) 9f87dc04dc3f4d5b15ef3072c531cf46327278307df8e149fa966462cd40c10b
TxnMetadataValue(0)
      ProducerID           41
      ProducerEpoch        0
      TimeoutMillis        120000
      State                PrepareCommit
      Topics               __consumer_offsets=>[13] e7c7d971626fbaf4bfb33975e57089167939e6acabb4c4fc534eb148462e45cc=>[4 5 12 16]  
      LastUpdateTimestamp  1666113337821
      StartTimestamp       1666113335311


__transaction_state partition 33 offset 8 at [2022-10-18 11:16:11.419]
TxnMetadataKey(0) 9f87dc04dc3f4d5b15ef3072c531cf46327278307df8e149fa966462cd40c10b
TxnMetadataValue(0)
      ProducerID           41
      ProducerEpoch        0
      TimeoutMillis        120000
      State                CompleteCommit
      Topics     
      LastUpdateTimestamp  1666113337821
      StartTimestamp       1666113335311
```

Kafka took 34 seconds to finish a commit within itself. This is causing duplicates, because a rebalance happened in the middle and moved partitions.

Technically, I should be using `RequireStableFetchOffsets`. I'm not sure why Kafka took 34 seconds to finish a transaction within itself, though -- the client goes to a lot of lengths to avoid the need for `RequireStableFetchOffsets`. I'll add the option and see how things go ...

twmb:
I've added a bunch of commits in https://github.com/twmb/franz-go/pull/224 that should address failures, and cherry picked this commit. I've looped tests for a few hours and have not seen any failures locally. Closing this and merging the other PR.

This also finally makes it worth it to enable branch protection (and require all commits to go through PRs), which is also nice because I've occasionally introduced a compile failure without testing (and immediately fixed).

Now that this is merged, I plan to address 214, 220, 221, 222 and then release 1.19. Hopefully the rest of this is quick.

mihaitodor:
That's really awesome, thank you for all this work @twmb!

twmb:
For full posterity in case I need it later,

```sh
#!/bin/bash

docker compose down
sleep 1
docker compose up -d

until rpk topic list; do
	sleep 1
done

while go test -run Txn > logs; do
	echo "whoo"
	docker compose down
	sleep 1
	docker compose up -d
	until rpk topic list; do
		sleep 1
	done
done
```

Also if you want to help with #225, I need to change it to expose a port ;)
Turns out redpanda nightly _does_ work, so long as I disable unsafe behavior in tests.

PR #225, opened 2022-10-19
franz-go: test against redpanda

twmb:
Turns out, redpanda does not like the EndBeginTxnUnsafe option. That's valid.


twmb:
Also to add: ignore integration tests against README, CHANGELOG only changes.

PR #263, opened 2022-11-30
Deduplicate code2err

PleasingFungus:
Instead of listing every Kafka error code twice (in the struct declarations and in the `code2err` list), use the former to populate the latter.

Also, fix `IneligibleReplica` and `NewLeaderElected` not being included in `code2err`.


PleasingFungus:
(Seems plausible that this file should be generated, rather than needing to make two massive lists which each list every Kafka error type. But maybe generating it is more work than is worthwhile.)

pkg/kerr/kerr.go
```diff
@@ -184,113 +185,125 @@ var (
...
+		PositionOutOfRange,
+		UnknownTopicID,
+		DuplicateBrokerRegistration,
+		BrokerIDNotRegistered,
+		InconsistentTopicID,
+		InconsistentClusterID,
+		TransactionalIDNotFound,
+		FetchSessionTopicIDError,
+		IneligibleReplica,
+		NewLeaderElected,
```
twmb, line 301: I'm actually not sure I'm too positive on this change -- the prior code makes it easy to see that I'm missing errors (currently 107 and 108, actually).

If we go with this change, I'd like this error list to be ordered to more easily visually see that errors are missing. that said, I do think the older list is easier to maintain -- and it changes so rarely that the extra bit of duplication in the numbers isn't really a maintenance burden.

twmb:
Given that this file is rarely touched, and that I think the numbers help me maintain the list, I'm going to close this -- I think the bottom list would be more compelling it were autogenerated from the top list, but that's a good chunk more complexity and it'd require stuffing go:generate validation into some test pipeline to ensure `go generate` is never missed.

I did go ahead and add the two missing errors: https://github.com/twmb/franz-go/pull/269/commits/bb3d9c1b6e6ffbafce3ae0446db7c7b64000e298

PleasingFungus:
That's perfectly fair. FWIW, my IDE showed me that the last two errors were unused - that's the only reason I noticed.

twmb:
Theoretically they could be used by the ecosystem…but the chance of that is
near 0 since this project does not have extremely high visibility and the
errors are newish for new advanced requests

PR #265, opened 2022-11-30
Use atomic.Bool in Go 1.19+

PleasingFungus:
Instead of using our own atomicBool type, let's use the standard atomic.Bool type in versions of Go that contain it.

### DIFF
--- a/pkg/kgo/atomic_maybe_work.go
+++ b/pkg/kgo/atomic_maybe_work.go
@@ -2,27 +2,6 @@ package kgo
 
 import "sync/atomic"
 
-// a helper type for some places
-type atomicBool uint32
-
-func (b *atomicBool) set(v bool) {
-	if v {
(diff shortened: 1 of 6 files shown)


twmb (commented):
This change would be nice, but more in the context of switching to 1.19+ wholesale. Having the old code exist but gated to a version means the new atomic.Bool aspect doesn't provide any benefit. So, that said -- I think I'll close this, but thank you for the effort and idea

PleasingFungus:
That's perfectly fair. Thanks for the prompt response!

twmb:
This might actually be relevant -- I'm going to introduce atomic Int and use proper APIs for 1.19+ too. I know atomic.Int is special cased internally with proper alignment on arm32. Not sure what's up with #286, but hopefully proper atomic types will fix it.

PR #266, opened 2022-11-30
Use a named enum for compression codecs

PleasingFungus:
Rather than using int literals everywhere.

If we want to be a little more cautious, we could use explicit values for each `codecType` constant rather than just the values implied by `iota` - e.g. `codecGzip = 1`, etc.

### DIFF
--- a/pkg/kgo/compression.go
+++ b/pkg/kgo/compression.go
@@ -24,31 +24,41 @@ func (s *sliceWriter) Write(p []byte) (int, error) {
 
 var sliceWriters = sync.Pool{New: func() interface{} { r := make([]byte, 8<<10); return &sliceWriter{inner: r} }}
 
+type codecType int8
+
+const (
+	codecNone = iota
+	codecGzip
+	codecSnappy
+	codecLZ4
+	codecZstd
+)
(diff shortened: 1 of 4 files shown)


twmb (approved):

pkg/kgo/compression.go
```diff
@@ -24,31 +24,41 @@ func (s *sliceWriter) Write(p []byte) (int, error) {
 
 var sliceWriters = sync.Pool{New: func() interface{} { r := make([]byte, 8<<10); return &sliceWriter{inner: r} }}
 
+type codecType int8
+
+const (
+	codecNone = iota
```
cespare, line 30: These consts should have type `codecType`.

PleasingFungus: Whoops, good catch. Opened https://github.com/twmb/franz-go/pull/270 .

cespare: @PleasingFungus and I work together.

twmb: Awesome! Happy to see a prominent member of the Go community in my humble repo 😅

PR #267, opened 2022-12-01
Fix various docs typos

PleasingFungus:
And minor grammar errors.

This also moves `The cooperative balancer` and `Static membership` into the `Consumer groups` section of `producing and consuming`, since those topics are (I believe) only relevant for users of consumer groups. (Perhaps consumer groups should have its own dedicated document? Its headings are nested quite deeply...)


docs/producing-and-consuming.md
```diff
@@ -202,7 +202,7 @@ Because an EOS consumer is difficult to implement correctly, all details have
 been abstracted away to a [`GroupTransactSession`][20] type. See the
 [transactions](./transactions.md) page for more details.
 
-### The cooperative balancer
```
twmb, line 205: I see what you mean about putting these under the consumer group only, but these are important enough concepts that I'd leave them as ### rather than ##### (here and below)

Other changes in this PR look good though

twmb:
Made the changes in https://github.com/twmb/franz-go/pull/268, closing this PR (getting ready to release imminently)

PR #270, opened 2022-12-01
Fix missing codecType

PleasingFungus:
Oversight in #266.

### DIFF
--- a/pkg/kgo/compression.go
+++ b/pkg/kgo/compression.go
@@ -27,7 +27,7 @@ var sliceWriters = sync.Pool{New: func() interface{} { r := make([]byte, 8<<10);
 type codecType int8
 
 const (
-	codecNone = iota
+	codecNone codecType = iota
 	codecGzip
 	codecSnappy
 	codecLZ4
@@ -121,6 +121,7 @@ out:
 				}
 			}
 			c.gzPool = sync.Pool{New: func() interface{} { c, _ := gzip.NewWriterLevel(nil, level); return c }}
+		case codecSnappy: // (no pool needed for snappy)
 		case codecLZ4:
 			level := codec.level
 			if level < 0 {


pkg/kgo/compression.go
```diff
@@ -137,6 +137,7 @@ out:
 			}
 			w.Close()
 			c.lz4Pool = sync.Pool{New: fn}
+		case codecSnappy:
```
twmb, line 140: I'd prefer this be above lz4, since it's 2 (0 none, 1 gzip, 2 snappy)

But, this case doesn't exist currently because it's an unneeded case -- there is no pool no initialize. I'd either drop the case, or add a one line comment for why the case exists: `// no pool needed for snappy`

PleasingFungus: Sure, I can add a comment. I added this line to satisfy your linter - it was [lamenting](https://github.com/twmb/franz-go/actions/runs/3595872490) `missing cases in switch of type codecType: codecSnappy (exhaustive)`.

twmb:
Mind squashing the three commits into one, `kgo codec: fix missing codecType`?

Also this can be merged but wait until the next release

pkg/kgo/compression.go
twmb: interesting, perhaps that linter is triggering now because this is a dedicated type

PleasingFungus:
Sure, done. I also tweaked the comment back toward your original suggestion, since on further consideration, having a comment saying 'this line exists to silence the linter' feels like a bad code smell. (If we really just wanted to silence the linter, we could add a comment disabling that check for this function.)

twmb:
awesome, thank you!

PleasingFungus:
No problem. Thanks for being responsive! :)

PR #272, opened 2022-12-06
group balancer: support KIP-792

twmb:
This improves problems in the face of re-joining consumer zombies.


twmb:
Depends on #304

Issue #213, opened 2022-10-04
Hook interfaces don't pass context


brennan-airtime:
I see there is another issue that talks about an otel plugin. I don't know what the extent of that is but some thoughts after making my own proxy on top of kgo to add tracing.

I also understand the exact changes I'm suggesting would break the api, I'm more just thinking out loud.

```go
type HookProduceRecordBuffered interface {
	OnProduceRecordBuffered(*Record)
}
```

If we take producing a record for example, the hook is not passed the context the record was published with which means it cannot be used for opentelemetry trace context or baggage propagation. If the hook were passed the Context the text map propagator could be used to serialize the trace context into the headers of the record before publishing.

Consuming is trickier as the api doesn't allow passing back a different context when the record is processed:

```go
func (cl *Client) PollFetches(ctx context.Context) Fetches
```

But perhaps an api that is callback focused would make it possible and would require new hooks to help drive the callbacks to wrap them in tracing.

```go
func (cl *Client) PollFetches(ctx context.Context, func (ctx context.Context, Fetches))
// But each message could be part of a different trace so really we need to invoke a callback per record
func (cl *Client) PollFetches(ctx context.Context, func (ctx context.Context, record *Record, err FetchError))
```

brennan-airtime:
Oh I see an open PR that stuffs the context into the Record. That would work ok for producing.

On the consumer side it would still mean manually instrumented tracing to track processing the records but maybe that is ok. I don't have a good sense for how much of that would be heavily application specific.

yianni:
@brennan-airtime ,  Here's how we propagate tracing context in the kotel plugin:

[github.com/yianni/franz-go/blob/kotel-refactor/plugin/kotel/tracing.go](https://github.com/yianni/franz-go/blob/kotel-refactor/plugin/kotel/tracing.go)

brennan-airtime:
@yianni Ah ok so you're using the `Ctx` stuffed into the record.

Then I suppose to create the `${r.Topic} process` spans I instrument that in my application code by pulling out the record.Ctx that was populated by the fetch hook.  Which honestly feels about right given all the different ways I could choose to process a batch of fetched records.

twmb:
I'll be merging #201 today (adding ctx to record), but 1.8 is still a bit off pending other features (I'm a bit slow for closing the loop on some of these)

twmb:
Merged -- agreed with the discussion above, the fetch hook can be used to initialize / populate the record.Context and it's about the best way to do it.

[...]

PR #294, opened 2022-12-19
Add Kotel plugin

yianni:
This PR adds OpenTelemetry support to Franz-go as a plugin, allowing for the export of traces and metrics using OpenTelemetry. We have made every effort to adhere to the Open Telemetry specification and follow conventions for span names, attributes, and other elements, as outlined in the [Messaging Semantic Conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md).

Our main goal is to track the movement of requests from their source to their destination. To do this, we create three spans for traces, one of which is optional. The OnProduceRecordBuffered hook represents the send span, while the OnProduceRecordUnbuffered hook completes the send span and records any errors. The OnFetchRecordBuffered hook represents the receive span, and may also create a log-append span. This log-append span is a debatable feature but was requested by our team.

Both producer and consumer clients can use tracing contexts for context propagation. For example, a producer client may instrument an HTTP request and track it as it moves through Kafka, while a consumer client can extract tracing contexts from records and propagate them downstream to Kafka or other systems. Consumers can also add process spans after consuming records.

For metrics, we have followed the same logic used in the existing metric plugin for metrics.

We hope that these changes will improve the functionality of Franz-go and provide valuable insights through the use of OpenTelemetry. We have made every effort to ensure that these changes adhere to the Open Telemetry specification and follow best practices. If you have any questions or concerns, please don't hesitate to let us know. We look forward to your approval of this PR.


plugin/kotel/carrier_test.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/twmb/franz-go/pkg/kgo"
+)
+
+func TestNewRecordCarrier(t *testing.T) {
+	type args struct {
+		record *kgo.Record
+	}
+	tests := []struct {
+		name string
+		args args
```
twmb, line 16: Why not `record *kgo.Record`, and remove args & the args type?

Same thought below, w.r.t. the fields type as well -- although I think I know why args and fields exist, it seems a bit unnecessary here and is adding a bit of visual noise that's a bit difficult to parse

plugin/kotel/metrics.go
```diff
@@ -0,0 +1,221 @@
...
+
+// Hooks ---------------------------------------------------------------------
+
+func (m *Meter) OnBrokerConnect(meta kgo.BrokerMetadata, _ time.Duration, _ net.Conn, err error) {
+	node := strnode(meta.NodeID)
+	if err != nil {
+		m.metrics.connectErrs.Add(context.Background(), 1, attribute.String("node_id", node))
+		return
+	}
+	m.metrics.connects.Add(context.Background(), 1, attribute.String("node_id", node))
```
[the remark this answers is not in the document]
yianni: Synchronous measurements can have an associated trace context.

"Synchronous instruments are called in a request context, meaning they potentially have an associated tracing context and distributed correlation values. Multiple metric events may occur for a synchronous instrument within a given collection interval. Note that synchronous instruments may be called outside of a request context, such as for background computation. In these scenarios, we may simply consider the Context to be empty.." [1]

[1] https://github.com/open-telemetry/oteps/blob/main/text/metrics/0098-metric-instruments-explained.md#sync-vs-async-instruments

plugin/kotel/metrics.go
```diff
@@ -0,0 +1,221 @@
...
+	bufferedFetchRecords   int64
+}
+
+func (m *Meter) NewMetrics() Metrics {
+	// connects and disconnects
+
+	connects, _ := m.meter.SyncInt64().Counter(
+		"connects_total",
+		instrument.WithUnit(unit.Dimensionless),
+		instrument.WithDescription("Total number of connections opened, by metrics"),
```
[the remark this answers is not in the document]
yianni: It's supposed to be `, by broker`,  so this is a lousy copy-paste from the Prometheus plugin.

plugin/kotel/metrics.go
```diff
@@ -0,0 +1,221 @@
...
+
+	bufferedProduceRecords int64
+	bufferedFetchRecords   int64
+}
+
+func (m *Meter) NewMetrics() Metrics {
+	// connects and disconnects
+
+	connects, _ := m.meter.SyncInt64().Counter(
+		"connects_total",
```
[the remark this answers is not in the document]
yianni: It looks like the naming conventions for metrics in otel differ from those in Prometheus. According to the OpenTelemetry Specification [1], it is recommended to nest metrics into a hierarchy, such as `messaging.kafka.client`. Additionally, the OTEL specification provides general guidelines and instrument naming conventions [2] that we can follow. Let me know if you have any suggestions or thoughts on this."

[1] https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/README.md#general-guidelines
[2] https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/README.md#instrument-naming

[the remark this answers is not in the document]
yianni: Here are some real-life examples from instrumentation libraries:
cassandra[1]
http[2] 
runtime[3]

[1] https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/github.com/gocql/gocql/otelgocql/instrument.go#L55:L89
[2] https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/common.go#L34:L37
[3] https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/runtime/runtime.go#L206:L285

plugin/kotel/go.mod
```diff
@@ -0,0 +1,26 @@
...
+	github.com/klauspost/compress v1.15.13 // indirect
+	github.com/kr/pretty v0.1.0 // indirect
+	github.com/pierrec/lz4/v4 v4.1.17 // indirect
+	github.com/pmezard/go-difflib v1.0.0 // indirect
+	github.com/twmb/franz-go/pkg/kmsg v1.2.0 // indirect
+	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
+)
+
+replace github.com/twmb/franz-go => ../..
```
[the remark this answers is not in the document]
yianni: I guess that will break it from running until the module is published

plugin/kotel/carrier.go
```diff
@@ -0,0 +1,52 @@
...
+type RecordCarrier struct {
+	record *kgo.Record
+}
+
+// NewRecordCarrier creates a new RecordCarrier.
+func NewRecordCarrier(record *kgo.Record) RecordCarrier {
+	return RecordCarrier{record: record}
+}
+
+// Get retrieves a single value for a given key.
```
twmb, line 19: nit: `for a given key if it exists.`

plugin/kotel/carrier_test.go
```diff
@@ -0,0 +1,132 @@
...
+		val string
+	}
+
+	tests := []struct {
+		name   string
+		record *kgo.Record
+		kv1    kv
+		kv2    kv
+		kv3    kv
+		kv4    kv
```
twmb, line 105: This is a bit of an odd test since there's only one case, but, I'd change this to `kvs []kv` and get rid of kv1 through kv4, and then in the actual logic, range over this rather than calling Set 4 times

plugin/kotel/metrics.go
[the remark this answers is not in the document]
twmb: kk, sounds good -- looks like this PR should switch to some dot format. I'm not sure what type of namespacing would be best here -- do y'all want to take on an initial stab at namespacing then I can review?

plugin/kotel/metrics.go
```diff
@@ -0,0 +1,217 @@
...
+)
+
+type Meter struct {
+	provider metric.MeterProvider
+	meter    metric.Meter
+	metrics  Metrics
+}
+
+// MetricsOption interface used for setting optional config properties.
+type MetricsOption interface {
```
twmb, line 36: To be a bit more consistent with other areas of franz-go, can `Option` be renamed `Opt`?

Also, since this is configuring a `Meter`, should this be `MeterOpt`?

plugin/kotel/version.go
```diff
@@ -0,0 +1,21 @@
...
+				return dep.Version
+			}
+		}
+	}
+	return "unknown"
+}
+
+// SemVersion is the semantic version to be supplied to tracer/meter creation.
+func SemVersion() string {
+	return "semver:" + Version()
```
twmb, line 20: Is this `semver:` prefix standard for otel? I'm wondering why SemVersion is needed / exists, vs. just using Version.

plugin/kotel/tracing.go
```diff
@@ -0,0 +1,162 @@
...
+
+// Hooks ---------------------------------------------------------------------
+
+// OnProduceRecordBuffered Begins the producer span
+func (t *Tracer) OnProduceRecordBuffered(r *kgo.Record) {
+	attrs := []attribute.KeyValue{
+		semconv.MessagingSystemKey.String("kafka"),
+		semconv.MessagingDestinationKindTopic,
+		semconv.MessagingDestinationKey.String(r.Topic),
+		semconv.MessagingKafkaClientIDKey.String(strconv.FormatInt(r.ProducerID, 10)),
```
twmb, line 88: Is this meant to be the actual client ID? Looking at the otel docs, I think so -- if so, the client ID can / should be an option to the tracer

(I admit it is ugly that you'll need to specify the client ID in two locations)

plugin/kotel/tracing.go
```diff
@@ -0,0 +1,162 @@
...
+		semconv.MessagingMessageIDKey.String(strconv.FormatInt(r.Offset, 10)),
+		semconv.MessagingKafkaPartitionKey.Int64(int64(r.Partition)),
+	)
+
+	if err != nil {
+		span.SetStatus(codes.Error, err.Error())
+		span.RecordError(err)
+	}
+
+	span.End()
```
twmb, line 118: defer span.End() at the top?

plugin/kotel/tracing.go
```diff
@@ -0,0 +1,162 @@
...
+
+	r.Context = spanContext
+}
+
+// OnProduceRecordUnbuffered Completes the producer span
+func (t *Tracer) OnProduceRecordUnbuffered(r *kgo.Record, err error) {
+	span := trace.SpanFromContext(r.Context)
+
+	span.SetAttributes(
+		semconv.MessagingMessageIDKey.String(strconv.FormatInt(r.Offset, 10)),
```
twmb, line 109: Why no destination key here, but there is one above?

plugin/kotel/tracing.go
```diff
@@ -0,0 +1,162 @@
...
+	// Optionally add the LogAppend span if the record timestamp is set in Kafka
+	if r.Attrs.TimestampType() == 1 {
+		opts := []trace.SpanStartOption{
+			trace.WithTimestamp(r.Timestamp),
+			trace.WithAttributes(attrs...),
+			trace.WithSpanKind(trace.SpanKindConsumer),
+		}
+
+		nextSpanContext, nextSpan = t.tracer.Start(ctx, fmt.Sprintf("%s logappend", r.Topic), opts...)
+		nextSpan.End()
```
twmb, line 150: Should this fall into the below logic? I'm wondering why there are two spans here.

plugin/kotel/version.go
yianni: We are following the same versioning pattern used in other OpenTelemetry libraries as can be seen in the link provided [1].

[1] https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-go-contrib%20semversion&type=code

plugin/kotel/tracing.go
yianni: If we are to proceed with this approach, we should also consider adding the consumer group name and any other necessary information to the options to fully comply with the messaging specification outlined in [1]. This will ensure consistency and compliance with the standard. 

[1] https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#apache-kafka

plugin/kotel/tracing.go
yianni: Based on my understanding, since the span already exists and we are later referencing and closing it, the original attributes are inherited. This allows us to later add any additional data returned from Kafka to these existing attributes.

plugin/kotel/metrics.go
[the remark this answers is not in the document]
yianni: Great, we'll work on figuring this out. Additionally, here's a useful reference to follow [1].

[1] https://opentelemetry.io/docs/reference/specification/metrics/semantic_conventions/instrumentation/kafka/

plugin/kotel/tracing.go
yianni: As per a requirement, we need to track events both when they land in Kafka (conditional log append span) and when the message is received by the client (aka receive span). This will give us a complete picture of message flows, and a better understanding of how the system is working.

Essentially these two spans record different timestamps.  Here is a screenshot to visualize this

Kafka tracing

I've brought up the log append span as a debatable aspect of this implementation in the PR introduction. @brunsgaard has more knowledge and experience on this, would be valuable to hear their thoughts on the significance of it, and why it is beneficial to have it as part of this implementation.

However, if it's not deemed important, I am open to the idea of removing the log append span, and potentially achieving a similar goal by passing the `r.Timestamp` to the receive span attributes, to track the time when the message was received. However, I don't know whether this would cause a cardinality explosion on the metrics database.

Please let me know your thoughts on this.

twmb: I see, this makes sense -- I need to read the otel docs a bit to understand why these spans are being ended immediately (unless you have a quick tldr), but I see the point of tracking when they landed on the broker. +1 to keeping this.

yianni: We want to end the receive span before starting the process span. Also, if we end the receive span in the hook (hidden away from users), we can guarantee that it will always be closed. Although, starting and ending the process span is up to the user. We are open to any suggestions.

yianni: P.S., the latest commit has removed the log append span to simplify the code further.

plugin/kotel/tracer.go
```diff
@@ -0,0 +1,183 @@
...
+	if t.clientID != "" {
+		attrs = append(attrs, semconv.MessagingKafkaClientIDKey.String(t.clientID))
+	}
+
+	opts := []trace.SpanStartOption{
+		trace.WithAttributes(attrs...),
+		trace.WithSpanKind(trace.SpanKindProducer),
+	}
+
+	spanContext, _ := t.tracer.Start(r.Context, fmt.Sprintf("%s send", r.Topic), opts...)
```
twmb, line 124: I'd switch the `fmt` to `r.Topic + " send"`, it's faster and potentially fewer allocs.

plugin/kotel/tracer.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	if t.clientID != "" {
+		attrs = append(attrs, semconv.MessagingKafkaClientIDKey.String(t.clientID))
+	}
+
+	if t.consumerGroup != "" {
+		attrs = append(attrs, semconv.MessagingKafkaConsumerGroupKey.String(t.consumerGroup))
+	}
+
```
twmb, line 169: as a very minor nit, I'd delete the spaces around these conditionals -- it adds a lot of vertical space to this function (similar thought in other functions).

optional comment though, this is more of a style preference comment

plugin/kotel/tracer.go
```diff
@@ -0,0 +1,183 @@
...
+	opts := []trace.SpanStartOption{
+		trace.WithAttributes(attrs...),
+		trace.WithSpanKind(trace.SpanKindConsumer),
+	}
+
+	if r.Context == nil {
+		r.Context = context.Background()
+	}
+	ctx := t.propagators.Extract(r.Context, NewRecordCarrier(r))
+	newCtx, span := t.tracer.Start(ctx, fmt.Sprintf("%s receive", r.Topic), opts...)
```
twmb, line 179: similar comment as in buffered -- I'd switch to `r.Topic + " receive"`, for speed and potentially fewer allocs

plugin/kotel/tracer.go
```diff
@@ -0,0 +1,183 @@
...
+	span.End()
```
twmb, line 180: Looking at this a bit and looking at [your comment here](https://github.com/twmb/franz-go/pull/294#discussion_r1083103977), I'd think this needs to go into a new hook that isn't in this PR -- `OnFetchRecordUnbuffered`. As is, this span isn't measuring anything besides the time it takes to call the `End` function?

plugin/kotel/README.md
```diff
@@ -0,0 +1,151 @@
...
+Please visit the  [OpenTelemetry documentation](https://opentelemetry.io/docs) for additional information about
+OpenTelemetry and how it can be used in your `franz-go` projects.
+
+## Tracing
+
+`kotel`'s tracing module allows you to track the lineage of `kgo.Record` objects through a series of `franz-go` hooks:
+
+1) HookProduceRecordBuffered
+2) HookProduceRecordUnbuffered
+3) HookFetchRecordBuffered
```
twmb, line 27: per my comment below this might need a `4) HookFetchRecordUnbuffered`

plugin/kotel/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+cl, err := kgo.NewClient(
+	kgo.WithHooks(kotelService.Hooks()...),
+    // ...other opts
+)
+```
+
+`kotel` enables you to enrich records with ancestor trace data, such as trace context from a `http.Request` object that
+may already contain a parent span from an instrumentation library. This can be useful for tracking the lineage of your
+records and providing additional context for tracing and debugging purposes.
```
twmb, line 57: This section duplicates the wording in the intro paragraph and the wording in the line right under the `Tracing` header. I'd drop these two sentences entirely since both points were already mentioned

plugin/kotel/README.md
```diff
@@ -0,0 +1,151 @@
...
+    wg.Add(1)
+    record := &kgo.Record{Topic: "my-topic", Value: []byte("bar")}
+    
+    cl.Produce(ctx, record, func(_ *kgo.Record, err error) {
+    defer wg.Done()
+    if err != nil {
+        fmt.Printf("record had a produce error: %v\n", err)
+        span.SetStatus(codes.Error, err.Error())
+        span.RecordError(err)
+    }
```
twmb, line 78: As a code sample, this bit is formatted incorrectly -- I'd switch to tabbing for this entire function and, more importantly, add a level of indentation here.

For simplification you _could_ switch to `ProduceSync(...).First()`, but I actually prefer the wg aspect usually

plugin/kotel/README.md
```diff
@@ -0,0 +1,151 @@
...
+    })
+    wg.Wait()
+    // ...
+}
+```
+
+`kotel` enables consumers to extract ancestor trace data from kafka records and continue the trace in downstream
+processing steps. This can be useful for providing additional context and visibility into the processing of records.
```
twmb, line 86: Similar thought to the above one, I think the _next_ line is much more informative and this line can be dropped.

plugin/kotel/README.md
```diff
@@ -0,0 +1,151 @@
...
+	if errs := fetches.Errors(); len(errs) > 0 {
+		panic(fmt.Sprint(errs))
+	}
+
+	iter := fetches.RecordIter()
+	for !iter.Done() {
+		record := iter.Next()
+		ctx, span := tracer.Start(r.Context, fmt.Sprintf("%s process", r.Topic))
+		span.SetAttributes(attribute.String("baz", "qux"))
+		span.End()
```
twmb, line 105: It may be good to add a `// process record here` before the `span.End` line, to show that the span is meant to be wrapped around processing.

plugin/kgmetrics/kgmetrics.go
```diff
@@ -18,7 +18,7 @@
 //
 // This can be used in a client like so:
 //
-//	m := kgmetrics.NewMetrics()
+//	m := kgmetrics.NewInstruments()
```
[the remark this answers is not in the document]
yianni: Thanks , this is my mistake

plugin/kotel/tracer.go
yianni: Yeah, that's right. We previously looked into the `OnFetchRecordUnbuffered`. When adding that hook and running our example, the hook was never called. We didn't look into it further.

twmb: I see why now, I think it's a bug in `kgo`. This is both a buffered and unbuffered hook, and the hook exits on the first type switch. I'll patch.

twmb: Patched in v1.11.2, PTAL

plugin/kotel/meter.go
```diff
@@ -0,0 +1,246 @@
...
+	return meterOptFunc(func(m *Meter) {
+		if provider != nil {
+			m.provider = provider
+		}
+	})
+}
+
+// Instruments -------------------------------------------------------------------
+
+type Instruments struct {
```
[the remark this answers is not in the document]
yianni: I'm okay with keeping the "metrics" name if it's clearer. I got the "instruments" terminology from the metrics API specs[1], and it states:

The Metrics API consists of these main components:

    [MeterProvider](https://opentelemetry.io/docs/reference/specification/metrics/api/#meterprovider) is the entry point of the API. It provides access to Meters.
    [Meter](https://opentelemetry.io/docs/reference/specification/metrics/api/#meter) is the class responsible for creating Instruments.
    [Instrument](https://opentelemetry.io/docs/reference/specification/metrics/api/#instrument) is responsible for reporting [Measurements](https://opentelemetry.io/docs/reference/specification/metrics/api/#measurement).
```

[1] https://opentelemetry.io/docs/reference/specification/metrics/api/

**Comment by yianni** [comment]
@twmb, Most of the requested changes, have been made; however, I am still considering the naming for the metrics and will update that later. Feel free to let me know if there's anything else I can do in the meantime.

File: plugin/kotel/meter_test.go [hunk 25]
```diff
@@ -0,0 +1,39 @@
...
+		want *Meter
+	}{
+		{
+			name: "With MeterProvider",
+			opts: []MeterOpt{MeterProvider(provider)},
+			want: &Meter{
+				provider: provider,
+				meter: provider.Meter(
+					instrumentationName,
+					metric.WithInstrumentationVersion(semVersion()),
```
twmb, line 26: ... sorry for the back and forth here but, does `SemVersion` actually need to be public to end users? I don't know myself

plugin/kotel/meter.go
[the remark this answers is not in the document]
twmb: +1 to keeping instruments, just looking for what's standard in the otel world

plugin/kotel/meter_test.go
yianni: End-users may not need `SemVersion` as it's mainly used for internal tracking and tagging of the traces/metrics. Do you have a use case in mind?

Semi-related: end-users will create providers to create meters and tracers. The same providers would most likely be passed in kotel, as we demonstrated in the example. Should these meters and tracers be exposed so things outside kotel are included in the "Instrumentation Library"? This decision leads to other things such as providing predefined process tracers with defaults as asked by end-users https://github.com/twmb/franz-go/issues/213#issuecomment-1269267559.

Currently, only the meters and tracers created in kotel and started in the hooks are considered part of the library. Please take a look at the trace export to understand what I'm saying.

```
{
        "Name": "topic send",
        ...
        "InstrumentationLibrary": {
                "Name": "github.com/twmb/franz-go/plugin/kotel",
                "Version": "semver:v0.0.0-00010101000000-000000000000",
                "SchemaURL": "https://opentelemetry.io/schemas/1.12.0"
        }
}
{
        "Name": "request",
        ...
        "InstrumentationLibrary": {
                "Name": "request-service",
                "Version": "",
                "SchemaURL": ""
        }
}
{
        "Name": "topic receive",
        ...
        "InstrumentationLibrary": {
                "Name": "github.com/twmb/franz-go/plugin/kotel",
                "Version": "semver:v0.0.0-00010101000000-000000000000",
                "SchemaURL": "https://opentelemetry.io/schemas/1.12.0"
        }
}
{
        "Name": "topic process",
        ...
        "InstrumentationLibrary": {
                "Name": "process-service",
                "Version": "",
                "SchemaURL": ""
        }
}
```

twmb: > Do you have a use case in mind?

No use case. If it's not obviously needed by end users, then I err on keeping it private for now until there's a need.

> Should these meters and tracers be exposed so things outside kotel are included in the "Instrumentation Library"?

What would the API look like? Not following from the example but I think an API would be a bit clearer.

yianni: Based on my understanding, the `semVersion` is currently set to private.


### 1) Idea 1 - Clients use kotel.Tracer

To expose the Kotel tracer, we can make the Tracer field public. This will allow users to create spans using the Kotel tracer like so:

```
kotelTracer := kotel.NewTracer([]kotel.TracerOpt{}...)
ctx, span := kotelTracer.Tracer.Start(context.Background(),"awsomeSpan", []trace.SpanStartOption{}...)
```

### 2) Idea 2 - Add WithProcessTracer Method to kotel.Tracer

We can also add a new method `WithProcessTracer` to the Tracer struct, to provide users with a tracer that is up to the Otel messaging specification.

```
func (t *Tracer) WithProcessTracer(r *kgo.Record) (context.Context, trace.Span) {
	// Set up the span options.
	attrs := []attribute.KeyValue{
		semconv.MessagingSystemKey.String("kafka"),
		semconv.MessagingDestinationKindTopic,
		semconv.MessagingDestinationKey.String(r.Topic),
		semconv.MessagingOperationProcess,
		semconv.MessagingKafkaPartitionKey.Int64(int64(r.Partition)),
	}
	if r.Key != nil {
		attrs = append(attrs, semconv.MessagingKafkaMessageKeyKey.String(string(r.Key)))
	}
	if t.clientID != "" {
		attrs = append(attrs, semconv.MessagingKafkaClientIDKey.String(t.clientID))
	}
	if t.consumerGroup != "" {
		attrs = append(attrs, semconv.MessagingKafkaConsumerGroupKey.String(t.consumerGroup))
	}
	opts := []trace.SpanStartOption{
		trace.WithAttributes(attrs...),
		trace.WithSpanKind(trace.SpanKindConsumer),
	}

	if r.Context == nil {
		r.Context = context.Background()
	}
	// Start a new span using the provided context and options.
	return t.tracer.Start(r.Context, r.Topic+" process", opts...)
}
```

It would be called like so:

```
func processRecord(record *kgo.Record, tracer kotel.Tracer) {
	_, span := tracer.ProcessTracer(record)
	// Simulate some work.
	time.Sleep(1 * time.Second)
	// End the span when function exits.
	defer span.End()
	fmt.Printf(
		"processed offset '%s' with key '%s' and value '%s'\n",
		strconv.FormatInt(record.Offset, 10),
		string(record.Key),
		string(record.Value),
	)
}
```

### Conclusion:
In conclusion, for idea 2, If we want to encapsulate all 3 spans required from the Otel messaging spec, then the simplest way I can think of is creating a WithProcessTracer method. It will save our users a step from creating this in their own code and guarantees all the attributes are set correctly. However, for idea 1, we should not do this and leave it as is. It makes sense to separate the boundary between internal and outside instrumentation. Users can continue to create their tracers, and their applications using their instrumentation library.

Let me know what you think about idea 2. I can make this change in the PR.

twmb: I like the second idea more -- if I'm reading it correctly, it allows us to do things the right way and hide away complexity from the user. Is there a reason people might want the control of the first option?

yianni: No, they can create their own tracers. :)

yianni:
I've added the process span method. Further, the readme is a lot simpler. I had to go into some detail to explain the flow of spans since it's a little complicated. 

The last thing is to finalize the metrics naming. I would like to have the pr done by the end of the day.

brennan-airtime:
I'm not sure if this is safe to do `semconv.MessagingKafkaMessageKeyKey.String(string(r.Key))`.

Since otel strings are ultimately transferred over a protobuf I imagine the protobuf restrictions on strings is to be respected:

https://developers.google.com/protocol-buffers/docs/proto3
> A string must always contain UTF-8 encoded or 7-bit ASCII text, and cannot be longer than 2^32.

The kafka record key can be arbitrary binary data, not necessarily valid utf8.

I've had to do a similar check when logging in `zap`

```
var key zapcore.Field
if utf8.Valid(record.Key) {
	key = zap.String("key", string(record.Key))
} else {
	key = zap.Binary("key", record.Key)
}
```

twmb:
Interesting. In that case, it sounds like this should do something similar to encoding/json: `utf8.DecodeRune` until the input slice is exhausted, appending to a temporary `[]rune` and replacing invalid characters with `\ufffd`. To avoid too much garbage (since per-record hooks are already a 10% perf hit), you could check `utf8.Valid` first.

twmb:
@yianni lmk if you'd like to change the naming or merge as is

yianni:
@twmb I'm okay to merge as is

twmb:
Sounds good to me. Worst case, anybody interested can:
* add a new option to rename some of the metrics / drop `.count` if they so desire
* add an option to hex encode / utf8 replacement character replace invalid utf8

twmb:
@yianni I've pushed `plugin/kotel/v1.0.0` for this, thank you for the great PR and the patience on the long comment chains and sometimes too-nitpicky comments!

PR #298, opened 2023-01-02
Add create topic admin function

### LINKED ISSUE #297: Simpler function for creating a single topic in admin client
**Problem**

Creating a single topic is possibly one of the most popular functions used in the franz-go admin client. Currently there's only a function for creating multiple topics with the same configuration which may be handy. However, if you only want to create one topic and check if that was successful the required error checking remind me of using the `kmsg` package. I don't like the fact that I have to check for an error at multiple places to figure out whether my operation was successful or not:

```go
		createTopicResponses, err := k.admCl.CreateTopics(
			ctx,
			1,
			1,
			map[string]*string{
				"cleanup.policy": kadm.StringPtr("compact"),
				"segment.bytes":  kadm.StringPtr("8388608"), // 8 MiB
			},
			k.cfg.kafkaTopicName,
		)
		if err != nil {
			return fmt.Errorf("failed to create topic: %w", err)
		}

		topicResponse, exists := createTopicResponses[k.cfg.kafkaTopicName]
		if !exists {
			return fmt.Errorf("no topic response for the requested topic creation request")
		}
		if topicResponse.Err != nil {
			return fmt.Errorf("failed to create topic: %w", err)
		}
```

**Proposal:**

I think there should be an API that offers the creation of a single topic:

```go
// CreateTopic issues a create topics request with the given partitions,
// replication factor, and (optional) configs for the given topic name. Under the hood,
// this uses the default 15s request timeout and lets Kafka choose where to
// place partitions.
//
// This package includes a StringPtr function to aid in building config values.
//
// If the topic could not be created this function will return an error. An error may be returned
// due to authorization failure, a failed network request, a missing controller or other issues.
func (cl *Client) CreateTopic(
	ctx context.Context,
	partitions int32,
	replicationFactor int16,
	configs map[string]*string,
        topic string) (CreateTopicResponse, error) {}
```

**Note:** While this looks very similar to the `CreateTopics` API - just for a single topic - the error handling here is more opinionated as errors reported on the Kafka API (for each individual topic) will also be returned as plain errors. This does not make sense for the `CreateTopics` method as partial failure/success may be an option. To make this difference clearer one could also consider naming this `CreateTopicSimple()`

weeco:
Fixes #297

### DIFF
--- a/pkg/kadm/topics.go
+++ b/pkg/kadm/topics.go
@@ -2,6 +2,7 @@ package kadm
 
 import (
 	"context"
+	"errors"
 	"sort"
 
 	"github.com/twmb/franz-go/pkg/kerr"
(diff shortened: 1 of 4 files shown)


pkg/kadm/topics.go
```diff
@@ -87,10 +88,53 @@ func (rs CreateTopicResponses) On(topic string, fn func(*CreateTopicResponse) er
...
+//
+// If the topic could not be created this function will return an error. An error may be returned
+// due to authorization failure, a failed network request, a missing controller or other issues.
+func (cl *Client) CreateTopic(
+	ctx context.Context,
+	partitions int32,
+	replicationFactor int16,
+	configs map[string]*string,
+	topic string,
+) (CreateTopicResponse, error) {
```
weeco, line 110: The `CreateTopicResponse` struct has an error property which we already check as part of the function body and return it as the second argument (error) if it is not nil. 

Should we create a separate struct for this `CreateTopic` response to make clear that the user does not have to check the `Err` property?

twmb: I think we can just document this function fully that it returns CreateTopicResponse.Err, so if they check the return, they do not need to check the field.

pkg/kadm/topics.go
```diff
@@ -87,10 +88,53 @@ func (rs CreateTopicResponses) On(topic string, fn func(*CreateTopicResponse) er
...
+	}
+
+	return response, response.Err
+}
+
 // CreateTopics issues a create topics request with the given partitions,
 // replication factor, and (optional) configs for every topic. Under the hood,
 // this uses the default 15s request timeout and lets Kafka choose where to
 // place partitions.
+// Version 4 of the underlying create topic request was introduced in Kafka 2.4.0 and
```
twmb, line 134: Extra line above this, for paragraph separation

pkg/kadm/topics.go
```diff
@@ -87,10 +88,53 @@ func (rs CreateTopicResponses) On(topic string, fn func(*CreateTopicResponse) er
 	return CreateTopicResponse{}, kerr.UnknownTopicOrPartition
 }
 
+// CreateTopic issues a create topics request with the given partitions,
+// replication factor, and (optional) configs for the given topic name. Under the hood,
+// this uses the default 15s request timeout and lets Kafka choose where to
+// place partitions.
+// Version 4 of the underlying create topic request was introduced in Kafka 2.4.0 and
```
twmb, line 95: Extra line above this, for paragraph separation.

pkg/kadm/topics.go
```diff
@@ -87,10 +88,53 @@ func (rs CreateTopicResponses) On(topic string, fn func(*CreateTopicResponse) er
 	return CreateTopicResponse{}, kerr.UnknownTopicOrPartition
 }
 
+// CreateTopic issues a create topics request with the given partitions,
+// replication factor, and (optional) configs for the given topic name. Under the hood,
+// this uses the default 15s request timeout and lets Kafka choose where to
+// place partitions.
```
twmb, line 94:
```
place partitions. This function exists to complement CreateTopics,
// making the single-topic creation case easier to handle.
```
wdyt?

PR #299, opened 2023-01-02
KIP-405 proto / kerr bumps

twmb:
This did not make Kafka 3.4, this will wait until Kafka 3.5.


twmb:
Closing until Kafka does a release so this stops hanging around.

PR #314, opened 2023-01-22
kgo: patch HookFetchRecordUnbuffered

twmb:
Previously, if a hook implemented both HookFetchRecordBuffered and HookFetchRecordUnbuffered, then HookFetchRecordUnbuffered would never be called. The type switch in this logic would always match the buffered case and never call unbuffered.

Reorganizing the conditionals a little bit ensures that the correct hook is always called.


twmb:
Tested with a hook that prints when things are buffered and unbuffered.

PR #320, opened 2023-01-23
add CommitMarkedOffsets when using a custom onRevoked and AutoCommitMarks

celrenheit:
Hello,

When using a goroutine per partition with autocommit marks like in the provided example, the committed offsets are not the ones marked but the ones polled. This is because we issue a commit of uncommitted offsets on revoke [here](https://github.com/twmb/franz-go/blob/9d4480e85ca28b8cff03f78594acc3259b1d839e/examples/goroutine_per_partition_consuming/autocommit_marks/main.go#L78), uncommitted being `Uncommitted offsets are always updated on calls to PollFetches.`. 
Since we are consuming multiple polls (thanks to the [bufferred channel](https://github.com/twmb/franz-go/blob/9d4480e85ca28b8cff03f78594acc3259b1d839e/examples/goroutine_per_partition_consuming/autocommit_marks/main.go#L68)) and marking them asynchronously, if processing fails or is stopped, the uncommitted offsets are already updated by the different polls and we will skip messages.

The default on revoke callback is [not impacted](https://github.com/twmb/franz-go/blob/9d4480e85ca28b8cff03f78594acc3259b1d839e/pkg/kgo/consumer_group.go#L2533) by this.

This PR adds a `CommitMarkedOffsets` method in order to reduce the boilerplate of using `CommitOffsetsSync(ctx, cl.MarkedOffsets(), etc..)` when using `AutoCommitMarks`.

I hope that I am not missing something here.

### DIFF
--- a/pkg/kgo/consumer_group.go
+++ b/pkg/kgo/consumer_group.go
@@ -2331,25 +2331,11 @@ func (cl *Client) MarkCommitRecords(rs ...*Record) {
 // Retriable errors are retried up to the configured retry limit, and any
 // unretriable error is returned.
 //
-// This function is useful as a simple way to commit offsets if you have
-// disabled autocommitting. As an alternative if you want to commit specific
-// records, see CommitRecords.
-//
-// Simple usage of this function may lead to duplicate records if a consumer
-// group rebalance occurs before or while this function is being executed. You
-// can avoid this scenario by calling CommitRecords in a custom
-// OnPartitionsRevoked, but for most workloads, a small bit of potential
-// duplicate processing is fine. See the documentation on DisableAutoCommit
-// for more details. You can also avoid this problem by using
-// BlockRebalanceOnCommit, but that option comes with its own tradeoffs (refer
-// to its documentation).
-//
 // The recommended pattern for using this function is to have a poll / process
 // / commit loop. First PollFetches, then process every record, then call
 // CommitUncommittedOffsets.
 //
-// If you do not want to wait for this function to complete before continuing
-// processing records, you can call this function in a goroutine.
+// As an alternative if you want to commit specific records, see CommitRecords.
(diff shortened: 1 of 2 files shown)


pkg/kgo/consumer_group.go
```diff
@@ -2352,8 +2352,43 @@ func (cl *Client) MarkCommitRecords(rs ...*Record) {
...
+
+// CommitMarkedOffsets issues a synchronous offset commit for any
+// partition that has been consumed from that has marked offsets.
+// Retriable errors are retried up to the configured retry limit, and any
+// unretriable error is returned.
+//
+// This function is useful as a simple way to commit offsets if you have
+// marked offsets with MarkCommitRecords when using AutoCommitMarks.
+// As an alternative if you want to commit specific
+// records, see CommitRecords.
```
twmb, line 2366: nit: formatting of this block  seems to have odd wrapping. (I usually just wrap to 80col with vim with `gwap`)

twmb (approved):
Makes sense, this probably should have been added when `MarkedOffsets` was added. Can the commit message be changed to something like,
```
pkg/kgo: add Client.CommitMarkedOffsets
```
?

I wish there were a way not to duplicate this documentation for a third time, perhaps it should just refer to documentation on CommitRecords (same goes for CommitUncommittedOffsets), but that can be in a follow up commit if ever.

pkg/kgo/consumer_group.go
celrenheit: indeed

celrenheit: this should fixed now

celrenheit:
I tweaked the documentation, tell me if that looks good to you.

twmb:
lgtm, I'm going to wait on merging this until more features become a candidate for a v1.12 release.

PR #361, opened 2023-02-16
sticky: optimize more

twmb:
* Uses some unsafe tricks to reduce some allocations in the sticky package
* Extensively documents the reasoning behind the unsafe
* Decodes userdata directly to drop pre-existing member allocations a good amount


twmb:
Technically the unsafe trick is not explicitly documented as safe, closing for now until I learn / realize it is.

PR #367, opened 2023-02-21
check hooks and flatten slices of hooks

### LINKED ISSUE #365: WithHooks() can fail in non-obvious ways
Since `kgo.Hook` is an empty interface anything implements it, even `[]kgo.Hook`. When adding `kotel` I was surprised when the tracing wasn't firing at all because I didn't inline the slice.

```golang
// opps, []kgo.Hooks is valid but implements no hooks
kgo.WithHooks(kotelService.Hooks())
// good
kgo.WithHooks(kotelService.Hooks()...)
```

I'm wondering if it would be palatable to change `WithHooks` to inspect the passed in hooks and flatten the slice/s if it detects it was passed something that implements no individual hooks. The inspection would only be done at client construction time.

That way the following invocations are all equivalent.

```golang
var xs, ys []kgo.Hook
var someHook kgo.Hook
var all []kgo.Hook
for _, v := range xs {
  all = append(all, v)
}
for _, v := range ys {
  all = append(all, v)
}
all = append(all, someHook)

kgo.WithHooks(all)
kgo.WithHooks(all...)
kgo.WithHooks(xs, ys, someHook)
kgo.WithHooks(xs..., ys..., someHook)
```

brennan-airtime:
Since `kgo.Hook` is an empty interface `[]kgo.Hook` also satisfies the empty interface which leads to silent failures if not passed to `kgo.WithHooks` properly.

```golang
kgo.WithHooks(hooks...) // works
kgo.WithHooks(hooks) // nothing called
```

By doing a little extra work inside of the config validation these mistakes can be fixed up or be turned into runtime errors.

Closes #365.

### DIFF
--- a/pkg/kgo/config.go
+++ b/pkg/kgo/config.go
@@ -388,7 +388,34 @@ func (cfg *cfg) validate() error {
 		return errors.New("invalid group partition assigned/revoked/lost functions set when a group was not specified")
 	}
 
+	processedHooks, err := processHooks(cfg.hooks)
+	if err != nil {
+		return err
+	}
+	cfg.hooks = processedHooks
+
 	return nil
+}
+
+// processHooks will inspect and recusively unpack slices of hooks stopping
+// if the instance implements any hook interface. It will return an error on
+// the first instance that implements no hook interface
+func processHooks(hooks []Hook) ([]Hook, error) {
+	var processedHooks []Hook
+	for _, hook := range hooks {
+		if implementsAnyHook(hook) {
+			processedHooks = append(processedHooks, hook)
+		} else if moreHooks, ok := hook.([]Hook); ok {
+			more, err := processHooks(moreHooks)
+			if err != nil {
+				return nil, err
+			}
+			processedHooks = append(processedHooks, more...)
+		} else {
+			return nil, errors.New("found an argument that implements no hook interfaces")
+		}
+	}
+	return processedHooks, nil
 }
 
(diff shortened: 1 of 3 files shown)


pkg/kgo/hooks.go
```diff
@@ -356,3 +362,43 @@ type HookFetchRecordUnbuffered interface {
...
+		return true
+	case HookProduceRecordUnbuffered:
+		return true
+	case HookFetchRecordBuffered:
+		return true
+	case HookFetchRecordUnbuffered:
+		return true
+	default:
+		return false
+	}
```
twmb, line 403: This can be squashed a bit,
```
case HookNewClient,
  HookBrokerConnect,
  HookBrokerDisconnect,
  ..all other hooks
  HookFetchRecordUnbuffered:
    return true
}
return false
```

pkg/kgo/config.go
```diff
@@ -388,9 +388,36 @@ func (cfg *cfg) validate() error {
...
+	cfg.hooks = processedHooks
+
 	return nil
 }
 
+// processHooks will inspect and recusively unpack slices of hooks stopping
+// if the instance implements any hook interface. It will return an error on
+// the first instance that implements no hook interface
+func processHooks(hooks []Hook) ([]Hook, error) {
+	processedHooks := []Hook{}
```
twmb, line 404: nit: I generally prefer `var processedHooks []Hook` (only adding this nit because of the other comment)

twmb (approved):
three nit / style comments but the logic lgtm

twmb (approved):

twmb:
lgtm, will merge once I prepare a minor release.

PR #390, opened 2023-03-13
kotel remove use of deprecated unit package

brennan-airtime:


twmb:
I'll do this in #371, and also un-export the units there.

PR #401, opened 2023-03-15
Fix for segment benchmark

bgranvea:
This PR enables async commits to have a fair comparison with franz-go.

### DIFF
--- a/examples/bench/compare/segment/main.go
+++ b/examples/bench/compare/segment/main.go
@@ -137,6 +137,7 @@ func main() {
 			Brokers:         brokers,
 			Topic:           *topic,
 			ReadLagInterval: -1,
+			CommitInterval:  time.Second * 5,
 		}
 		if *group != "" {
 			cfg.GroupID = *group


examples/bench/compare/segment/main.go
```diff
@@ -137,6 +137,7 @@ func main() {
 			Brokers:         brokers,
 			Topic:           *topic,
 			ReadLagInterval: -1,
+			CommitInterval:  time.Second * 5,
```
twmb, line 140: This lgtm -- is this the same thing as using FetchMessages as you brought up in the issue?

bgranvea: With `FetchMessages` you must commit manually by calling `CommitMessages`. With `ReadMessages`, commit is managed automatically, at each message if `CommitInterval=0` (the default) or every `CommitInterval`.

So I think `ReadMessages + CommitInterval=5s` is the closest to the configuration used for franz-go in benchmark.

PR #408, opened 2023-03-23
kfake: simulate rebalancing of partitions

jan-g:
This permits tests using kfake to simulate the
movement of a partition leader from one broker to
another.

### DIFF
--- a/pkg/kfake/cluster.go
+++ b/pkg/kfake/cluster.go
@@ -1,6 +1,7 @@
 package kfake
 
 import (
+	"errors"
 	"fmt"
 	"net"
 	"sync"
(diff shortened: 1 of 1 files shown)


pkg/kfake/cluster.go
```diff
@@ -183,6 +186,10 @@ func (c *Cluster) run() {
 			creq = w.creq
 		case <-c.die:
 			return
+		case fn := <-c.adminCh:
+			// Run a custom request in the context of the cluster
+			fn()
```
twmb, line 191: I like this a lot, this is the same thing I do in kgo/metadata.go: https://github.com/twmb/franz-go/blob/1a59c2d62d0d0ebd115f4f507379189644972c1f/pkg/kgo/metadata.go#L183

twmb (approved):
lgtm, great idea w/ the admin channel

PR #412, opened 2023-03-25
kgo: add NodeName for easy user formatting of internal seed node IDs

twmb:


twmb:
Reopining in new minor features branch

PR #436, opened 2023-04-17
Add offset to otel trace

owenhaynes:


plugin/kotel/tracer.go
```diff
@@ -124,6 +124,7 @@ func (t *Tracer) WithProcessSpan(r *kgo.Record) (context.Context, trace.Span) {
 		semconv.MessagingDestinationKey.String(r.Topic),
 		semconv.MessagingOperationProcess,
 		semconv.MessagingKafkaPartitionKey.Int64(int64(r.Partition)),
+		semconv.MessagingKafkaMessageOffset.Int64(r.Offset),
```
twmb, line 127: When is WithProcessSpan meant to be used? If this is while consuming, this is valid. If this is while producing, the Offset is not yet set -- it is only set when being unbuffered -- and this attribute will be invalid.

owenhaynes: With "WithProcessSpan" needs to be using in the consumption of records manually invoked by the user only for none batch processing, so during a fetch loop

owenhaynes: example:

```go
fetches.EachRecord(func(r *kgo.Record) {
   ctx, span :=  tracer.WithProcessSpan(r)
  ...
}
```

twmb: Can it be used while producing?
```
r := kgo.StringRecord("foo")
ctx, span := tracer.WithProcessSpan(r)
cl.Produce(context.Background(), r, ...)
doSomethingWithSpan()
```
?

owenhaynes: It should not be from what I understand, as the readme does not show this. I can update the function comment to say to only use during consumption?

twmb:
Makes sense, thank you. I'll likely tag this weekend once I figure out what to do with #438 (which may have to wait until kgo v1.14, which has nothing at the moment)

twmb:
funcation indeed

PR #438, opened 2023-04-21
POC - Batch Trace receive

owenhaynes:
POC not working code to show how batch tracing could be implemented.

following guidance from
https://opentelemetry.io/blog/2022/instrument-kafka-clients/
https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/messaging/#batch-receiving

The python libraries sort of already implement the above
https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-confluent-kafka/src/opentelemetry/instrumentation/confluent_kafka/__init__.py

### DIFF
--- a/pkg/kgo/hooks.go
+++ b/pkg/kgo/hooks.go
@@ -374,6 +374,10 @@ type HookFetchRecordBuffered interface {
 	OnFetchRecordBuffered(*Record)
 }
 
+type HookFetchRecordsBuffered interface {
+	OnFetchRecordsBuffered(topic string, partition int32, records []*Record)
+}
+
 // HookFetchRecordUnbuffered is called when a fetched record is unbuffered.
 //
 // A record can be internally discarded after being in some scenarios without
--- a/plugin/kotel/tracer.go
+++ b/plugin/kotel/tracer.go
@@ -9,14 +9,14 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/propagation"
-	semconv "go.opentelemetry.io/otel/semconv/v1.12.0"
+	semconv "go.opentelemetry.io/otel/semconv/v1.18.0"
 	"go.opentelemetry.io/otel/trace"
 )
 
 var ( // interface checks to ensure we implement the hooks properly.
 	_ kgo.HookProduceRecordBuffered   = new(Tracer)
 	_ kgo.HookProduceRecordUnbuffered = new(Tracer)
-	_ kgo.HookFetchRecordBuffered     = new(Tracer)
+	_ kgo.HookFetchRecordsBuffered    = new(Tracer)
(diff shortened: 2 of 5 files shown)


pkg/kgo/hooks.go
```diff
@@ -374,6 +374,10 @@ type HookFetchRecordBuffered interface {
 	OnFetchRecordBuffered(*Record)
 }
 
+type HookFetchRecordsBuffered interface {
+	OnFetchRecordsBuffered(topic string, partition int32, records []*Record)
```
owenhaynes, line 378: @twmb Only added this for a talking point. Not happy with this.

We need a way to  inject a root context into all of the messages to share the same root context. We then have a issue getting this context so that we can end the parent span. Currently  OnFetchRecordUnbuffered ends the span but not usable for the batch receive implementation.  


We need someway of storing the context which has been created for the parent span.

plugin/kotel/tracer.go
```diff
@@ -137,9 +137,14 @@ func (t *Tracer) WithProcessSpan(r *kgo.Record) (context.Context, trace.Span) {
 		trace.WithSpanKind(trace.SpanKindConsumer),
 	}
 
+	ctx := t.propagators.Extract(r.Context, NewRecordCarrier(r))
```
owenhaynes, line 140: We now extract the producer span from the record during processing.

plugin/kotel/tracer.go
```diff
@@ -220,15 +216,11 @@ func (t *Tracer) OnFetchRecordBuffered(r *kgo.Record) {
...
-	}
-	// Extract the span context from the record.
-	ctx := t.propagators.Extract(r.Context, NewRecordCarrier(r))
-	// Start the "receive" span.
-	newCtx, _ := t.tracer.Start(ctx, r.Topic+" receive", opts...)
+	newCtx, _ := t.tracer.Start(context.Background(), topic+" receive", opts...)
 	// Update the record context.
-	r.Context = newCtx
+	for _, r := range records {
+		r.Context = newCtx
```
owenhaynes, line 222: we give each record the new batch parent span

owenhaynes:
another possibility is to move the receiving span out of the hooks and have a facade  around the Fetches type which wraps in the tracing. Saving the need to make changes to kgo it self.

pkg/kgo/hooks.go
twmb: I was taking a look at this PR again but I don't have an idea for how to solve the unbuffering batch cancel aspect. Do you have an idea for it or is this PR infeasible?

owenhaynes: Yeah I still don't know how to resolve this. May just be better for now to not have the batch trace and just have the processing trace, to avoid the confusing trace output that happens atm. Will open a fresh PR this week which address that and will close this at time so we can think more about it

twmb:
Closing, based on prior latest comment (though this idea is interesting)

PR #441, opened 2023-04-27
kprom v1.x update

brunsgaard:
I thought that there was room for improvement in the kprom plugin. So I created this draft PR for further discussion.

I tried to consolidate the best parts from `examples/hooks_and_logging/expansive_prometheus` and `plugins/kprom`.

@twmb will you please take a look when you have time.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.18
 require (
 	github.com/klauspost/compress v1.16.7
 	github.com/pierrec/lz4/v4 v4.1.18
-	github.com/twmb/franz-go/pkg/kmsg v1.6.0
+	github.com/twmb/franz-go/pkg/kmsg v1.6.1
 	golang.org/x/crypto v0.11.0
 )
 
--- a/go.sum
+++ b/go.sum
@@ -2,7 +2,7 @@ github.com/klauspost/compress v1.16.7 h1:2mk3MPGNzKyxErAw8YaohYh69+pa4sIQSC0fPGC
 github.com/klauspost/compress v1.16.7/go.mod h1:ntbaceVETuRiXiv4DpjP66DpAtAGkEQskQzEyD//IeE=
 github.com/pierrec/lz4/v4 v4.1.18 h1:xaKrnTkyoqfh1YItXl56+6KJNVYWlEEPuAQW9xsplYQ=
 github.com/pierrec/lz4/v4 v4.1.18/go.mod h1:gZWDp/Ze/IJXGXf23ltt2EXimqmTUXEy0GFuRQyBid4=
-github.com/twmb/franz-go/pkg/kmsg v1.6.0 h1:9vBl1OzVf1YQWcTwBNrm5FK6Oa46d67Ea+xHVDIXhlQ=
-github.com/twmb/franz-go/pkg/kmsg v1.6.0/go.mod h1:jSnRA0Cn1/8Kj25pAHmuQSSWPssD4Rtlv/qgtiA1Mpg=
+github.com/twmb/franz-go/pkg/kmsg v1.6.1 h1:tm6hXPv5antMHLasTfKv9R+X03AjHSkSkXhQo2c5ALM=
+github.com/twmb/franz-go/pkg/kmsg v1.6.1/go.mod h1:se9Mjdt0Nwzc9lnjJ0HyDtLyBnaBDAd7pCje47OhSyw=
 golang.org/x/crypto v0.11.0 h1:6Ewdq3tDic1mg5xRO4milcWCfMVQhI4NkqWWvqejpuA=
 golang.org/x/crypto v0.11.0/go.mod h1:xgJhtzW8F9jGdVFWZESrid1U1bjeNy4zgy5cRr/CIio=


plugin/kprom/kprom.go
```diff
@@ -1,260 +1,299 @@
...
+		Help:        "Total number of records fetched, by nodeid",
+	}, []string{"nodeid", "topic"})
+
+	// e2e timing
+
+	m.e2eTimings = factory.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace:   namespace,
+		Subsystem:   subsystem,
+		ConstLabels: constLabels,
+		Name:        "e2e_broker_latency_seconds",
```
weeco, line 269: I'd rename that to node_latency rather than broker_latency so that it's consistent with the rest of the client naming. Also we use the label "nodeid".

plugin/kprom/kprom.go
```diff
@@ -1,260 +1,299 @@
...
+
+	// connects and disconnects
+
+	m.connects = factory.NewCounterVec(prometheus.CounterOpts{
+		Namespace:   namespace,
+		Subsystem:   subsystem,
+		ConstLabels: constLabels,
+		Name:        "connects_total",
+		Help:        "Total number of connections opened, by nodeid",
+	}, []string{"nodeid"})
```
weeco, line 123: Do you prefer `nodeid` over `node_id`, is there a convention for naming the labels without using underscores? I think I'd prefer `node_id` and `client_id` for better readability like it was before.

plugin/kprom/kprom.go
```diff
@@ -1,260 +1,299 @@
...
+	}, []string{"nodeid", "topic"})
+
+	// e2e timing
+
+	m.e2eTimings = factory.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace:   namespace,
+		Subsystem:   subsystem,
+		ConstLabels: constLabels,
+		Name:        "e2e_broker_latency_seconds",
+		Help:        "Round trip time to Kafka, by nodeid",
```
weeco, line 270: I think this metric requires a bit more description. What exactly is being measured there? I'm unsure what E2E refers to (does it cover all kafka api requests, where is measuring starting/ending). Haven't been able to figure it out just by looking at the PR diff here.

weeco (commented):
A couple good additions! I left a few minor comments.

brunsgaard:
@weeco Thanks for taking a look, I incorporated you feedback in 70b479e8b58416cc035e06665b8df4070d9cfc8a and rebased on upstream, Are there anything thing else, or should I promote the daft as ready for PR?

plugin/kprom/kprom.go
brunsgaard: I updated the help string 

> Total number of seconds for the a write of a request followed by a read of that requests's response, by node_id, this is the sum of TimeToWrite, ReadWait and TimeToRead defined on the BrokerE2E struct

BrokerE2E struct is defined here https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#BrokerE2E

plugin/kprom/kprom.go
brunsgaard: agreed, and updated

plugin/kprom/kprom.go
brunsgaard: 💯

plugin/kprom/config.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// WithSybsystem allows the user to provide a sybsystem name for metrics,
+// defaults to "kafka"
+func WithSubsystem(ss string) Opt {
+	return opt{func(c *cfg) { c.subsystem = ss }}
+}
+
+// WithBuckets allows the user to specify the buckets used for kprom
+// histrograms, defaults to
```
twmb, line 58: histrograms => histograms

Also what do these buckets mean? It might be good to indicate 1ms, 2ms, 4ms, etc.
This is more detailed than prometheus.DefBuckets -- are the default buckets not really ... sufficient?

twmb:
I think the main breaking change is removing namespace as a required argument to NewMetrics. We can keep this in the v1.x series if you add back in a required namespace. A few options in this have `With` prefixed, but exist in the old version without `With`. I personally go back and forth on whether options should have With, because once you have a bunch of options, the `With` becomes a lot of noise (kgo actually started out with `With` but I dropped it quickly).

I'd only add each set of timings if requested, and not opt into any set of timings by default. They quickly become a huge set of metrics / cardinality that some orgs don't want to deal with.

plugin/kprom/config.go
brunsgaard: The default buckets might be a good default, but I think for Kafka the latencies we expect and want to measure are often in a range that is lower then what the defaults are designed for. I pulled a couple of metrics from our services to illustrate. And this is e2e, the detailed e2e metrics will be lower. I know everybody is having different setups, but I think it makes sense to have buckets that provide more insight into the lower ranges. But I also see the argument to use the defaults and let users overwrite them. Maybe users should even be able to overwrite them per histogram.

```
P90
histogram_quantile(0.9, sum by (client_id, le) (rate(ns_kafka_e2e_node_latency_seconds_bucket[3m])))
{clientid="ex"}     5.94 ms
{clientid="ar"}     96.5 ms
{clientid="us"}     3.06 ms
{clientid="br"}     90.9 ms
{clientid="de"}     2.45 ms

P50
histogram_quantile(0.5, sum by (client_id, le) (rate(ns_kafka_e2e_node_latency_seconds_bucket[3m])))
{clientid="ex"}     2.33 ms
{clientid="ar"}     30.7 ms
{clientid="us"}     0.923 ms
{clientid="br"}     30.1 ms
{clientid="de"}     1.54 ms
```

brunsgaard:
I dont think we should keep it in 1.x, the metrics and labels change name and dimensions. Also I dont think a ns should be required and I would like it to work as kgo.NewClient, where all config follows the opt pattern.

brunsgaard:
@twmb I made some updates to the plugin, making the Histograms optional, worked with Opts and I added documentation and examples. 

I think there are too many breaking changes to metric names, labels, opts, interfaces etc and I dont see how I will be able to keep V1 compatibility and still reach my original goal of reworking the kprom plugin consolidating the best parts from both the kprom V1, the expansive_prometheus example and rethinking the contract. 

Let me know where you want to go from here, I am open to take a couple of more iterations on this, but if you already now know that you dont want to introduce a V2 with breaking changes I should properly close the PR here.

twmb:
This was complicated to analyze so I needed to find some time to sit down and actually diff everything in detail.

# Breakage

## APIs

ADDED
* var DefBuckets
* type FetchProduceLabel (Topic | NodeID) // comment: uppercase NodeId -- allows topic and node_id labels to be optional (required in old)
* type Histogram (ReadWait | ReadTime | WriteWait | WriteTime | RequestDurationE2E | RequestThrottled
* type HistogramOpts
* func (*Metrics) OnBrokerE2E
* func (*Metrics) OnBrokerThrottle
* func Buckets([]float64) Opt
* func FetchProduceLabels(...FetchProduceLabel) Opt
* func Histograms(...Histogram) Opt
* func HistogramsFromOpts(...HistogramOpts) Opt
* func Namespace(string) Opt
* func Subsystem(string) Opt

CHANGED
* func NewMetrics(namespace string, opts ...Opt) => NewMetrics(opts ...Opt)

REMOVED
* func (*Metrics) Handler (add doc: providing prometheus metrics for the gatherer)
* func (*Metrics) OnBrokerRead
* func (*Metrics) OnBrokerWrite
* func Registry(RegistererGatherer) Opt
* func Gatherer(prometheus.Gatherer) Opt
* func GoCollectors() Opt
* func HandlerOpts(promhttp.HandlerOpts) Opt

## Metrics

ADDED
* (label) client_id ConstLabel on everything
* write_wait_seconds, histogram
* write_time_seconds, histogram
* read_wait_seconds, histogram
* read_time_seconds, histogram
* request_duration_e2e_seconds, histogram
* request_throttled_seconds, histogram
* produce_compressed_bytes
* produce_batches_total
* produce_records_total
* fetch_compressed_bytes
* fetch_batches_total
* fetch_records_total

RENAMED
* connects_total => conn_connects_total
* connect_errors_total => conn_connect_errors_total
* disconnects_total => conn_disconnects_total
* produce_bytes_total => produce_uncompressed_bytes
* fetch_bytes_total => fetch_uncompressed_bytes

## Internal defaults changed

* node_id dropped from fetch & produce bytes
* kafka subsystem added

# Proposals

It's possible to do a lot of this without breaking APIs.

Keep all old APIs
* OnBrokerRead, OnBrokerWrite change to noop in favor of OnBrokerE2E replacement
* No harm to keep Registry, Gatherer, GoCollectors, HandlerOpts, (*Metrics) Handler

Drop Namespace option, drop internal "kafka" subsystem default
* _purely_ option driven might be a decent goal, but if we can keep a 1.x release that's preferable, and namespaces are basically _required_ in Prometheus so it made sense to me to make it a required argument
* Default subsystem "kafka" makes less sense if you produce to Redpanda / MSK / Event Hubs -- I'd rather force a person to specify the subsystem here ;)

Add WithClientLabel() Opt
* Replaces new breaking ConstLabel default

Keep old names for 5 renamed metrics
* Presumably the conn_ prefix is to help for grouping but it's not critical
* Keeping (produce|fetch)_bytes_total can avoid breakage and can be explained away in docs.
  * The reason I chose uncompressed bytes only is because per-broker write bytes shows you the compressed sizes (albeit without per-topic)

Change 6 new produce&fetch metrics, as well as FetchProduceLabel and FetchProduceLabels
* new option FetchAndProduceDetail(...Detail), with internal default (UncompressedBytes, ByTopic, ByNode)
* new enum type, Detail (UncompressedBytes, CompressedBytes, Batches, Records, ByTopic, ByNode)
  * Can even add ConsistentNaming to force rename (produce|fetch)_bytes_total to (produce|fetch)_uncompressed_bytes

I'm not entirely positive about the histogram APIs yet but they don't break the current APIs at all. What do you think about the above proposals?

freeformz:
Would be nice to be able to unregister all the metrics too: https://github.com/twmb/franz-go/pull/469. That may not be necessary with for us though with a WithClientLabel API addition.

brunsgaard:
I updated the PR, I have not tested the changes yet, but now the PR should conform to the proposal. If you have time please glance over it.

brunsgaard:
This is ready for for review. 

@freeformz I also implemented `Unregister`, but I did it as part of the `kgo.HookClientClosed`.

twmb:
No idea how I borked your master and closed this PR, I moved a few things around and fixed a few docs, reopened as https://github.com/twmb/franz-go/pull/505, this lgtm. I'm going to test locally -- I'm a bit worried about installing histograms to the factory by default _but_ I _think_ they don't show up because they aren't ever incremented.

PR #442, opened 2023-04-28
Add Benthos to the README users list

Jeffail:
I felt it would look inconsistent to put emojis in the list itself so I'll add them here: ❤️🫦


twmb (approved):
Awesome tyvm

PR #449, opened 2023-05-03
kotel semconv use correct v1.18.0 conventions for tracing attributes

owenhaynes:

PR #450, opened 2023-05-05
kotel: Update span operation name from 'send' to 'publish'

yianni:
Building on PR #449, this PR follows the new messaging semantic by renaming the span operation for producing from 'send' to 'publish'.


twmb:
I'm assuming this is ok to release in a patch?

twmb:
plugin/kotel/v1.2.0
thanks!

PR #455, opened 2023-05-23
Upgrade to otel metric v1.16.0

lahsivjar:
otel/metric is now available as a stable release. The `go.opentelemetry.io/otel/metric/global` package is removed in favor of  `go.opentelemetry.io/otel`.


twmb:
Thank you, tagged as plugin/kotel/v1.3.0

PR #460, opened 2023-05-25
Schema Registry error codes

lovromazgon:
This PR adds the type `ErrorCode` and constants for known error codes taken from https://docs.confluent.io/platform/current/schema-registry/develop/api.html.

### DIFF
--- a/pkg/sr/api.go
+++ b/pkg/sr/api.go
@@ -14,6 +14,31 @@ import (
 //     https://docs.confluent.io/platform/current/schema-registry/develop/api.html
 //
 
+const (
+	ErrorCodeSubjectNotFound  ErrorCode = 40401
+	ErrorCodeVersionNotFound  ErrorCode = 40402
+	ErrorCodeSchemaNotFound   ErrorCode = 40403
+	ErrorCodeExporterNotFound ErrorCode = 40450
+
+	ErrorCodeMissingOrInvalidExporterName   ErrorCode = 40950
+	ErrorCodeMissingOrInvalidExporterConfig ErrorCode = 40951
+	ErrorCodeInvalidExporterSubjects        ErrorCode = 40952
+	ErrorCodeExporterAlreadyExists          ErrorCode = 40960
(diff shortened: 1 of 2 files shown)


twmb:
I might take this and change it into
```
type Error struct {
    Code int
    Message string
}

func (e *Error) Error() string { return fmt.Sprintf("%v: %s", e.Code, e.Message) }
```
and then have a bunch of
```
ErrXyz = &Error{...}
```
for all of these errors, similar to the kerr package. What do you think?

lovromazgon:
Sure, error comparisons would definitely be nicer then. The only drawback I see is that messages would be hardcoded in advance and not updated if the server starts responding with a different message. Although that can be seen as a feature 🙃 

I can adjust the code to use that structure, it should stay in the same package though, right?

lovromazgon:
I changed the code to use an `Error` type, it's not done yet, the error messages need to be filled in. I left `ResponseError` as a wrapper around `Error` so that additional info is printed when formatting it with `"%+v"`. It also has the `Unwrap` method so that the actual error can be compared using `errors.Is`.

The only issue I see with this approach is that we lose the dynamic message returned by the server. For example, the server actually responds with the message `Schema 123 not found`, but since we convert it into a hardcoded error object it would probably read something like `Schema not found` instead.

Let me know if this was what you had in mind.

twmb:
I think you have a point here. I'll try to take a look more tomorrow, busy today and probably tomorrow to actually.

lovromazgon:
Whoops, I renamed the branch in my fork and it closed the PR, sorry about that. But now that I think about it, I will leave it closed as this is more of a "nice to have", I'd rather focus on the other two schema registry related PRs for the time being.

twmb:
For posterity, just in case:
```
40401 -- subject not found
40402 -- version not found
40403 -- schema not found

409 -- incompatible schema

42201 -- invalid schema or schema type
42202 -- invalid version
42203 -- invalid compatibility
42204 -- invalid mode
42205 -- operation not permitted

50001 -- failure in backend data store
50003 -- failure forwarding the request to the primary
50004 -- unknown leader
```

PR #461, opened 2023-05-25
Add method for retrieving schema by ID

lovromazgon:
This adds the method `SchemaByID` which returns the whole `Schema` object for a given schema ID. In contrast to `SchemaTextByID` this one also returns the schema type to the caller.

It also changes how `SchemaTextByID` retrieves the schema text, by using the dedicated endpoint.

### DIFF
--- a/pkg/sr/api.go
+++ b/pkg/sr/api.go
@@ -140,6 +140,13 @@ func (cl *Client) Subjects(ctx context.Context, deleted HideShowDeleted) ([]stri
 	return subjects, cl.get(ctx, path, &subjects)
 }
 
+// SchemaByID returns the schema for a given schema ID.
+func (cl *Client) SchemaByID(ctx context.Context, id int) (Schema, error) {
+	// GET /schemas/ids/{id}
+	var s Schema
+	return s, cl.get(ctx, fmt.Sprintf("/schemas/ids/%d", id), &s)
+}
+
 // SchemaTextByID returns the actual text of a schema.
 //
 // For example, if the schema for an ID is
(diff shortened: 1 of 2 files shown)


pkg/sr/api.go
```diff
@@ -150,12 +157,12 @@ func (cl *Client) Subjects(ctx context.Context, deleted HideShowDeleted) ([]stri
 //
 //	{"type":"boolean"}
 func (cl *Client) SchemaTextByID(ctx context.Context, id int) (string, error) {
-	// GET /schemas/ids/{id}
-	var s Schema
-	if err := cl.get(ctx, fmt.Sprintf("/schemas/ids/%d", id), &s); err != nil {
+	// GET /schemas/ids/{id}/schema
+	var s []byte
```
twmb, line 161: Does this work? When the unmarshal destination is a []byte, the json package auto-base64-decodes. Doesn't this need to be `string`?

lovromazgon: Yeah I added it as a utility to get the raw body if needed, like here for example. I mean the alternative would be to leave the method as it is and go through the other endpoint that returns a JSON.

lovromazgon: @twmb any chance we could get this merged? I see a failing CI action, but it seems that it's not related to the code changes and was caused by the redpanda container not stopping correctly.

twmb: Ah I didn't realize I left this unmergd.

twmb:
Also I'd be curious if your cache could be somewhat replaced by github.com/twmb/go-cache/cache :D

lovromazgon:
That's a shameless plug, if I've ever seen one 😄 The "mostly lock-free" part intrigues me, my current implementation is very straight forward and naive, it locks _a lot_. I might actually give it a go, thanks! 👍

twmb:
More importantly is that your current implementation can cause a bunch of inflight requests when a schema ID isn't known -- i.e. there's no request collapsing. I'm not sure if "bunch of inflight" is a possibility (didn't skim the code tooooo much) but request collapsing is nice.

lovromazgon:
Good catch, thanks for pointing that out!

PR #463, opened 2023-05-31
kotel: add fetch_records.count and produce_records.count metrics

sbuliarca:
We are considering switching to this awesome library in our company from sarama.
We are currently visualizing the throughput of consumers and producers by the number of records they consume and produce.
While the number of bytes gives a good indication of the throughput, I think that adding the number of messages in the default metrics would be a good addition, as together with the number of bytes would paint a more complete picture on the throughput.


twmb (approved):
lgtm thanks!

twmb:
released in plugin/kotel/v1.4.0

Issue #453, opened 2023-05-17
Schema registry module tagged release


lovromazgon:
I'm wondering if there's a reason that there's no tagged release for the schema registry code in `pkg/sr`? Having proper tags would make it easier to update the dependency with dependabot.

twmb:
The main reason is that I want to either use this myself extensively _or_ have some community feedback on how the package feels before stabilizing it, to be sure that the API isn't kludgy / that it shouldn't change. I partially would like to have plugins for Avro / Schema Registry available as well, so the entire thing is just plug & play for producing & consuming. The plugins would take a lot of time / I don't have time (this is more for the schema registry actually). But, if you'd like to try things out and let me know how it is -- I'd try to set aside some time to move it forward. I don't think there's anything more that I want to do with _this_ package, but look&feel feedback would be good. I'm planning to revisit [this](https://github.com/redpanda-data/redpanda/pull/4861) PR someday which uses the `sr` package and that might be a forcing function for me to play with the look&feel more. I don't anticipate changes at this point, though, since writing that PR originally is what allowed me to play with the `sr` package in the first place. The main thing to add to that PR is the producing & consuming components.

lovromazgon:
Well, let me give you some context first. My plan is to use the library in [Conduit](https://github.com/ConduitIO/conduit) to provide encoder and decoder processors which use schemas from the schema registry to encode/decode a payload (similar to Kafka Connect [key/value converters](https://docs.confluent.io/platform/current/schema-registry/connect.html#example-converter-properties)). I've started playing around with the schema registry client and the serde interface and the experience is quite good, especially compared to existing alternatives out there. I can give you feedback on what I found so far.

I found some minor issues like [some comments](https://github.com/twmb/franz-go/blob/4ec11aead75b9a6918cea91cbba9b673ac6d213d/pkg/sr/api.go#L292-L294) not matching [function names](https://github.com/twmb/franz-go/blob/4ec11aead75b9a6918cea91cbba9b673ac6d213d/pkg/sr/api.go#L305-L308) and the [mode not being set](https://github.com/twmb/franz-go/blob/4ec11aead75b9a6918cea91cbba9b673ac6d213d/pkg/sr/api.go#L616) (I plan to open a PR with fixes).

[...]

PR #466, opened 2023-06-01
Serde Encode Options

lovromazgon:
This introduces encode options in `Serde.Encode` and related functions. There are currently two encode options, `ID` and `Index`. I reused the existing `Index` option and adjusted it to function both as ` SerdeOpt` and `EncodeOpt`.

The caller can now explicitly specify an ID and index when encoding an object, which overrides the default behavior of choosing the encoding function based on the type of the value being encoded. The primary motivation behind this is to support the encoding of dynamic data stored in a map by explicitly choosing the schema for encoding it. Besides that, it could also be used to register multiple schemas/encode functions for the same type and explicitly choose which one to use based on the ID.

It also fixes some minor bugs:
- When registering a schema without supplying `DecodeFn`, `Decode` now returns `ErrNotRegistered`
- When using `DecodeNew` on a schema without an associated type it now returns `ErrNotRegistered` instead of a panic

Related to the discussion in https://github.com/twmb/franz-go/issues/453.

### DIFF
--- a/pkg/sr/serde.go
+++ b/pkg/sr/serde.go
@@ -27,11 +27,32 @@ var (
 
 type (
 	// SerdeOpt is an option to configure a Serde.
-	SerdeOpt interface{ apply(*tserde) }
+	SerdeOpt interface{ applySerde(*tserde) }
 	serdeOpt struct{ fn func(*tserde) }
+
+	// EncodeOpt is an option to configure Serde.Encode and related functions.
+	EncodeOpt  interface{ applyEncode(*encodeOpts) }
+	encodeOpt  struct{ fn func(*encodeOpts) }
+	encodeOpts struct {
+		id    int
+		index []int
+	}
+
+	// SerdeEncodeOpt is an option to configure Serde and Serde.Encode.
+	SerdeEncodeOpt interface {
+		SerdeOpt
+		EncodeOpt
+	}
+	serdeEncodeOpt struct {
+		serdeFn  func(*tserde)
+		encodeFn func(*encodeOpts)
+	}
 )
 
-func (o serdeOpt) apply(t *tserde) { o.fn(t) }
+func (o serdeOpt) applySerde(t *tserde)               { o.fn(t) }
+func (o encodeOpt) applyEncode(opts *encodeOpts)      { o.fn(opts) }
+func (o serdeEncodeOpt) applySerde(t *tserde)         { o.serdeFn(t) }
+func (o serdeEncodeOpt) applyEncode(opts *encodeOpts) { o.encodeFn(opts) }
 
 // EncodeFn allows Serde to encode a value.
 func EncodeFn(fn func(any) ([]byte, error)) SerdeOpt {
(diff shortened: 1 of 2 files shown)


twmb:
Mind changing this PR to what we've agreed on in https://github.com/twmb/franz-go/pull/467#discussion_r1231620122 ?

lovromazgon:
I created a separate PR that includes those changes (#TODO). It is mutually exclusive with this PR, both kinda do the same thing, but with different approaches. The new PR is based on your [comment](https://github.com/twmb/franz-go/pull/467#discussion_r1232771653) that we don't necessarily need to separate `RegisterOpt` and `EncodeOpt`. I'll try to articulate what benefits and drawbacks I see for each approach and leave it for you to decide which one you prefer.

Having separate `RegisterOpt` and `EncodeOpt` (i.e. this PR) allows us to prevent the user from sending `ID` into `Register`, as that function already expects the ID as a normal argument. In the new PR I had to decide what to do if that happens (see [these lines](https://github.com/twmb/franz-go/pull/506/files#diff-1ce708d325f43d3a342ca86f68fd11ada2f8747bc9c75f58b94a706c086c0c0dR207-R210)). Options `ID` and `Index` also need special treatment in function `Encode`, as they are not applied on `tserde`, but rather they decide which `tserde` to use. In this PR this is solved by having a [struct](https://github.com/twmb/franz-go/pull/466/files#diff-1ce708d325f43d3a342ca86f68fd11ada2f8747bc9c75f58b94a706c086c0c0dR36-R39) that is mutated using `EncodeOpt`. In the new PR options `ID` and `Index` are separated from all options and treated separately (see [this function](https://github.com/twmb/franz-go/pull/506/files#diff-1ce708d325f43d3a342ca86f68fd11ada2f8747bc9c75f58b94a706c086c0c0dR477)). The internals of the new approach seem a bit hacky, although for the caller it might be simpler and more flexible to have only one set of options that can be passed both to `Register` and `Encode` (possibly also to `Decode`, see [this comment](https://github.com/twmb/franz-go/pull/506/commits/1ade586c0fcdcde72a6c81cd6dbe250d432e9dbb#r1257427075)).

Let me know what you think and feel free to close one of these PRs, I'll then focus on the one that's left.

lovromazgon:
Hey @twmb, hope you're well! I was wondering if you had any updates on the PRs? I haven't received any response and I just wanted to make sure everything is okay. Let me know if there's anything I can do to help. Thanks!

twmb:
I'm going to focus on 506 first (which I think includes some of this) and have some thoughts about the EncodingOpt additions to the Encode functions -- I've come to a thought a few days ago that I think we should essentially double the encode functions. I think variadic options for every individual encoding function is a bit odd and _way_ more extensive than any encoder I've seen. I wonder if we can either have a new extendable struct their (a la slog's NewEncoder functions) or if we can have two APIs (Encode, EncodeWithOpt).

twmb:
Finally left my comment on 506 -- also can 506 be rebased on the latest / resolve the TODOs about the header PR?

twmb:
Closing in favor of continuing this in #506

PR #467, opened 2023-06-02
Serde Header

lovromazgon:
This PR introduces `SerdeHeader` and defines `ConfluentHeader` as the default implementation. The behavior of `Serde` is the same, ~one small but important difference is a fixed bug in the encoding of the message index (see the second part of the PR description)~. The changes are mostly just a refactoring and extraction of the header encoding/decoding logic. It makes `Serde` a bit more flexible and future-proof, since the header format can be switched to a different implementation.

The motivation behind this change is to expose the logic for encoding/decoding the Confluent wire format header. Using this change the caller can now decode the ID manually if needed and fetch the schema from the schema registry.

```go
val, err := serde.DecodeNew(b)
if errors.Is(err, sr.ErrNotRegistered) {
	id, _ := serde.Header().DecodeID(b)

	// fetch schema from schema registry using the ID...

	serde.Register(id, myVal, sr.DecodeFn(func(b []byte, v any) error {
		// decode using the schema ...
	})

	// retry decoding
	val, err = serde.DecodeNew(b)
}
```

The code above is just an illustration, I'm aware that the code would need to take care of request collapsing.

Related to the discussion in https://github.com/twmb/franz-go/issues/453.

### DIFF
--- a/pkg/sr/serde.go
+++ b/pkg/sr/serde.go
@@ -191,8 +191,11 @@ func (s *Serde) Register(id int, v any, opts ...SerdeOpt) {
 		}
 		return j
 	}
+	myDepth := len(t.index)
 	for i, idx := range t.index {
-		at.subindexDepth = max(at.subindexDepth, len(t.index)-i)
+		// SAFETY: tserdeMapClone deeply clones all maps through the index, so
+		// our modified value is being saved to a new map that is not being read.
+		at.subindexDepth = max(at.subindexDepth, myDepth-i)
 		m[k] = at
 
 		m = at.subindex
--- a/pkg/sr/serde.go
+++ b/pkg/sr/serde.go
@@ -132,7 +132,7 @@ func (s *Serde) SetDefaults(opts ...SerdeOpt) {
 }
 
 // SetHeader configures which header should be used when encoding and decoding
-// values. If the header is set to nil it falls back to ConfluentHeader.
+// values. If the header is set to nil it falls back to confluentHeader.
 func (s *Serde) SetHeader(header SerdeHeader) {
 	s.header = header
 }
@@ -140,7 +140,7 @@ func (s *Serde) SetHeader(header SerdeHeader) {
 // Header returns the configured header.
 func (s *Serde) Header() SerdeHeader {
 	if s.header == nil {
-		return ConfluentHeader{}
+		return defaultSerdeHeader
 	}
 	return s.header
 }
@@ -363,16 +363,18 @@ type SerdeHeader interface {
 	DecodeIndex(in []byte, maxLength int) (index []int, out []byte, err error)
 }
 
-// ConfluentHeader is a SerdeHeader that produces the Confluent wire format. It
+var defaultSerdeHeader = new(confluentHeader)
+
+// confluentHeader is a SerdeHeader that produces the Confluent wire format. It
 // starts with 0, then big endian uint32 of the ID, then index (only protobuf),
 // then the encoded message.
 //
 // https://docs.confluent.io/platform/current/schema-registry/serdes-develop/index.html#wire-format
-type ConfluentHeader struct{}
+type confluentHeader struct{}
 
 // AppendEncode appends an encoded header to b according to the Confluent wire
 // format and returns it. Error is always nil.
-func (ConfluentHeader) AppendEncode(b []byte, id int, index []int) ([]byte, error) {
+func (*confluentHeader) AppendEncode(b []byte, id int, index []int) ([]byte, error) {
 	b = append(
 		b,
 		0,
@@ -399,7 +401,7 @@ func (ConfluentHeader) AppendEncode(b []byte, id int, index []int) ([]byte, erro
 // DecodeID strips and decodes the schema ID from b. It returns the ID alongside
 // the unread bytes. If the header does not contain the magic byte or b contains
 // less than 5 bytes it returns ErrBadHeader.
-func (ConfluentHeader) DecodeID(b []byte) (int, []byte, error) {
+func (*confluentHeader) DecodeID(b []byte) (int, []byte, error) {
 	if len(b) < 5 || b[0] != 0 {
 		return 0, nil, ErrBadHeader
 	}
@@ -412,7 +414,7 @@ func (ConfluentHeader) DecodeID(b []byte) (int, []byte, error) {
 // ID should already be stripped away). If maxLength is greater than 0 and the
 // encoded data contains more indices than maxLength the function returns
 // ErrNotRegistered.
-func (ConfluentHeader) DecodeIndex(b []byte, maxLength int) ([]int, []byte, error) {
+func (*confluentHeader) DecodeIndex(b []byte, maxLength int) ([]int, []byte, error) {
 	buf := bytes.NewBuffer(b)
 	l, err := binary.ReadVarint(buf)
 	if err != nil {
--- a/pkg/sr/serde.go
+++ b/pkg/sr/serde.go
@@ -421,6 +421,9 @@ func (ConfluentHeader) DecodeIndex(b []byte, maxLength int) ([]int, []byte, erro
 	if l == 0 { // length 0 is a shortcut for length 1, index 0
 		return []int{0}, buf.Bytes(), nil
 	}
+	if l < 0 { // index length can't be negative
+		return nil, nil, ErrBadHeader
+	}
 	if maxLength > 0 && int(l) > maxLength { // index count is greater than expected
 		return nil, nil, ErrNotRegistered
 	}
--- a/pkg/sr/serde.go
+++ b/pkg/sr/serde.go
@@ -163,26 +163,28 @@ func (s *Serde) Register(id int, v any, opts ...SerdeOpt) {
 	defer s.mu.Unlock()
 
 	// Type mapping is easy: we just map the type to the final tserde.
-	// We defer the store because we modify the tserde below, and we
-	// may delete a type key.
 	dupTypes := make(map[reflect.Type]tserde)
 	for k, v := range s.loadTypes() {
 		dupTypes[k] = v
 	}
+
+	// For IDs, we deeply clone any path that is changing.
+	dupIDs := tserdeMapClone(s.loadIDs(), id, t.index)
+
+	// We defer the store because we modify the tserde below, and we
+	// may delete a type key.
 	defer func() {
 		dupTypes[typeof] = t
 		s.types.Store(dupTypes)
+		s.ids.Store(dupIDs)
 	}()
-
-	// For IDs, we deeply clone any path that is changing.
-	m := tserdeMapClone(s.loadIDs(), id, t.index)
-	s.ids.Store(m)
 
 	// Now we have a full path down index initialized (or, the top
 	// level map if there is no index). We iterate down the index
 	// tree to find the end node we are initializing.
 	k := id
-	at := m[k]
+	m := dupIDs
+	at := dupIDs[k]
 	max := func(i, j int) int {
 		if i > j {
 			return i
--- a/pkg/sr/serde.go
+++ b/pkg/sr/serde.go
@@ -17,7 +17,7 @@ var (
 
 	// ErrBadHeader is returned from Decode when the input slice is shorter
 	// than five bytes, or if the first byte is not the magic 0 byte.
-	ErrBadHeader = errors.New("5 byte header for value is missing or does no have 0 magic byte")
+	ErrBadHeader = errors.New("5 byte header for value is missing or does not have 0 magic byte")
 )
 
 type (
@@ -97,7 +97,7 @@ type Serde struct {
 	mu    sync.Mutex
 
 	defaults []SerdeOpt
-	header   SerdeHeader
+	h        SerdeHeader
 }
 
 var (
@@ -131,12 +131,24 @@ func (s *Serde) SetDefaults(opts ...SerdeOpt) {
 	s.defaults = opts
 }
 
-// Header returns the configured header.
-func (s *Serde) Header() SerdeHeader {
-	if s.header == nil {
+// DecodeID decodes an ID from in, returning the ID and the remaining bytes,
+// or an error.
+func (s *Serde) DecodeID(b []byte) (id int, out []byte, err error) {
+	return s.header().DecodeID(b)
+}
+
+// DecodeIndex decodes at most maxLength of a schema index from in, returning
+// the index and remaining bytes, or an error. It expects b to be the output of
+// DecodeID (schema ID should already be stripped away).
+func (s *Serde) DecodeIndex(in []byte, maxLength int) (index []int, out []byte, err error) {
+	return s.header().DecodeIndex(in, maxLength)
+}
+
+func (s *Serde) header() SerdeHeader {
+	if s.h == nil {
 		return defaultSerdeHeader
 	}
-	return s.header
+	return s.h
 }
 
 // Register registers a schema ID and the value it corresponds to, as well as
@@ -178,7 +190,7 @@ func (s *Serde) Register(id int, v any, opts ...SerdeOpt) {
 	// tree to find the end node we are initializing.
 	k := id
 	m := dupIDs
-	at := dupIDs[k]
+	at := m[k]
 	depth := len(t.index)
 	max := func(i, j int) int {
 		if i > j {
@@ -255,7 +267,7 @@ func (s *Serde) AppendEncode(b []byte, v any) ([]byte, error) {
 		return b, ErrNotRegistered
 	}
 
-	b, err := s.Header().AppendEncode(b, int(t.id), t.index)
+	b, err := s.header().AppendEncode(b, int(t.id), t.index)
 	if err != nil {
 		return nil, err
 	}
@@ -323,9 +335,7 @@ func (s *Serde) DecodeNew(b []byte) (any, error) {
 }
 
 func (s *Serde) decodeFind(b []byte) ([]byte, tserde, error) {
-	h := s.Header()
-
-	id, b, err := h.DecodeID(b)
+	id, b, err := s.DecodeID(b)
 	if err != nil {
 		return nil, tserde{}, err
 	}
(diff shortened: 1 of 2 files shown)


twmb:
The documentation is worded confusingly and I got tripped up by that the first time I read it. The "reading the array backwards" describes the following sentence itself, not the wire encoding. The following sentence first describes the inner struct -- the 0th index of the inner struct -- and then describes the outer struct -- the 1st index of the outer struct. The paragraph beneath this describes the actual encoding of the array, " So the above array [1, 0] is encoded as the variable length ints 2,1,0 where the first 2 is the length"

lovromazgon:
Oh you're right, I read it again carefully. I double checked what they do in [confluent-kafka-go](https://github.com/confluentinc/confluent-kafka-go/blob/acd3a1e4ea9c988d5d96361a5adf73d5495696b5/schemaregistry/serde/protobuf/protobuf.go#L325-L328) and seems like the order was correct as it was.

I'll revert this part 👍

pkg/sr/serde.go
```diff
@@ -316,44 +312,103 @@ func (s *Serde) DecodeNew(b []byte) (any, error) {
...
+func (ConfluentHeader) DecodeIndex(b []byte) ([]int, []byte, error) {
+	buf := bytes.NewBuffer(b)
+	l, err := binary.ReadVarint(buf)
+	if err != nil {
+		return nil, nil, err
+	}
+	if l == 0 { // length 0 is a shortcut for length 1, index 0
+		return []int{0}, buf.Bytes(), nil
+	}
+	index := make([]int, l)
```
twmb, line 405: This trust the value to be formatted correctly too much -- if somebody has a busted encoder and `l` here is unreasonably large, this will crash the program. I'm going to pull this branch and modify it a little bit (which also will make it easier to see the diff here). I think Header can be a SerdeOpt as well, passed to NewSerde, to set the default -- and change it per type (not that anybody would).

twmb: FWIW it _may_ be ok to allocate blindly, but I want to play around with the API a bit here to see if something else feels good. The old code iteratively advances into the index so it doesn't allocate everything per what's on the wire.

pkg/sr/serde.go
```diff
@@ -316,44 +312,105 @@ func (s *Serde) DecodeNew(b []byte) (any, error) {
...
-		b = r.b
 	}
 	if !t.exists {
-		return nil, t, ErrNotRegistered
+		return nil, tserde{}, ErrNotRegistered
 	}
 	return b, t, nil
 }
 
-type bReader struct{ b []byte }
```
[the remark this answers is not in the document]
twmb: I think I often misremember that bytes.Reader is relatively small, and mix it up with bytes.Buffer being a pretty big struct -- but it looks like that's not even true anymore, since the `bootstrap` field was removed in 2018: https://github.com/golang/go/commit/9c2be4c22d78cebc52458ba7298a470a3be0cdce

pkg/sr/serde.go
```diff
@@ -316,44 +312,103 @@ func (s *Serde) DecodeNew(b []byte) (any, error) {
...
+	}
+	id := binary.BigEndian.Uint32(b[1:5])
+	return int(id), b[5:], nil
+}
+
+// DecodeIndex strips and decodes the index from b. It returns the index
+// alongside the unread bytes. It expects b to be the output of DecodeID (schema
+// ID should already be stripped away).
+func (ConfluentHeader) DecodeIndex(b []byte) ([]int, []byte, error) {
+	buf := bytes.NewBuffer(b)
```
twmb, line 397: bytes.Reader

pkg/sr/serde.go
twmb: I think a way to solve it would be to have a maxLength field, and then to keep the max index length per registered ID.

pkg/sr/serde.go
lovromazgon: I wanted to use the reader initially, but it lacks the `Bytes` method that the buffer provides, for returning unread bytes.

pkg/sr/serde.go
lovromazgon: > I think a way to solve it would be to have a maxLength field, and then to keep the max index length per registered ID.

Have a look at the [last commit](https://github.com/twmb/franz-go/pull/467/commits/b107644c68a2a1dfcd993d416683533f92aaf93a), is that what you had in mind?

pkg/sr/serde.go
```diff
@@ -134,6 +131,20 @@ func (s *Serde) SetDefaults(opts ...SerdeOpt) {
 	s.defaults = opts
 }
 
+// SetHeader configures which header should be used when encoding and decoding
```
twmb, line 134: wdyt about this this being `SetGlobal`, and then `SerdeHeader` being a `GlobalOpt` -- i.e. something that applies to all serdes? y/n?

pkg/sr/serde.go
```diff
@@ -134,6 +131,20 @@ func (s *Serde) SetDefaults(opts ...SerdeOpt) {
...
+// values. If the header is set to nil it falls back to ConfluentHeader.
+func (s *Serde) SetHeader(header SerdeHeader) {
+	s.header = header
+}
+
+// Header returns the configured header.
+func (s *Serde) Header() SerdeHeader {
```
twmb, line 141: I think this can be private

pkg/sr/serde.go
```diff
@@ -316,44 +323,108 @@ func (s *Serde) DecodeNew(b []byte) (any, error) {
...
+// ConfluentHeader is a SerdeHeader that produces the Confluent wire format. It
+// starts with 0, then big endian uint32 of the ID, then index (only protobuf),
+// then the encoded message.
+//
+// https://docs.confluent.io/platform/current/schema-registry/serdes-develop/index.html#wire-format
+type ConfluentHeader struct{}
+
+// AppendEncode appends an encoded header to b according to the Confluent wire
+// format and returns it. Error is always nil.
+func (ConfluentHeader) AppendEncode(b []byte, id int, index []int) ([]byte, error) {
```
twmb, line 369: Can these methods be on a pointer? Empty structs are unpacked IIRC unless things have changed in the last few releases -- being on a pointer _should_ be faster.

pkg/sr/serde.go
twmb: wdyt about re-adding bReader in light of this^^?

pkg/sr/serde.go
```diff
@@ -172,7 +183,16 @@ func (s *Serde) Register(id int, v any, opts ...SerdeOpt) {
...
 	at := m[k]
-	for _, idx := range t.index {
+	max := func(i, j int) int {
+		if i > j {
+			return i
+		}
+		return j
+	}
+	for i, idx := range t.index {
+		at.subindexDepth = max(at.subindexDepth, len(t.index)-i)
```
twmb, line 193: it'd be clearer to capture `myDepth := len(t.index)` just above the for loop, so this line becomes `at.subindexDepth = max(at.subindexDepth, myDepth)`

pkg/sr/serde.go
```diff
@@ -172,7 +183,16 @@ func (s *Serde) Register(id int, v any, opts ...SerdeOpt) {
 	// tree to find the end node we are initializing.
```
twmb, line 183: Can you add a `defer` before s.ids.Store(m) on line L168? I think the code is racy (before this PR)

pkg/sr/serde.go
twmb: Can you add a comment above this line:
```
// SAFETY: tserdeMapClone deeply clones all maps through the index,
// so our modified value is being saved to a new map that is not being read.
```

pkg/sr/serde.go
```diff
@@ -316,44 +323,108 @@ func (s *Serde) DecodeNew(b []byte) (any, error) {
...
-		return nil, t, ErrNotRegistered
+		return nil, tserde{}, ErrNotRegistered
 	}
 	return b, t, nil
 }
 
-type bReader struct{ b []byte }
+// SerdeHeader encodes and decodes a message header.
+type SerdeHeader interface {
+	AppendEncode(b []byte, id int, index []int) ([]byte, error)
```
twmb, line 355: // AppendEncode encodes a schema ID and optional index to b,
// returning the updated slice or an error

pkg/sr/serde.go
```diff
@@ -316,44 +323,108 @@ func (s *Serde) DecodeNew(b []byte) (any, error) {
...
+	DecodeID(in []byte) (id int, out []byte, err error)
```
twmb, line 356:
```
// DecodeID decodes an ID from in, returning the ID and the remaining bytes, or an error.
```

I think this supports some future change where an ID is stuffed in the header -- Decode would need to change, but not this function DecodeID

pkg/sr/serde.go
```diff
@@ -316,44 +323,108 @@ func (s *Serde) DecodeNew(b []byte) (any, error) {
...
+	DecodeIndex(in []byte, maxLength int) (index []int, out []byte, err error)
```
twmb, line 357: // DecodeIndex decodes at most maxLength of a schema index from in, returning the
// index and remaining bytes, or an error.

pkg/sr/serde.go
```diff
@@ -316,44 +323,108 @@ func (s *Serde) DecodeNew(b []byte) (any, error) {
...
+// alongside the unread bytes. It expects b to be the output of DecodeID (schema
+// ID should already be stripped away). If maxLength is greater than 0 and the
+// encoded data contains more indices than maxLength the function returns
+// ErrNotRegistered.
+func (ConfluentHeader) DecodeIndex(b []byte, maxLength int) ([]int, []byte, error) {
+	buf := bytes.NewBuffer(b)
+	l, err := binary.ReadVarint(buf)
+	if err != nil {
+		return nil, nil, err
+	}
```
twmb, line 414:
```
if l < 0 { return error }
```
I think currently this will panic if <0

twmb:
Sorry for the delay, finally did a deep review. I don't think I have any comments besides what I've posted now.

pkg/sr/serde.go
```diff
@@ -316,44 +323,108 @@ func (s *Serde) DecodeNew(b []byte) (any, error) {
...
+	DecodeID(in []byte) (id int, out []byte, err error)
+	DecodeIndex(in []byte, maxLength int) (index []int, out []byte, err error)
+}
+
+// ConfluentHeader is a SerdeHeader that produces the Confluent wire format. It
+// starts with 0, then big endian uint32 of the ID, then index (only protobuf),
+// then the encoded message.
+//
+// https://docs.confluent.io/platform/current/schema-registry/serdes-develop/index.html#wire-format
+type ConfluentHeader struct{}
```
twmb, line 365: Can this be private? Not sure it needs to be public at the moment -- we can default internally to the private confluent header, but then the user can override.

twmb: And, if we do that, this can be something like

```go
type tConfluentHeader struct{}

var confluentHeader = new(tConfluentHeader)
```

and then line 143 above can return `confluentHeader` the pointer variable (per the pointer methods comment below)

pkg/sr/serde.go
lovromazgon: I can do that if you feel strongly about it 😉 I'm still curious why it's worth avoiding `bytes.Buffer`? I created a [simple benchmark](https://go.dev/play/p/Giaik-DzloL) for `bReader` and `bytes.Buffer`, both seem to have virtually the same performance and no allocations.
```
go test -bench=. -benchmem .
goos: darwin
goarch: amd64
pkg: github.com/twmb/franz-go/pkg/sr
cpu: Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
BenchmarkBReader-16        	148830937	         7.766 ns/op	       0 B/op	       0 allocs/op
BenchmarkBytesBuffer-16    	149000121	         7.886 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/twmb/franz-go/pkg/sr	4.065s
```

pkg/sr/serde.go
lovromazgon: Ugh, that would defeat the purpose of why I added this in the first place. I'd like to have access to the header and decode the ID manually (see my example in the PR description). Do you think there's a better way to achieve this? Another way would be to create the method `func (*Serde) DecodeID([]byte) (int, error)` which passes the call to the underlying `SerdeHeader`.

pkg/sr/serde.go
lovromazgon: I made the type private, but used different names to avoid the `t` prefix, hope that's fine:
```
type confluentHeader struct{}

var defaultSerdeHeader = new(confluentHeader)
```

Although if that's a convention you are using for private types (i.e. `tserde`, `tConfluentHeader` etc.) let me know.

pkg/sr/serde.go
lovromazgon: It wasn't as simple, because we change `m` below. If we are registering a schema with indices, `m` ends up pointing to the last map of subindices.

I added a variable `dupIDs` to match `dupTypes` and store both of them in the same deferred call.

pkg/sr/serde.go
lovromazgon: Sure, it would make it easy to add more options in the future. I think naming could be better though. IMO `SerdeOpt` sounds like a better fit for configuring `Serde` (i.e. global options), while the current options could be something like `RegisterOpt` (since they are used in `Register`). This naming would play well with the changes I proposed in the [other PR](https://github.com/twmb/franz-go/pull/466), which introduces `EncodeOpt`, we end up with:
- `SerdeOpt` configures `Serde`
- `RegisterOpt` configures `Serde.Register`
- `EncodeOpt` configures `Serde.Encode`

While we're at it, it might make sense to change `Opt` to `ClientOpt` and introduce `NewSerde(opts ...SerdeOpt)` to match `NewClient`. WDYT? I can remove `SetHeader` from this PR and introduce the new options in a separate PR, to not increase the scope too much.

lovromazgon:
Thanks for the thorough review! 👍 I tried to address all comments in one way or another.

pkg/sr/serde.go
twmb: The t in this case means type, I do like it but I'm ok either way.

pkg/sr/serde.go
twmb: * +1 to adding NewSerde, which also removes the need for SetDefaults (so that can / should be removed).
* +1 to changing Opt to ClientOpt
* +1 to changing the current register opts to RegisterOpt, and then changing the new opt introduced here to SerdeOpt _however_ all RegisterOpts need to implement SerdeOpt so that they can be used as defaults.
* I'm not sure about the difference between RegisterOpt and EncodeOpt. I need to look at the other PR. I remember glancing at it and not being sure why it was split. Also I wonder if EncodingOpt is better than both RegisterOpt and EncodeOpt (although I could favor EncodeOpt for brevity)

pkg/sr/serde.go
twmb: I don't feel too strongly since the benchmarks show things are the same

pkg/sr/serde.go
```diff
@@ -132,15 +132,15 @@ func (s *Serde) SetDefaults(opts ...SerdeOpt) {
 }
 
 // SetHeader configures which header should be used when encoding and decoding
-// values. If the header is set to nil it falls back to ConfluentHeader.
+// values. If the header is set to nil it falls back to confluentHeader.
```
twmb: `back to the default Confluent header.`

pkg/sr/serde.go
```diff
@@ -163,26 +163,28 @@ func (s *Serde) Register(id int, v any, opts ...SerdeOpt) {
...
-	m := tserdeMapClone(s.loadIDs(), id, t.index)
-	s.ids.Store(m)
-
 	// Now we have a full path down index initialized (or, the top
 	// level map if there is no index). We iterate down the index
 	// tree to find the end node we are initializing.
 	k := id
-	at := m[k]
+	m := dupIDs
+	at := dupIDs[k]
```
twmb, line 181: this line could have stayed the same given the prior line, fwiw :)

pkg/sr/serde.go
twmb: Oh I see now. Would you be ok with `DecodeID`? That's a bit closer in spirit to the current API choices of the package, but I'd be ok with `Header()` if need be. There doesn't seem to be any purpose to exposing a getter for the header _besides_ decoding the ID of a slice -- I can't imagine a person will want to get the header to append an ID / I don't know what a use case for that could be. If you can think of a use case for Header besides ID decoding, I'm +1 to keeping the getter, otherwise I'm on the fence.

pkg/sr/serde.go
twmb: Also I'm sorry this has been such a slow review cycle, to put it short: I was involved with something that was time consuming sun up to sun down and also went on for ~3w (coincidentally, right when you opened the PR) -- and I'm trying to not be inside on the weekends.

lovromazgon: No worries! I'll get back to this PR in a day or two 👍

twmb: I might have time on Thursday to take this over if you're not able to get to it (I admit that this review cycle has gone on for ... quite a while).

lovromazgon: So the situation turned around, because I'm pretty busy now and I'm not sure I'll get a chance to look into this before the weekend. If you get the time, feel free to take over the work that's left.

pkg/sr/serde.go
```diff
@@ -134,6 +131,26 @@ func (s *Serde) SetDefaults(opts ...SerdeOpt) {
 	s.defaults = opts
 }
 
+// DecodeID decodes an ID from in, returning the ID and the remaining bytes,
```
twmb, line 134: `decodes an ID from b` but I can sneak this into a quickfix PR that I plan to do

twmb:
This PR LGTM, I say squash and merge it.
This is now missing the ability to set a header but I think per this comment chain: https://github.com/twmb/franz-go/pull/467#discussion_r1231620122 that will be in a follow up PR.

lovromazgon:
Sounds good to me! I'll take care of the follow-up in a second. I already have a separate branch with the changes, I just had something urgent come up so I didn't get to open the PR, sorry about that.

PR #469, opened 2023-06-06
Add the ability to unregister metrics.

freeformz:
We have a program that creates a new kgo client for each long runnning request to the service and need a way to have unique metrics per long running client/connection, but don't want the metrics to hand around and be reported after the client is closed.

This allows us to unregister the metrics after the client is closed.

### DIFF
--- a/plugin/kprom/kprom.go
+++ b/plugin/kprom/kprom.go
@@ -298,3 +298,20 @@ func (m *Metrics) OnFetchBatchRead(meta kgo.BrokerMetadata, topic string, _ int3
 	node := strnode(meta.NodeID)
 	m.fetchBytes.WithLabelValues(node, topic).Add(float64(fbm.UncompressedBytes))
 }
+
+// Unregister the metrics if a registry was provided or created. Returns true if all metrics were unregistered.
+func (m *Metrics) Unregister() bool {
+	if m.cfg.reg == nil {
+		return false
+	}
+
+	return m.cfg.reg.Unregister(m.connects) &&
+		m.cfg.reg.Unregister(m.connectErrs) &&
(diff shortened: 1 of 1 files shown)


plugin/kprom/kprom.go
```diff
@@ -298,3 +298,21 @@ func (m *Metrics) OnFetchBatchRead(meta kgo.BrokerMetadata, topic string, _ int3
...
+
+	a := m.cfg.reg.Unregister(m.connects)
+	b := m.cfg.reg.Unregister(m.connectErrs)
+	c := m.cfg.reg.Unregister(m.disconnects)
+	d := m.cfg.reg.Unregister(m.writeErrs)
+	e := m.cfg.reg.Unregister(m.writeBytes)
+	f := m.cfg.reg.Unregister(m.readErrs)
+	g := m.cfg.reg.Unregister(m.readBytes)
+	h := m.cfg.reg.Unregister(m.produceBytes)
+	i := m.cfg.reg.Unregister(m.fetchBytes)
```
twmb, line 316: More future proof might be something like,
```go
var failed bool
for _, unreg := range []prometheus.Collector {
    ....
} {
    failed = failed || m.cfg.reg.Unregister(unreg)
}
return failed
```
but this bit is fine with me for now

freeformz:
hmm. With this I still get a panic when another client comes on. Could be a race between deregistration and the new client being setup.

freeformz:
The WithClientLabel suggested in https://github.com/twmb/franz-go/pull/441 would be a better solution. I'm going to hack something locally for now.

freeformz:
Actually, the problem was the failure to unregister the GaugeFuncs registered in OnNewClient.

PR #470, opened 2023-06-07
Add logrus plugin

JGShaw:
This PR adds a logging plugin for using [logrus](https://github.com/sirupsen/logrus).


plugin/klogrus/klogrus.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// New returns a new Logger.
+func New(lr *logrus.Logger) *Logger {
+	return &Logger{lr}
+}
+
+// Level is for the kgo.Logger interface.
+func (l *Logger) Level() kgo.LogLevel {
+	return logrusToKgoLevel(l.lr.Level)
```
twmb, line 21: Can this use GetLevel? It looks like Level can be updated atomically with SetLevel

JGShaw:
Ive updated the code to use `GetLevel` as suggested, it makes sense to use that. Thank you!

I also spotted an error where I had not initialized the empty field map so I have added a fix for that as well.

plugin/klogrus/klogrus.go
```diff
@@ -0,0 +1,63 @@
...
+	return logrusToKgoLevel(l.lr.GetLevel())
+}
+
+// Log is for the kgo.Logger interface.
+func (l *Logger) Log(level kgo.LogLevel, msg string, keyvals ...any) {
+	logrusLevel, levelMatched := kgoToLogrusLevel(level)
+	if levelMatched {
+		fields := make(logrus.Fields, len(keyvals)/2)
+		for i := 0; i < len(keyvals); i += 2 {
+			fields[keyvals[i].(string)] = keyvals[i+1]
```
twmb, line 30: I'm going to change this to `fmt.Sprint(keyvals[i])` in a last-minute PR just for safety

PR #478, opened 2023-06-23
kfake: Expose the current broker's node ID during request interception

jan-g:
In order to preserve the existing API, we add a `RequestNodeId`, valid only during a control function, which returns the node id of the broker that the current request was attached to.

### DIFF
--- a/pkg/kfake/cluster.go
+++ b/pkg/kfake/cluster.go
@@ -35,6 +35,7 @@ type (
 		controlMu          sync.Mutex
 		control            map[int16][]controlFn
 		keepCurrentControl atomic.Bool
+		currentBroker      atomic.Pointer[broker]
 
 		data   data
 		pids   pids
@@ -226,7 +227,7 @@ func (c *Cluster) run() {
 		}
 
 		kreq := creq.kreq
-		kresp, err, handled := c.tryControl(kreq)
+		kresp, err, handled := c.tryControl(kreq, creq.cc.b)
 		if handled {
 			goto afterControl
 		}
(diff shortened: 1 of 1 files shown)


pkg/kfake/cluster.go
```diff
@@ -353,7 +354,14 @@ func (c *Cluster) KeepControl() {
...
 
-func (c *Cluster) tryControl(kreq kmsg.Request) (kresp kmsg.Response, err error, handled bool) {
+// RequestNodeId is solely valid from within a control function; it returns
+// the broker id that the request was received by.
+func (c *Cluster) RequestNodeId() int32 {
+	return c.currentBroker.Load().node
+}
+
+func (c *Cluster) tryControl(kreq kmsg.Request, b *broker) (kresp kmsg.Response, err error, handled bool) {
+	c.currentBroker.Store(b)
```
twmb, line 364: Can this `defer c.currentBroker.Store(nil)`?

pkg/kfake/cluster.go
```diff
@@ -353,7 +354,14 @@ func (c *Cluster) KeepControl() {
 	c.keepCurrentControl.Swap(true)
 }
 
-func (c *Cluster) tryControl(kreq kmsg.Request) (kresp kmsg.Response, err error, handled bool) {
+// RequestNodeId is solely valid from within a control function; it returns
+// the broker id that the request was received by.
+func (c *Cluster) RequestNodeId() int32 {
+	return c.currentBroker.Load().node
```
twmb, line 360: Can this return -1 if Load() returns nil? Currently this will panic if called too early

PR #505, opened 2023-07-08
kprom: add options to significantly expand metrics gathering

twmb:


twmb:
https://github.com/twmb/franz-go/pull/441#issuecomment-1627396033

brunsgaard:
In regard to:

>  I'm a bit worried about installing histograms to the factory by default but I think they don't show up because they aren't ever incremented.

Metrics are not collected and exported before a the first call to `.Add`, `.Incr`, `.Observe`, so we should be good 👍

twmb:
* I did `go doc -all` for master and brunsgaard:brun/master (this branch), there are no api incompatibilities
* I ran a produce & fetch locally and diffed the output from the prom handler, there were no metric naming differences
* I was a bit worried that registering/unregistering always in OnNewClient and OnClientClosed might prove problematic because a person might start two clients at once -- but we would have already ran into problems today because OnNewClient already registers two gauges

AFAICT this LGTM. I'm going to merge and tag. Thank you, this is a great addition.

twmb:
Tagged as plugin/kprom/v1.1.0

brunsgaard:
Your last point on unregistration is good, and I did not think about that. If you have multiple clients and one of them close, what happens to the metrics for the remaining clients, my feeling is that there will be unintended behavior.

twmb:
I'd expect that promauto causes panics today because of double registration, though?

PR #506, opened 2023-07-09
Serde options refactor

lovromazgon:
This is a follow-up to https://github.com/twmb/franz-go/pull/467#discussion_r1231620122.

This PR includes the following changes:

- Rename `Opt` to `ClientOpt`
- Rename `SerdeOpt` to `EncodingOpt`
- Introduce a new `SerdeOpt`, which will ultimately be used to set the `SerdeHeader` (this can only be included once https://github.com/twmb/franz-go/pull/467 is merged)
- Introduce the encoding option `ID` which lets you specify which schema ID to use when encoding (specifically useful when encoding dynamic data)


pkg/sr/serde.go
```diff
@@ -415,34 +415,61 @@ func (s *Serde) DecodeNew(b []byte) (any, error) {
...
-	if len(t.subindex) > 0 {
-		r := bReader{b}
-		br := io.ByteReader(&r)
-		l, err := binary.ReadVarint(br)
-		if l == 0 { // length 0 is a shortcut for length 1, index 0
-			t = t.subindex[0]
+	var t tserde
+	if idopt != nil {
+		// Load tserde based on the supplied ID.
+		t = s.loadIDs()[int(idopt.ID())]
```
lovromazgon, line 427: The last commit adds the ability to supply `EncodingOpt` to `Decode`, simply to have parity with `Encode`, but I'm not convinced if it makes sense. I included it to run it by you, but I can revert the change if needed.

This is how it works. If you supply options `ID` and `Index` to `Decode` it will find `tserde` based solely on the supplied options, it will not decode the header. In fact, it will assume there is _no header_ and that it's dealing only with the raw encoded bytes. What I don't like about this is that `Serde` can't encode bytes without a header unless you specify your own `SerdeHeader` implementation which skips prepending the header. It would allow you to track the ID and index separately from the encoded payload, although I'm not sure such a use case exists.

pkg/sr/serde.go
```diff
@@ -221,17 +284,17 @@ func tserdeMapClone(m map[int]tserde, at int, index []int) map[int]tserde {
...
 }
 
 // AppendEncode appends an encoded value to b according to the schema registry
 // wire format and returns it. If EncodeFn was not used, this returns
 // ErrNotRegistered.
-func (s *Serde) AppendEncode(b []byte, v any) ([]byte, error) {
-	t, ok := s.loadTypes()[reflect.TypeOf(v)]
-	if !ok || (t.encode == nil && t.appendEncode == nil) {
-		return b, ErrNotRegistered
+func (s *Serde) AppendEncode(b []byte, v any, opts ...EncodingOpt) ([]byte, error) {
```
twmb, line 294: What's the point of having options at the point of encoding?
Same thought on decoding.
I'd like to avoid options on these functions if possible -- would it be possible to have a block of `Dyn(Append)?Encode` functions?

lovromazgon: Hey @twmb! First of all, sorry, this work fell through the cracks. I'll try to find some time over the weekend and clean up the code. IIUC you're proposing to have `DynEncode` and `DynAppendEncode` that take `EncodingOpt` and leave the existing encode functions as they are, correct? I'll try it out.

lovromazgon: As promised, I added the changes. It simplified the code quite a bit - no more `ID` option, no more special treatment of any option, no more options passed to encode and decode. Instead we now have `DynEncode` and related functions, which expect parameters `id` and `index` to determine the correct `tserde`.

I'm looking forward to hear your feedback 😉

twmb: Random question now that I'm seeing this PR:

Is there any reason to require registering the types at all and then using DynEncode, vs. having top-level, non Serde method functions,

```go
func Encode(id int, index []int, v any, enc func(any) ([]byte, error)) ([]byte, error)
func AppendEncode(b []byte, id int, index []int, v any, enc (func([]byte, any) ([]byte, error)) ([]byte, error)
```
?

I'm wondering: what use case do you have that allows you to know the id and index you want to encode, but not know the encode function?

lovromazgon: Oh well, this was another long delay, sorry about that. You know, it's the standard apology, being busy. I'll try to be more responsive now and help you resolve this PR.

I merged the changes from master and resolved the conflicts introduced in the PR you mentioned.

> what use case do you have that allows you to know the id and index you want to encode, but not know the encode function?

It's not that I don't know the encode function, I do know it actually. It's that `Serde` also appends the header and I don't want to duplicate that logic outside of `Serde`. Another reason to register the type is that the next time the type gets decoded in a different path of the code, we already have a `Serde` that knows how to decode it, no need to fetch the schema.

twmb: No worries on the delay, I'm just as slow in the past few months due to games and traveling and holidays and work :)

I'm not sure I still understand but I'm sure I would if I see what you're trying to do -- do you have an example of how you intend to use the dyn functions that can't be served with the current API?

lovromazgon: Here's [the line](https://github.com/ConduitIO/conduit/blob/dbe5b8618ed92adb10650496fb73dee81146c39b/pkg/processor/schemaregistry/encoder.go#L53) where we call `Serde.Encode` and provide an ID of the schema to use. (For completeness, [here's](https://github.com/ConduitIO/conduit/blob/92d9681ed66abd17d5bef2b10ec434ee89fba818/pkg/processor/schemaregistry/decoder.go#L42) `Serde.Decode`).

I re-read your suggestion a few comments above and I think that having top-level, non Serde functions for `Encode` and `AppendEncode` that let you specify the ID and the encode func could work. The calls wouldn't go through the `Serde` type, but I realized we don't reuse the same object in the decode and encode paths of the code, so no harm done. And the solution feels cleaner than having `DynEncode`.

I'll see if I can get some time in the next few weeks to work this into the PR.

pkg/sr/serde.go
```diff
@@ -258,28 +221,31 @@ func (s *Serde) Encode(v any) ([]byte, error) {
...
-	if t.appendEncode != nil {
-		return t.appendEncode(b, v)
-	}
-	encoded, err := t.encode(v)
-	if err != nil {
-		return nil, err
+	appendEncode := t.appendEncode
+	if appendEncode == nil {
+		// Fallback to t.encode.
+		appendEncode = func(b []byte, v any) ([]byte, error) {
```
twmb, line 239: I think this causes an extra alloc in the encode path.

twmb (commented):
I think I'm +1 on the changes now, minus the one comment about AppendEncode. I'll see about rebasing your branch on the latest release and see about getting this merged.

pkg/sr/serde.go
```diff
@@ -258,28 +221,31 @@ func (s *Serde) Encode(v any) ([]byte, error) {
...
+// registered, this returns ErrNotRegistered.
 func (s *Serde) AppendEncode(b []byte, v any) ([]byte, error) {
-	t, ok := s.loadTypes()[reflect.TypeOf(v)]
-	if !ok || (t.encode == nil && t.appendEncode == nil) {
-		return b, ErrNotRegistered
-	}
+	// Load tserde based on the registered type.
+	t := s.loadTypes()[reflect.TypeOf(v)]
 
-	b, err := s.header().AppendEncode(b, int(t.id), t.index)
```
twmb, line 270: Looks like this was lost (and not added in one of the new top level functions)

twmb: Oh I see, it's used in the new call to AppendEncode below.

twmb: I might rather have a little bit of duplication, not sure yet.

twmb: Yep leaning duplication; I rebased on master and squashed all of your commits and did a few touchups in #742.

twmb:
Merged in #742, ty for the work (and patience over time 😅  😬)

I think the main thing I want to think about before stabilizing the sr package is whether the top level Encode and AppendEncode functions should take all their args as fields, vs. some form that allows for future growth. I'm currently leaning to keep the current form (though maybe swap the `h` argument before `id` and `index`).

twmb:
@lovromazgon I'm swapping SerdeHeader before id,index in #743

PR #530, opened 2023-08-04
examples,plugin,kerberos,kfake,kadm: ./bump-deps.sh

twmb:
No real notable changes for kerberos,kfake,kadm


twmb:
Merging, tests pass locally (no change to franz-go) -- CI container setup seems to be failing at the moment, as it does intermittently.

PR #541, opened 2023-08-24
Fix typo in parseReadSize

alistairking:
Previously the check was against `maxSize`, which comes from the consumer configuration, so there's no point checking that to see if it is "HTTP".


twmb:
Very good find, and this probably explains a not-as-expected error message I've seen in the past (but didn't pay much attention to at the time). I might leave this open for a bit while I decide whether to issue this as a patch release alone or wait for something else.

twmb:
Scratch that I fixed the lint in a separate PR. I'll re-run CI

PR #554, opened 2023-09-01
Add DeltaStream to README.md

curtisr7:
DeltaStream uses franz-go internally for all go microservices.


twmb (approved):
Awesome, thank you!

PR #577, opened 2023-09-21
pkg/sr: use ptr type for SchemaMetadata and SchemaRuleSet

shubham030:
`omitempty` doesn't work for structs, as a result, the schema contains zero value of SchemaMetadata and SchemaRuleSet which fails to create the schema on redpanda.

https://pkg.go.dev/encoding/json#Marshal

### DIFF
--- a/pkg/sr/api.go
+++ b/pkg/sr/api.go
@@ -80,10 +80,10 @@ type (
 		References []SchemaReference `json:"references,omitempty"`
 
 		// SchemaMetadata is arbitrary information about the schema.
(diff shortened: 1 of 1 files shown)


pkg/sr/api.go
```diff
@@ -667,6 +667,7 @@ func (cl *Client) ResetCompatibility(ctx context.Context, subjects ...string) []
 		go func() {
 			defer wg.Done()
 			var c SetCompatibility // unmarshals with "compatibility"
+			fmt.Println("set comp", c)
```
twmb, line 670: pointers lgtm, but please remove this :D

Issue #559, opened 2023-09-13
setting offset metadata in a transaction


iamnoah:
I'm looking to add offset metadata to an offset commit in a transaction. In the past, with the Java client, I've used the offset metadata to store the offset into a non-Kafka data source that I'm copying from. 

I found `kgo.PreCommitFnContext` which lets me set the metadata for a normal consumer offset commit, but I don't see anything analogous for a `GroupTransactSession`, in fact I noticed that the metadata for a transactional offset commit is always set to the member ID.

So what I'm looking for is:

 0. Load offsets with metadata, then loop:
 1. Start a transaction
 2. Produce some records loaded from a data source (using the loaded metadata as my cursor)
 3. Commit offsets with metadata that records my position in the data source
 4. End transaction

3 is where I'm stuck. My motivation here is to get exactly once semantics since the data source position and records are committed together. I've omitted rebalance handling, etc.

Is there a way to do this that I'm missing? If not, would a transactional hook like `PreCommitFnContext` for transactions be a welcome PR?

twmb:
I'm +1 to adding support for PreCommitFnContext to GroupTransactSession's End function -- this would be a good addition and a good place to annotate the commit message. Unfortunately due to API constraints -- the precommit function currently takes kmsg.OffsetCommitRequest whereas you're going to want kmsg.TxnOffsetCommitRequest -- we'll need to add a new PreTxnCommitFnContext.

lmk if you'd like to add support for it.

iamnoah:
@twmb thanks for adding this!

Trying it out, the missing piece is that the topic I am committing offsets against may not have changed offsets to commit, but I still want to commit and modify the metadata because my position in my external source has changed. The least bad way seemed like another hook to force specific offsets to be committed (see linked PR,) but I'm not picky about the exact API, as long as I can force a commit with new metadata as part of a transaction.

twmb:
What do you think about instead changing the behavior in the source: you can modify the input request, add your offsets, and then update `postcommit` based on the TxnOffsetCommit response on current line 291? This would also avoid adding a new API (and I'm less keen to bump a minor version for a bit), and instead issue this as a patch release?

iamnoah:
@twmb responded in the linked ticket.

PR #578, opened 2023-09-21
kgo: add TxnForceOffsets

iamnoah:
Needed to allow changes to the offset Metadata without a change to Offset.

This is the missing piece for #559

### DIFF
--- a/pkg/kgo/txn.go
+++ b/pkg/kgo/txn.go
@@ -208,6 +208,15 @@ func (s *GroupTransactSession) failed() bool {
 	return s.revoked || s.lost
 }
 
+type txnForceOffsetsT struct {}
+
+var txnForceOffsets txnForceOffsetsT
+
+// TxnForceOffsets forces GroupTransactSession.End to commit with the given offsets, even if nothing has changed.
+func TxnForceOffsets(ctx context.Context, uncommitted map[string]map[int32]EpochOffset) context.Context {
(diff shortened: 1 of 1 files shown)


twmb:
Copying from my reply on the issue:

What do you think about instead changing the behavior in the source: you can modify the input request, add your offsets, and then update `postcommit` based on the TxnOffsetCommit req+resp on current line 291? This would also avoid adding a new API (and I'm less keen to bump a minor version for a bit), and instead issue this as a patch release?

iamnoah:
@twmb I'm not 100% sure I understand what you are proposing. The issue I ran into is that `PreTxnCommitFnContext` never has a chance to modify the request if [uncommitted is unchanged.](https://github.com/twmb/franz-go/blob/master/pkg/kgo/txn.go#L1063) Is there a way I can force a change in uncommitted that will always result in valid offsets?

The problem seems to arise because the uncommitted offsets API is only concerned with the numerical offsets and not the metadata. I don't see a way to force `End` to actually make a commit if I only want to modify the metadata. In some cases I could do something hacky like increment/decrement by 1, but in the case of a new, empty topic, I run into the problem that there may not be another valid offset to update to (and there are probably other reasons trying to hack around it like that is ill advised.)

iamnoah:
@twmb #580 does it without new API and seems fine in my testing, however I did have to reorder at least the produceID load, so I'm not sure how safe a change this is.

PR #580, opened 2023-09-26
kgo: allow PreTxnCommitFnContext to modify empty offsets

iamnoah:
This builds the TxnOffsetCommitRequest early so that the hook can modify it. If the modified request has no topics to commit, then we abort as though uncommitted was empty. Conversely, the hook can add partitions to an otherwise empty commit to enable setting metadata, or even just refreshing the offset commit.

This moves initialization of the producerID before addOffsetsToTxn. I think that should not matter, but it's a non-trivial change in internal behavior.

### DIFF
--- a/pkg/kgo/txn.go
+++ b/pkg/kgo/txn.go
@@ -8,8 +8,9 @@ import (
 	"sync"
 	"time"
 
-	"github.com/twmb/franz-go/pkg/kerr"
 	"github.com/twmb/franz-go/pkg/kmsg"
+
+	"github.com/twmb/franz-go/pkg/kerr"
 )
 
 // TransactionEndTry is simply a named bool.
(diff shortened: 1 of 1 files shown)


twmb:
This is similar to the approach I was thinking, but I saw a bunch of things to be careful about. I'll review this hopefully this weekend.

iamnoah:
@twmb bump. FWIW, I've been running with this in production for a few weeks without observing any issues.

twmb:
Yes sorry, got a bit delayed due to buying a game... aim is this week.

pkg/kgo/txn.go
```diff
@@ -1060,7 +1061,13 @@ func (cl *Client) commitTransactionOffsets(
 		onDone(kmsg.NewPtrTxnOffsetCommitRequest(), kmsg.NewPtrTxnOffsetCommitResponse(), errNotGroup)
 		return nil
 	}
-	if len(uncommitted) == 0 {
+
+	req, err := g.prepareTxnOffsetCommit(ctx, uncommitted)
```
twmb, line 1065: This needs to be in g.mu because building the request uses the g.generation and g.memberID fields, which are updated after join group (within a lock). Need to go, will look at this in a bit to see where it can go exactly unless you eyeball this very closely.

twmb:
I've taken this commit and opened #605 which leads with a commit making g.memberID and g.generation a single atomic value -- eliminating the concurrency issue I mentioned. Closing this in favor of 605

PR #596, opened 2023-10-13
feat(kslog): implemented kgo.Logger plugin with slog

sbuliarca:
This is a kgo.Logger wrapping [slog](https://pkg.go.dev/log/slog).
It will be really useful to us, as we're switching to it, so I though it might help others too.

slog is not experimental any more only from go 1.21, so whoever wants to use this plugin will need to use go version >= 1.21

### DIFF
--- a/plugin/kslog/README.md
+++ b/plugin/kslog/README.md
@@ -1,0 +1,13 @@
+kslog
+===
+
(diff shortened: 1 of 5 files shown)


twmb (approved):
Looks easy enough, first time I've seen slog in actual use... thanks!

PR #601, opened 2023-10-21
kgo: do not rotate the consumer session when pausing topics/partitions

### LINKED ISSUE #585: Pausing/Resuming partitions causes lots of errors
Hi,

I have noticed a problem with pausing and resuming of partitions that is causing lots of errors. When I stopped doing the pause/resume, the errors went away. Here is a screenshot of the Franz prometheus metrics when the application was actively pausing and resuming partitions:
pause_errors

Here is a screenshot after commenting out the pause/resume keeping everything else the same:
stop_pause

Here is a sample summary of errors during a 5 minute period from a pod:
```
 [
   {
    "msg": "read Fetch v6",
    "err": "context canceled",
    "count": 827
  },
  {
    "msg": "unable to open connection to broker",
    "err": "context canceled",
    "count": 117
  },
  {
    "msg": "wrote Fetch v6",
    "err": "context canceled",
    "count": 4
  },
  {
    "msg": "wrote Fetch v6",
    "err": "use of closed network connection",
    "count": 18
  },
  {
    "msg": "wrote Fetch v6",
    "err": "write tcp 10.73.0.182:35770->xx.xx.147.1:9093: use of closed network connection",
    "count": 1
  },
  {
    "msg": "wrote Fetch v6",
    "err": "write tcp 10.73.1.0:45126->xx.xx.147.1:9093: use of closed network connection",
    "count": 1
  },
  {
    "msg": "wrote Fetch v6",
    "err": "write tcp 10.73.1.0:54114->xx.xx.147.1:9093: use of closed network connection",
    "count": 1
  }
]

### PR DESCRIPTION
**Pull request by twmb**
Issue #489 asked to stop returning data after a partition was paused -- the original implementation of pausing kept returning any data that was in flight or already buffered, and simply stopped fetching new data.

489 was dealt with by bumping the consumer session, which kills all in flight fetch requests. This was easy, but can cause a lot of connection churn if pausing and resuming a lot -- which is #585.

The new implementation allows fetches to complete, but strips data from fetches based on what is paused at the moment the fetches are being returned to the client. This does make polling paused fetches very slightly slower (a map lookup per partition), but there's only so much that's possible. If a partition is paused, we drop the data and do not advance the internal offset. If a partition is not paused, we keep the data and return it -- same as before.

Closes #585.

### DIFF
--- a/pkg/kgo/source.go
+++ b/pkg/kgo/source.go
@@ -344,8 +344,64 @@ func (s *source) hook(f *Fetch, buffered, polled bool) {
 }
 
 // takeBuffered drains a buffered fetch and updates offsets.
-func (s *source) takeBuffered() Fetch {
-	return s.takeBufferedFn(true, usedOffsets.finishUsingAllWithSet)
+func (s *source) takeBuffered(paused pausedTopics) Fetch {
+	if len(paused) == 0 {
+		return s.takeBufferedFn(true, usedOffsets.finishUsingAllWithSet)
+	}
+	var strip map[string]map[int32]struct{}
+	f := s.takeBufferedFn(true, func(os usedOffsets) {
+		for t, ps := range os {
+			// If the entire topic is paused, we allowUsable all
+			// and strip the topic entirely.
+			pps, ok := paused.t(t)
+			if !ok {
+				continue
+			}
+			if strip == nil {
+				strip = make(map[string]map[int32]struct{})
+			}
+			if pps.all {
+				for _, o := range ps {
+					o.from.allowUsable()
+				}
+				strip[t] = nil // initialize key, for existence-but-len-0 check below
+				continue
+			}
+			stript := make(map[int32]struct{})
+			strip[t] = stript
+			for _, o := range ps {
+				if _, ok := pps.m[o.from.partition]; ok {
+					o.from.allowUsable()
+					stript[o.from.partition] = struct{}{}
+					continue
+				}
+				o.from.setOffset(o.cursorOffset)
+				o.from.allowUsable()
+			}
+		}
+	})
+	if strip != nil {
+		keep := f.Topics[:0]
+		for _, t := range f.Topics {
+			stript, ok := strip[t.Topic]
+			if ok {
+				if len(stript) == 0 {
+					continue // stripping this entire topic
+				}
+				keepp := t.Partitions[:0]
+				for _, p := range t.Partitions {
+					if _, ok := stript[p.Partition]; ok {
+						continue
+					}
+					keepp = append(keepp, p)
+				}
+				t.Partitions = keepp
+			}
+			keep = append(keep, t)
+		}
+		f.Topics = keep
+	}
+	return f
 }
 
 func (s *source) discardBuffered() {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: pkg/kgo/source.go [hunk 1]
```diff
@@ -344,8 +344,38 @@ func (s *source) hook(f *Fetch, buffered, polled bool) {
 }
 
 // takeBuffered drains a buffered fetch and updates offsets.
-func (s *source) takeBuffered() Fetch {
-	return s.takeBufferedFn(true, usedOffsets.finishUsingAllWithSet)
+func (s *source) takeBuffered(paused pausedTopics) Fetch {
+	var strip mtmps
```
stampy88, line 348: what do you think about returning early if `paused == nil`? It remove that after of the `eachOffset` "loop", e.g.
``` 
if paused == nil {
    return s.takeBufferedFn(true, usedOffsets.finishUsingAllWithSet)
}

var strip mtmps
f := s.takeBufferedFn(true, func(os usedOffsets) {
		os.eachOffset(func(o *cursorOffsetNext) {
			if paused.has(o.from.topic, o.from.partition) {
				o.from.allowUsable()
				strip.add(o.from.topic, o.from.partition)
				return
			}
			o.from.setOffset(o.cursorOffset)
			o.from.allowUsable()
		})
	})

**Comment by twmb** [comment]
Was changing this as you commented, the newest version is the least allocating and iterating, but a good bit more confusing

**Comment by twmb** [comment]
Although returning even early like your suggestion would avoid a few conditionals in the new version

File: pkg/kgo/source.go [hunk 2]
```diff
@@ -344,8 +344,68 @@ func (s *source) hook(f *Fetch, buffered, polled bool) {
 }
 
 // takeBuffered drains a buffered fetch and updates offsets.
-func (s *source) takeBuffered() Fetch {
-	return s.takeBufferedFn(true, usedOffsets.finishUsingAllWithSet)
+func (s *source) takeBuffered(paused pausedTopics) Fetch {
+	var strip map[string]map[int32]struct{}
+	f := s.takeBufferedFn(true, func(os usedOffsets) {
+		for t, ps := range os {
+			// If the entire topic is paused, we allowUsable all
```
stampy88, line 351: if you checked whether the topic was paused some way at the top here, you could get rid of some of the conditionals below. For example (may want to add a method or two to `pausedPartitions` if you didn't want to access `all` or `m`) :
``` 
pp, ok := pausedTopics[t]
if !ok {
	continue
}

// you know something is paused at this point
if strip == nil {
	strip = make(map[string]map[int32]struct{})
}

if pp.all {
	for _, o := range ps {
		o.from.allowUsable()
	}
	strip[t] = nil // initialize key, for existence-but-len-0 check below
	continue
}

// you know that specific partitions are paused at this point
stript := make(map[int32]struct{})
for _, o := range ps {
	if _, ok := pp.m[o.from.partition]; ok {
		o.from.allowUsable()
		stript[o.from.partition] = struct{}{}
		continue
	}
	o.from.setOffset(o.cursorOffset)
	o.from.allowUsable()
}
strip[t] = stript

twmb: Not entirely -- it's possible that you paused "foo", but this fetch contains only data for "bar". This block of checking/stripping runs on every fetch request until you unpause everything, _but_ once you pause something, what you paused is also not added to fetch requests. This stripping is only necessary basically for what was in flight at the time you added something to be paused.

pkg/kgo/source.go
twmb: Need to check len 0 though rather than nil since paused is never entirely nilled out once it is initialized; this commit has been amended.

pkg/kgo/source.go
twmb: Yes this logic is better, changing.

stampy88 (approved):

PR #619, opened 2023-11-17
use interface instead *logrus.Logger in plugin klogrus

peczenyj:
Hello

We use some wrappers around `*logrus.Logger` and we don't want to write a wrapper each time we need to set the kafka logger

Instead, since most of the time we use `logrus.FieldLogger` with some extra methods, I decided to send this small contribution where we can use this _minimal_ interface

```go
// FieldLogger interface.
type FieldLogger interface {
	logrus.FieldLogger
	GetLevel() logrus.Level
}
```

The result is still compatible with logrus, but on steroids.

I was thinking in use the pure `logrus.FieldLogger` and store the log level (using two constructors, for instance) but I think we need to react to possible changes in the log level in runtime. This solution is cleaner

Also I add one test based on kslog plugin.

Enjoy

### DIFF
--- a/plugin/klogrus/klogrus.go
+++ b/plugin/klogrus/klogrus.go
@@ -7,15 +7,35 @@ import (
 	"github.com/twmb/franz-go/pkg/kgo"
 )
 
+var (
+	_ FieldLogger = (*logrus.Logger)(nil)
+	_ kgo.Logger  = (*Logger)(nil)
+)
+
+// FieldLogger interface combines logrus.FieldLogger with GetLevel method
+// useful to represent a wrapper around *logrus.Logger.
+type FieldLogger interface {
+	logrus.FieldLogger
+	GetLevel() logrus.Level
+}
+
 // Logger provides the kgo.Logger interface for usage in kgo.WithLogger when
 // initializing a client.
 type Logger struct {
-	lr *logrus.Logger
+	lr FieldLogger
 }
 
-// New returns a new Logger.
+// New returns a new Logger using a *logrus.Logger instance.
 func New(lr *logrus.Logger) *Logger {
 	return &Logger{lr}
+}
+
+// NewFieldLogger returns a new Logger using a FieldLogger interface.
+// it is isofunctional with New constructor, except it can accept either
+// *logrus.Logger or a possible wrapper that implements logrus.FieldLogger
+// and includes GetLevel method.
+func NewFieldLogger(fl FieldLogger) *Logger {
+	return &Logger{fl}
 }
 
 // Level is for the kgo.Logger interface.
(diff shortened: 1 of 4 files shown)


peczenyj:
Sure, will work on it

peczenyj:
done @twmb

plugin/klogrus/klogrus.go
```diff
@@ -45,6 +61,8 @@ func kgoToLogrusLevel(level kgo.LogLevel) (logrus.Level, bool) {
 		return logrus.InfoLevel, true
 	case kgo.LogLevelDebug:
 		return logrus.DebugLevel, true
+	case kgo.LogLevelNone:
+		return logrus.TraceLevel, false
```
twmb, line 65: No reason to explicitly add this case, since the return just below is the same thing?

twmb (approved):
Nice, thanks

plugin/klogrus/klogrus.go
```diff
@@ -7,17 +7,33 @@ import (
 	"github.com/twmb/franz-go/pkg/kgo"
 )
 
+var (
+	_ FieldLogger = (*logrus.Logger)(nil)
+	_ kgo.Logger  = (*Logger)(nil)
+)
+
+// FieldLogger interface.
```
twmb, line 15: Can you change this doc line to expand just a bit on what a FieldLogger is or why I'd want to use it / what it does?

plugin/klogrus/klogrus.go
```diff
@@ -30,6 +32,9 @@ func New(lr *logrus.Logger) *Logger {
 }
 
 // NewFieldLogger returns a new Logger using a FieldLogger interface.
+// it is isofunctional with New constructor, except it can accept either
+// *logrus.Logger or a possible wrapper thaf implements logrus.FieldLogger
```
twmb, line 36: `thaf` => `that`

plugin/klogrus/klogrus.go
```diff
@@ -13,6 +13,8 @@ var (
 )
 
 // FieldLogger interface.
```
twmb, line 15: needs the period to be dropped here

twmb (dismissed):
nice thanks, I'll release this _likely_ today

twmb:
I haven't forgotten, just got behind. Why changes today?

peczenyj:
I had some free time :)

twmb:
Apologies on the delay, lgtm, I'll tag imminently.

twmb:
Released as [plugin/klogrus/v1.1.0](https://github.com/twmb/franz-go/releases/tag/plugin%2Fklogrus%2Fv1.1.0). I think v1.10.0 was tagged incorrectly with an @ rather than a /, so at least v1.1.0 is correct.

PR #628, opened 2023-12-03
consuming: reset to nearest if we receive OOOR while fetching

### LINKED ISSUE #621: Feature request: KIP-842
In [KIP-842](https://cwiki.apache.org/confluence/display/KAFKA/KIP-842%3A+Add+richer+group+offset+reset+mechanisms) the author requests the ability to have the following new parameters:

`auto.offset.reset.on.no.initial.offset` - to specify how offset reset is handled when a consumer is starting up. This can be set to:
- `none`
- `earliest_on_start`
- `latest_on_start`

`auto.offset.reset.on.invalid.offset`. - to specify how a consumer should behave if an invalid offset is suddenly received. This can be set to:
- `none` 
- `nearest`

We see users needing this quite often where they would usually like to consume from the beginning of a topic, unless there has been an issue, in which case they'd either like to default to latest or to nearest (which would avoid a reset back to the beginning).

As far as I can see this wouldn't need to be implemented upstream to be incorporated into Franz-go.

twmb:
If we receive OOOR while fetching after a fetch was previously successful, something odd happened in the broker. Either what we were consuming was truncated underfoot, which is normal and expected (for very slow consumers), or data loss occurred without a leadership transfer. We will reset to the nearest offset after our prior consumed offset (by time!) because, well, that's what's most valid: we previously had a valid offset, and now it is invalid.

Closes #621.

### DIFF
--- a/pkg/kgo/source.go
+++ b/pkg/kgo/source.go
@@ -152,6 +152,11 @@ type cursorOffset struct {
 	// details.
 	lastConsumedEpoch int32
 
+	// If we receive OFFSET_OUT_OF_RANGE, and we previously *know* we
+	// consumed an offset, we reset to the nearest offset after our prior
+	// known valid consumed offset.
+	lastConsumedTime time.Time
+
 	// The current high watermark of the partition. Uninitialized (0) means
 	// we do not know the HWM, or there is no lag.
 	hwm int64
@@ -506,6 +511,7 @@ func (s *source) takeNBuffered(paused pausedTopics, n int) (Fetch, int, bool) {
 			pCursor.from.setOffset(cursorOffset{
 				offset:            lastReturnedRecord.Offset + 1,
 				lastConsumedEpoch: lastReturnedRecord.LeaderEpoch,
+				lastConsumedTime:  lastReturnedRecord.Timestamp,
 				hwm:               p.HighWatermark,
 			})
 		}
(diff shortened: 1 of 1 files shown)


nvartolomei:
@twmb may I ask why we prefer resetting by time? Say the consumer did consume offset=10, now it does a fetch offset and gets offset_out_of_range error with log start offset = 15, why it doesn't just move to offset 15 and instead has to do the timequery dance?

twmb:
@nvartolomei this dance is to address data loss at the end, not the start:
* Broker has 11, 12, 13, 14, 15
* Client consumes 11, 12
* Broker loses 13, 14, 15
* Client asks for 13
* Broker returns OOOR
* Second client produces 13, 14, 15

The best to reset to is the new 13, which is, by time, the first thing after 12. Any other offset-reset scheme can only pick the start or end; picking the start is a whole-topic-reconsume we want to avoid, picking the end misses the new 13, 14, 15.

nvartolomei:
But for brokers conforming to KIP-320 that shouldn’t be a problem. From
what I recall, broker is allowed to return only: a) not a leader, b) leader
epoch fence.

If so then, this timequery dance shouldn’t be applied for a broker with
KIP-320 support as OOOR can only indicate a prefix truncation from which we
can advance to the earliest available offset.

PR #629, opened 2023-12-04
optimisation: avoid ReadAll and use bytes buffer instead 

### LINKED ISSUE #623: lz4 decompression causes a lot of allocations due to ioutil.ReadAll
While using a consumer group where messages were lz4 compressed I noticed most of the allocations were due to [ioutil.ReadAll](https://github.com/twmb/franz-go/blob/master/pkg/kgo/compression.go#L281). Should we consider using `io.Copy` (fixed buffer length) instead of `ioutil.ReadAll`  (dynamically resized array)? I can provide benchmark results if needed

kalbhor:
This PR helps resolve #623

### DIFF
--- a/pkg/kgo/compression.go
+++ b/pkg/kgo/compression.go
@@ -5,7 +5,7 @@ import (
 	"compress/gzip"
 	"encoding/binary"
 	"errors"
-	"io/ioutil"
+	"io"
 	"runtime"
 	"sync"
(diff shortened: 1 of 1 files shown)


twmb:
lgtm, mind squashing these commits?

kalbhor:
Done :+1:

PR #636, opened 2023-12-13
pkg/sr: improve base URL and resource path joining

hhromic:
While working with the schema registry client in `pkg/sr`, we started to receive "not found" HTTP errors seemingly out of nowhere while registering schemas with `CreateSchema()`.

After closer examination, I found that the culrpit was that the base URL for the schema registry had a trailing slash in some of our configurations. For example `http://localhost:18081/` instead of `http://localhost:18081`.

After further examination, I traced back the issue to this line:
https://github.com/twmb/franz-go/blob/c09dc92d2db141d00c9eeb8ea229f72869222ca3/pkg/sr/client.go#L103

Where it is evident that the complete request URL results in the form `http://localhost:18081//path/to/resource` (incorrect).
I observed that the "not found" errors happen in both, Confluent Schema Registry and Redpanda.

To resolve the issue, in this PR the base URL is joined to the resource path using `url.JoinPath()` instead (from package `net/url`). This function takes care of cleaning up any ./ or ../ element in the base path. Also, Any sequences of multiple / characters will be reduced to a single / as well. Therefore, the resulting joined path is normalized properly.

~~Given that `url.JoinPath()` also escapes `?` characters (not meant to be in an actual path but as the separator for upcoming query parameters), I also refactored to support passing query parameters to the client's request functions explicitly and separately instead of embedded in the path argument. This also provides for formal encoding of query parameters thanks to the `Encode()` method of the `url.Values` type. The only instances that use query parameters are `DeleteSubject()` and `DeleteSchema()`.~~  EDIT: see comments below for a better suggested approach.

### DIFF
--- a/pkg/sr/api.go
+++ b/pkg/sr/api.go
@@ -404,7 +404,7 @@ func (cl *Client) DeleteSubject(ctx context.Context, subject string, how DeleteH
 	// DELETE /subjects/{subject}?permanent={x}
 	path := pathSubject(subject)
 	if how == HardDelete {
-		path += "?permanent=true"
+		ctx = WithParams(ctx, hardDelete)
 	}
 	var versions []int
 	defer func() { sort.Ints(versions) }()
@@ -419,7 +419,7 @@ func (cl *Client) DeleteSchema(ctx context.Context, subject string, version int,
 	// DELETE /subjects/{subject}/versions/{version}?permanent={x}
 	path := pathSubjectVersion(subject, version)
 	if how == HardDelete {
-		path += "?permanent=true"
+		ctx = WithParams(ctx, hardDelete)
 	}
 	return cl.delete(ctx, path, nil)
 }
--- a/pkg/sr/client.go
+++ b/pkg/sr/client.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"time"
 )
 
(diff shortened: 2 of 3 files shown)


twmb (dismissed):
While valid, I think this PR can be made simpler. There are other parameters, but they're abstracted into the Params struct. What do you think about adding an internal param for DeleteHow?

The reason DeleteHow was left out of a param is because I explicitly wanted people to choose hard or soft deleted, in code, at the callsite (rather than through params). But, we can bundle that back into context params with, I think, this diff:

```diff
diff --git a/pkg/sr/api.go b/pkg/sr/api.go
index 8a1fbde..8988f2e 100644
--- a/pkg/sr/api.go
+++ b/pkg/sr/api.go
@@ -404,7 +404,7 @@ func (cl *Client) DeleteSubject(ctx context.Context, subject string, how DeleteH
        // DELETE /subjects/{subject}?permanent={x}
        path := pathSubject(subject)
        if how == HardDelete {
-               path += "?permanent=true"
+               ctx = WithParams(ctx, hardDelete)
        }
        var versions []int
        defer func() { sort.Ints(versions) }()
@@ -419,7 +419,7 @@ func (cl *Client) DeleteSchema(ctx context.Context, subject string, version int,
        // DELETE /subjects/{subject}/versions/{version}?permanent={x}
        path := pathSubjectVersion(subject, version)
        if how == HardDelete {
-               path += "?permanent=true"
+               ctx = WithParams(ctx, hardDelete)
        }
        return cl.delete(ctx, path, nil)
 }
diff --git a/pkg/sr/params.go b/pkg/sr/params.go
index 33eb3c5..b2d643e 100644
--- a/pkg/sr/params.go
+++ b/pkg/sr/params.go
@@ -22,6 +22,7 @@ type Param struct {
        subject         string
        page            *int
        limit           int
+       hardDelete      bool
 }

 // WithParams adds query parameters to the given context. This is a merge
@@ -167,6 +168,9 @@ var (
        // DeletedOnly is a Param that configures whether to return only
        // deleted schemas or subjects in certain get operations.
        DeletedOnly = Param{deletedOnly: true}
+
+       // hardDelete is internal, and set when DeleteHow == HardDelete.
+       hardDelete = Param{hardDelete: true}
 )

 // Format returns a Param that configures how schema's are returned in certain
```

That should undo the bulk of changes in this PR. +1 to keeping the url.JoinPath though.

What do you think?

hhromic:
> There are other parameters, but they're abstracted into the Params struct. What do you think about adding an internal param for DeleteHow?

Ah, I totally missed that there was that Params struct nicely manipulated via request context 🙈.
I agree that moving HardDelete to this struct makes things clean and straight forward.
And more importantly, avoids the encoding of `JoinPath` when the parameter is passed _raw_ into the path.

I updated now the branch with your diff suggestion and completed the code for it to fully work in this way.
Thanks for the feedback and let me know what do you think of this new version.

twmb (approved):
Awesome ty

PR #647, opened 2023-12-19
bump all deps, except klauspost/compress

twmb:
bumping klauspost/compress requires go1.19. We'll do this bump with v1.16.


twmb:
Realistically this does not require any tags because users of this library can update transitive dependencies independently.

PR #653, opened 2023-12-27
kotel/tracer: add tombstone attribute

yianni:
Add tombstone to tracer span attributes as defined in the otel conventions.

https://github.com/open-telemetry/opentelemetry-go/blob/v1.21.0/semconv/v1.18.0/trace.go#L2930-L2938


twmb:
Thanks, released as plugin/kotel/v1.4.1

PR #672, opened 2024-02-01
pkg/kgo: allow marking using an offset

sudo-sturbia:
Currently when using `AutoCommitMarks` option the only function available for marking is `MarkCommitRecords(rs ...*Record)`, this requires you to keep `Record`s in memory until they're marked (possibly for prolonged periods of time) unnecessarily since the marking itself doesn't really need anything other than the offsets.

I added a new function `MarkCommitOffsets` to do the same as `MarkCommitRecords` but accept a map of `EpochOffset`s instead, let me know what you think, thank you

### DIFF
--- a/pkg/kgo/consumer_group.go
+++ b/pkg/kgo/consumer_group.go
@@ -2433,6 +2433,44 @@ func (cl *Client) MarkCommitRecords(rs ...*Record) {
 	}
 }
 
+// MarkCommitOffsets marks offsets to be available for autocommitting. This
+// function is only useful if you use the AutoCommitMarks config option, see
+// the documentation on that option for more details. This function does not
+// allow rewinds.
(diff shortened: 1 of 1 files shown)


twmb:
This looks good to me. Are you ok waiting a bit -- maybe a while? This is the only thing that necessitates a minor release right now and I'd like to bundle it with another feature or two, or maybe wait for the Kafka 3.7 release if that's coming out soon enough.

sudo-sturbia:
Yes, that's okay, it'd be great if it makes it some time soon but I'm not in a rush, thank you

twmb:
Timeline update: I plan to work on 3.7 changes this week or weekend and then merge this into the next minor release. If there's a delay it's because I'm a bit busy at work.

twmb:
Sorry for the delay, the last 3mo were much busier than I anticipated. Releasing imminently!

PR #677, opened 2024-02-16
Update `README.md` to add `Sharechat` as a user

tennistetris:
At [Sharechat](https://sharechat.com/about) we use `franz-go` quite heavily with both RedPanda and Kafka. It is the recommended client internally for golang.


twmb:
Awesome thank you very much!

PR #682, opened 2024-02-29
Update `README.md` to add `Hilton` as a user

dennisharrison:
At Hilton we use `franz-go` with `kafka` for eventing within our `go` micro-services.


twmb (approved):
Awesome, and good to know, thank you!

PR #683, opened 2024-03-02
Update skipKeys for internal API keys

weeco:
There used to be a typo (API Key 58 / Envelope instead of API Key 57 / UpdateFeatures). I'm not entirely sure what the "Envelope" API does, but I think it was meant to be `UpdateFeatures` instead.

Ref: https://kafka.apache.org/protocol.html

### DIFF
--- a/pkg/kversion/kversion.go
+++ b/pkg/kversion/kversion.go
@@ -209,6 +209,9 @@ func (opt guessOpt) apply(cfg *guessCfg) { opt.fn(cfg) }
 //	 6: UpdateMetadata
 //	 7: ControlledShutdown
 //	27: WriteTxnMarkers
+//	57: UpdateFeatures
+//	64: UnregisterBroker
+//	67: AllocateProducerIds
 func SkipKeys(keys ...int16) VersionGuessOpt {
 	return guessOpt{func(cfg *guessCfg) { cfg.skipKeys = keys }}
 }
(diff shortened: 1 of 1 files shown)


pkg/kversion/kversion.go
```diff
@@ -337,7 +340,7 @@ func (vs *Versions) versionGuess(opts ...VersionGuessOpt) guess {
 		//
 		// TODO: add introduced-version to differentiate some specific
 		// keys.
-		skipKeys: []int16{4, 5, 6, 7, 27, 58},
+		skipKeys: []int16{4, 5, 6, 7, 27, 57, 64, 67},
```
twmb, line 343: Why add UpdateFeatures? It's meant to be supported with/without Kraft. FWIW it's basically so far unused so I'm ok to remove it.

Envelope should still be filtered here, it's Kraft specific for the non-raft broker to forward a request to the raft broker when needed.

+1 for 64 and 67, also both Raft APIs, although this should also probably filter 59, 62, and 63

weeco: `UpdateFeatures` is only used by brokers to update the list of finalized features or do I missunderstand this API? I think it's not used by clients and therefore I'd consider it part of the "internal" list, but maybe you had different criterias in mind what APIs should go into this list?

I actually thought 58 was a typo for 58, but I wasn't sure what the "Envelope" API does either.

Not sure what API `59`, `62` and `63` are, I can't find it in the Apache Kafka protocol docs and they were not reported by Apache Kafka 3.7.0 either.

twmb:
I've done this more comprehensively in #739, closing this.

PR #687, opened 2024-03-06
sr SubjectVersions calls pathSubjectWithVersion

poulikov:
SubjectVersions calls pathSubjectWithVersion instead of pathSubject to build a path to retrieve all versions for a given subject.


twmb:
Ah whoops good fix

PR #697, opened 2024-03-22
pkg/kgo: use codecType explicitly

cristaloleg:
Scrolling through code I found mysterious constants, looks like this should be `codecType`.

However, `codec` param is `byte` but `codecType` is `uint8` I don't think it's a problem. Changing function param will increase diff for no big reason.

Thanks for the `franz-go` BTW ;)


twmb:
Sorry for the delay, the last few months were much busier than anticipated. Releasing imminently!

PR #703, opened 2024-04-03
chore: bump kotel dependencies

gary-lloyd:
This PR bumps all of the kotel dependencies to update to the latest version of otel.
Note that it also bumps the franz-go dependency which might not be the right thing to do depending on the release strategy for plugins.


twmb:
Howdy, I'm actually going to (someday hopefully sooner than later) bump all deps in all modules in this repo with the top-level `bump-deps.sh`. I'm going to close this for now given the _eventual_ goal.

PR #707, opened 2024-04-12
kfake: fix handling ListOffsets when Timestamp is requested

### LINKED ISSUE #706: ListOffsetsAfterMilli returns unexpected results when called against kfake cluster
While testing a chunk of our code, that uses `kadm.ListOffsetsAfterMilli()` against a `kfake` cluster, we bumped into an unexpected behaviour.

We expect that listing the topic's offsets, using a "know" timestamp (e.g. the timestamp of the last produced record) should return a known offset. But in our tests, `kfake` returns either 0, or the offset **after** the last record. But are confusing and look different from what we observe in real Kafka.

To illustrate the problem, I've extracted an example into this repo https://github.com/narqo/test-kfake.

narqo:
Closes #706 

This PR fixes the behaviour of how `kfake` handles `ListOffsets` API, when the request includes partition's timestamp. That is, the expected behaviour is to return the offset of the first batch _after_ the requested timestamp. As I pictured in #706, the current behaviour in the mainline is the opposite.

I didn't find a better way testing it, that to use the same example repo, I created to demo the problem. Refer to [this branch](https://github.com/narqo/test-kfake/tree/kfake-listoffsets-ts), where I replaced the `kfake` with my forked version.

### DIFF
--- a/pkg/kfake/02_list_offsets.go
+++ b/pkg/kfake/02_list_offsets.go
@@ -75,10 +75,11 @@ func (c *Cluster) handleListOffsets(b *broker, kreq kmsg.Request) (kmsg.Response
 					sp.Offset = pd.highWatermark
 				}
 			default:
+				// returns the index of the first batch _after_ the requested timestamp
 				idx, _ := sort.Find(len(pd.batches), func(idx int) int {
 					maxEarlier := pd.batches[idx].maxEarlierTimestamp
 					switch {
-					case maxEarlier < rp.Timestamp:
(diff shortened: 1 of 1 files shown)


pracucci (approved):
Nice job!

twmb (approved):
So basically, if the maxEarlier timestamp (search needle) is larger than the current index's timestamp, go left. I think this wasn't detected previously because I haven't ran searching by timestamp against kfake.

PR #712, opened 2024-04-26
fix: client.Lag doesn't calculate lag for all consumergroups with client.Lag(ctx)

0xgirish:
image

Getting lag as nil, in case Lag method was called as `client.Lag(ctx)`, other methods of this library defaults to listing offsets and describing to all the groups/topics incase no group/topic is specified, Assuming the same behaviour for this method. 

e.g.
[Client.DescribeGroups](https://pkg.go.dev/github.com/twmb/franz-go/pkg/kadm#Client.DescribeGroups)
> DescribeGroups describes either all groups specified, or all groups in the cluster if none are specified.

[Client.ListTopics](https://pkg.go.dev/github.com/twmb/franz-go/pkg/kadm#Client.ListTopics)
> ListTopics issues a metadata request and returns TopicDetails. Specific topics to describe can be passed as additional arguments. If no topics are specified, all topics are requested. Internal topics are not returned unless specifically requested. To see all topics including internal topics, use ListTopicsWithInternal.

### DIFF
--- a/pkg/kadm/groups.go
+++ b/pkg/kadm/groups.go
@@ -1460,7 +1460,12 @@ func (cl *Client) Lag(ctx context.Context, groups ...string) (DescribedGroupLags
 		}
 		if g.Err != nil {
 			delete(set, g.Group)
+			continue
 		}
+
+		// if no groups were specified for lag calculation, we will calculate lag for all groups
+		// adding groups to the set again.
+		set[g.Group] = struct{}{}
 	}
 	if len(set) == 0 {
 		return lags, nil
--- a/pkg/kadm/groups.go
+++ b/pkg/kadm/groups.go
@@ -1463,8 +1463,9 @@ func (cl *Client) Lag(ctx context.Context, groups ...string) (DescribedGroupLags
 			continue
 		}
 
-		// if no groups were specified for lag calculation, we will calculate lag for all groups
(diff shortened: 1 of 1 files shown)


0xgirish:
After changes:

image

0xgirish:
@twmb please review.

pkg/kadm/groups.go
```diff
@@ -1460,7 +1460,12 @@ func (cl *Client) Lag(ctx context.Context, groups ...string) (DescribedGroupLags
 		}
 		if g.Err != nil {
 			delete(set, g.Group)
+			continue
 		}
+
+		// if no groups were specified for lag calculation, we will calculate lag for all groups
+		// adding groups to the set again.
```
twmb, line 1467: Can you change this to,
```
// If the input set of groups is empty, DescribeGroups returns all groups.
// We add to `set` here so that the Lag function itself can calculate
// lag for all groups.
```

The doc comment here was a bit confusing to me, so this took me a bit to realize what you're trying to do here.

Overall, good change though.

0xgirish: Sorry for the delay, I have updated the doc comment.

0xgirish:
@twmb I have addressed the review comment.

twmb (approved):

PR #718, opened 2024-05-08
fix: Skip calculate lag when a topic has no offset commits

kylege:
If topic has no commit offset for a consumer group, then we should not calculate lag for it.

For example, topic A current high water offset is 100, but has no message because of delete mode

And there will not be any offset commits for consumer group because there is no messages to consume

Function CalculateGroupLag will calculate lag 100 for the consumer group, which is wrong.

### DIFF
--- a/pkg/kadm/groups.go
+++ b/pkg/kadm/groups.go
@@ -1585,6 +1585,9 @@ func CalculateGroupLag(
 			}
 
 			tcommit := commit[t.Topic]
+			if tcommit == nil {
+				continue
+			}
 			tend := endOffsets[t.Topic]
 			for _, p := range t.Partitions {
 				var (
@@ -1594,15 +1597,14 @@ func CalculateGroupLag(
 					ok      bool
 				)
 
-				if tcommit != nil {
-					if pcommit, ok = tcommit[p]; !ok {
-						pcommit = OffsetResponse{Offset: Offset{
-							Topic:     t.Topic,
-							Partition: p,
-							At:        -1,
-						}}
-					}
+				if pcommit, ok = tcommit[p]; !ok {
+					pcommit = OffsetResponse{Offset: Offset{
+						Topic:     t.Topic,
+						Partition: p,
+						At:        -1,
+					}}
 				}
+
 				if tend == nil {
 					perr = errListMissing
 				} else {


twmb:
The current code is deliberate. If a consumer in a group is assigned a topic, then if that topic/partition has no commit, it means the entire topic/partition is lagged.

The original implementation did _not_ consider topics with no commits as candidates to have lag. This was problematic because people were trying to calculate lag for a group that was bugged and never actually committed. The lag showed as zero, but technically by definition of how lag is calculated, the calculation _should_ have shown the entire topic lagging.

kylege:
> The current code is deliberate. If a consumer in a group is assigned a topic, then if that topic/partition has no commit, it means the entire topic/partition is lagged.
> 
> The original implementation did _not_ consider topics with no commits as candidates to have lag. This was problematic because people were trying to calculate lag for a group that was bugged and never actually committed. The lag showed as zero, but technically by definition of how lag is calculated, the calculation _should_ have shown the entire topic lagging.

We have too may topics that cleanup.policy=delete, so the topics will have zero messages after some time, when a consumer starts, it will never commit offset because no messages consumed, so the current code will calculate a very big lag number other than zero.

So how can i fix this scenario? Should i implement the calculation myself like kowl did?

twmb:
We've actually had multiple reports of Kowl's (now Console) lag page being inaccurate -- for basically the reason that I do it differently in the code above.

However, I think given what you're saying, a different fix is to subtract the log start offset from the lag. If a topic has log end offset 30 _and_ log start offset 30 and no commit, I _think_ today it'll show a lag of 30, whereas this can be changed to show a lag of 0.

twmb:
What span of time are you attempting to capture, and with what information?
I was thinking to add this to ProduceBatchMetrics, but the ProduceBatchWritten hook is only called on batches that are successfully produced.

Do you want to capture from the moment a record enters Produce through ... when? If you want to capture failures, do you mean for the batch duration to be called on every failure&retry, or only the final failure at which point the promise is called?

twmb:
(commented on wrong tab, deleted)

twmb:
Check out #744, I think that solves this. I'm going to release kadm soon.

PR #720, opened 2024-05-08
fix: clear lastAssigned when revoking eager consumer

### LINKED ISSUE #686: Consumer group rebalancing bug when switching from eager to cooperative consumers
Hello 👋!

We've been using `franz-go` for a while and have our consumers currently configured with the `RangeBalancer`. We want to switch to the `CooperativeStickyBalancer` and were trying to follow the instructions listed in both [KIP-429](https://cwiki.apache.org/confluence/display/KAFKA/KIP-429%3A+Kafka+Consumer+Incremental+Rebalance+Protocol#KIP429:KafkaConsumerIncrementalRebalanceProtocol-Consumer) and the `franz-go` docs for [CooperativeStickyBalancer](https://github.com/twmb/franz-go/blob/33923b606f7d632f3e821ea36f58b1f25fa86472/pkg/kgo/group_balancer.go#L818-L828) which state that we essentially need to perform a double bounce to upgrade. However we noticed that during the first bounce where we add `cooperative-sticky` to our set of balancers, when our old `range` consumers would leave the consumer group we kept getting consumption lag on some of the partitions.

There's a minimal reproduction in [this repo here](https://github.com/hamdanjaveed/kgo-cooperative-bug-reproducer).

In the reproduction we:

* Spin up kafka and a producer that publishes events
* Spin up a consumer that uses the `[RangeBalancer]` balancer
* Spin up a consumer that uses the `[CooperativeStickyBalancer, RangeBalancer]` balancers
* Shutdown the `[RangeBalancer]` consumer
* The remaining consumer seems to revoke half of its partitions but never re-assigns them to itself

We also see that the issue seems to persist if we then spin up a new consumer that uses the `[CooperativeStickyBalancer, RangeBalancer]` balancers, some partitions are seemingly never re-assigned and are stuck. We found that performing a full restart of all the consumers in the consumer group once we're in this state fixes the issue (i.e. unsticks the partitions).

Let me know if we're missing something here!

hamdanjaveed:
If a consumer that can be cooperative but is currently eager revokes its assigned partitions as part of a rebalance it should also clear its tracking state between rebalances as it has revoked all of its partitions.

This was leading to some inconsistent behaviour when migrating from eager consumers to cooperative consumers following the migration instructions in [KIP-429](https://cwiki.apache.org/confluence/display/KAFKA/KIP-429%3A+Kafka+Consumer+Incremental+Rebalance+Protocol#KIP429:KafkaConsumerIncrementalRebalanceProtocol-Consumer).

Fixes #686

### DIFF
--- a/pkg/kgo/consumer_group.go
+++ b/pkg/kgo/consumer_group.go
@@ -639,6 +639,7 @@ func (g *groupConsumer) revoke(stage revokeStage, lost map[string][]int32, leavi
 			g.cfg.onRevoked(g.cl.ctx, g.cl, g.nowAssigned.read())
 		}
 		g.nowAssigned.store(nil)
+		g.lastAssigned = nil
 
 		// After nilling uncommitted here, nothing should recreate
 		// uncommitted until a future fetch after the group is


hamdanjaveed:
Tried writing a test for this but wasn't able to write one that properly exercised the code path here :(

Instead we've manually tested this fix by running consumers locally from this minimal repro: https://github.com/hamdanjaveed/kgo-cooperative-bug-reproducer

twmb:
Sorry for the delay, the last few months were much busier than anticipated. Releasing imminently!
(this fixes a _somewhat_ minor bug so I did not rush a patch release for this)

PR #745, opened 2024-05-27
SR: prevent panic when no decode function is registered

lovromazgon:
A small fix for a possible panic when a type is registered without a decode function. The test kinda hides this possibility, because `Serde` has a default `DecodeFn` option. This case is more likely to hit when `Serde` is instantiated without that option.

I removed the decode test case for schema ID 100, as it was the same as the one for ID 99, instead I added one with ID 101 which has a registered type, but no decode function.


twmb (approved):

twmb:
Thanks!

PR #755, opened 2024-06-05
Fix typo (update default Acks in comment) in config.go

Maks-D:
Type `Acks` description contains a mistake in the default value. According to the default config, the default value of acks is AllISRAcks (code https://github.com/twmb/franz-go/blob/7641d9e673e31217cc353ee8cfc473ba0946891d/pkg/kgo/config.go#L495)

### DIFF
--- a/pkg/kgo/config.go
+++ b/pkg/kgo/config.go
@@ -869,7 +869,7 @@ func DefaultProduceTopic(t string) ProducerOpt {
 // This controls the durability of written records and corresponds to "acks" in
 // Kafka's Producer Configuration documentation.
 //
-// The default is LeaderAck.
+// The default is AllISRAcks.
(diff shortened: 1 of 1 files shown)


twmb:
Curiously, this same question came up 2d ago internally. Copying my response:

---

The _type_ has a default value of 0. It’s a struct wrapping an int. 0 means leader ack.

The _option_ defaults internally to -1

This is an example of _why_ the client is option driven, rather than a Config struct — with a Config struct, anything you don’t specify has a 0 default, which would be leader ack. With options, I can internally define a struct with -1, and than you can override if you want.

---

So, the existing documentation is correct, but if you want to change this to mention that the _option_ defaults to whatever, I may be open to some wording change about that.

PR #767, opened 2024-06-13
kfake: Drop topic configuration validation

jan-g:
This isn't used at the moment; it gets in the way of using kfake to stub out behaviour that mocks custom cluster & topic config.


twmb:
+1. I might reintroduce some aspects of this later -- e.g. if I ever introduce handling compaction, it'll be worth validating the cleanup.policy. But, it's not worth failing generally for random configs.

PR #772, opened 2024-07-03
Update `README.md` to add `EasyPost` as a user

kev-chien:
We've been using `franz-go` at [EasyPost](https://github.com/EasyPost)


twmb:
Nice thank you!

PR #774, opened 2024-07-10
refactor: replace sliceWriter with bytes.Buffer and use it in decompression 

### LINKED ISSUE #623: lz4 decompression causes a lot of allocations due to ioutil.ReadAll

kalbhor:
This MR resolves #623 by using a byte buffer in decompression instead of always creating a new one. It also replaces the custom `sliceWriter` with the standard `bytes.Buffer` everywhere.

### DIFF
--- a/pkg/kgo/compression.go
+++ b/pkg/kgo/compression.go
@@ -14,15 +14,7 @@ import (
 	"github.com/pierrec/lz4/v4"
 )
 
-// sliceWriter a reusable slice as an io.Writer
-type sliceWriter struct{ inner []byte }
-
-func (s *sliceWriter) Write(p []byte) (int, error) {
-	s.inner = append(s.inner, p...)
-	return len(p), nil
-}
-
-var sliceWriters = sync.Pool{New: func() any { r := make([]byte, 8<<10); return &sliceWriter{inner: r} }}
+var byteBuffers = sync.Pool{New: func() any { return bytes.NewBuffer(make([]byte, 8<<10)) }}
 
 type codecType int8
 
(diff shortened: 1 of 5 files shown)


pkg/kgo/compression.go
```diff
@@ -202,7 +194,9 @@ func (c *compressor) compress(dst *sliceWriter, src []byte, produceRequestVersio
 		}
 
 	case codecSnappy:
-		dst.inner = s2.EncodeSnappy(dst.inner[:cap(dst.inner)], src)
+		out := s2.EncodeSnappy(dst.Bytes(), src)
+		dst.Reset()
```
twmb, line 198: This chunk of code _might_ actually be a reason to keep sliceWriters as well.

* EncodeSnappy (and below in zstd, EncodeAll) both take an input slice, and _if there is capacity_, uses it, _if not_, returns a new larger slice
* Now, the code doubly writes and potentially doubly allocates. If Encode returns a larger slices, dst.Write(out) actually also reallocates the internal slice. Even if it does not reallocate, the Write is an unnecessary copy.

One way to hack around this in the current logic is to change the function signature and to check:

```go
 func (c *compressor) compress(dst *bytes.Buffer, 

// ... other code

if len(out) == 0 {
  // skip
}
dst.Reset()
dstBytes := dst.Bytes()
dstBytes = dstBytes[:cap(dstBytes)]
if len(dstBytes) == 0 || &dstBytes[0] != &out[0] {
  // replace the bytes.Buffer itself, discard old bytes.Buffer
  *dstp = bytes.NewBuffer(dstBytes)
  dst = *dstp
}
```

^^Essentially, keep the old bytes.Buffer if the slice didn't change, otherwise replace the bytes.Buffer with a new one if the slice did change. It's really convoluted though, and this may be a reason to keep/use sliceWriters. wdyt?

kalbhor: Maybe we can utilize the methods in zstd & snappy that provide the length of expected decompressed bytes and grow the byte buffer accordingly beforehand? These methods are exported by the libs and these algos probably want us to leverage this info. This way we can be sure the code won't doubly allocate and neither copy again

image

twmb (approved):

pkg/kgo/compression.go
twmb: Great idea w/ MaxEncodedLen and MaxEncodedSize, I didn't know those methods existed on these compressors

PR #775, opened 2024-07-11
Add ShardErrors.Unwrap() support

pracucci:
I have an use case where I need to check whether the request context was canceled / timed out. It would be helpful to me if we could make `ShardErrors` supporting `Unwrap()`, so that `errors.Is()` works on it.


twmb:
+1, I'm aiming to release a franz-go patch release and kadm by / during this weekend.

twmb:
I thought I got this in the last release, my mistake on the delay here. This will be in 1.18 this evening.

Issue #769, opened 2024-06-14
RequestTimeoutOverhead appears to override Context cancellation


JFlath:
In the below example we see a produce request fail (due to a broker restart) and repeated attempts to connect to that broker with the default `RequestTimeoutOverhead` of 10s:

```2024-06-04T12:55:45.348+0100    INFO    producer_client kzap/kzap.go:114        produce request failed, triggering metadata update      {"broker": "1", "err": "unable to dial: dial tcp 10.XX.XX.XX:9092: connect: connection refused"}
2024-06-04T12:55:45.348+0100    DEBUG   producer_client kzap/kzap.go:110        retry batches processed {"wanted_metadata_update": true, "triggering_metadata_update": true, "should_backoff": false}
2024-06-04T12:55:45.348+0100    INFO    producer_client kzap/kzap.go:114        metadata update triggered       {"why": "failed produce request triggered metadata update"}
2024-06-04T12:55:46.114+0100    DEBUG   producer_client kzap/kzap.go:110        opening connection to broker    {"addr": "redpanda-1.redpanda.my_cluster.svc.cluster.local.:9092", "broker": "1"}
2024-06-04T12:55:46.115+0100    WARN    producer_client kzap/kzap.go:116        unable to open connection to broker     {"addr": "redpanda-1.redpanda.my_cluster.svc.cluster.local.:9092", "broker": "1", "err": "dial tcp 10.XX.XX.XX:9092: connect: connection refused"}

2024-06-04T12:55:46.343+0100    DEBUG   producer_client kzap/kzap.go:110        opening connection to broker    {"addr": "redpanda-1.redpanda.my_cluster.svc.cluster.local.:9092", "broker": "1"}
2024-06-04T12:55:56.352+0100    WARN    producer_client kzap/kzap.go:116        unable to open connection to broker     {"addr": "redpanda-1.redpanda.my_cluster.svc.cluster.local.:9092", "broker": "1", "err": "dial tcp 10.XX.XX.XX:9092: i/o timeout"}

2024-06-04T12:55:56.352+0100    DEBUG   producer_client kzap/kzap.go:110        opening connection to broker    {"addr": "redpanda-1.redpanda.my_cluster.svc.cluster.local.:9092", "broker": "1"}
2024-06-04T12:56:05.717+0100    DEBUG   producer_client kzap/kzap.go:110        reaped connections      {"time_since_last_reap": "20.009282863s", "reap_dur": "127.7µs", "num_reaped": 4}
[...]

PR #781, opened 2024-07-17
fix setting lz4 compression levels

### LINKED ISSUE #778: Unable to specify LZ4 compression level
CompressionCoded.WithLevel casts its argument to byte and maxes out at 127. - https://github.com/twmb/franz-go/blob/a5f2b710830e32fc3f90374be47ec59849807342/pkg/kgo/compression.go#L71

However, LZ4 compression levels are much larger than that - https://github.com/pierrec/lz4/blob/v4/options.go#L155 - they are defined as 0, 512, 1024, ... , 131072.

Specifying LZ4 compression level in franz-go therefore will always result in the default.

asg0451:
resolves #778

### DIFF
--- a/pkg/kgo/compression.go
+++ b/pkg/kgo/compression.go
@@ -41,7 +41,7 @@ const (
 // for that batch.
 type CompressionCodec struct {
 	codec codecType
-	level int8
+	level int
 }
 
 // NoCompression is a compression option that avoids compression. This can
@@ -68,10 +68,7 @@ func ZstdCompression() CompressionCodec { return CompressionCodec{codecZstd, 0} 
 //
 // If the level is invalid, compressors just use a default level.
 func (c CompressionCodec) WithLevel(level int) CompressionCodec {
-	if level > 127 {
-		level = 127 // lz4 could theoretically be large, I guess
-	}
-	c.level = int8(level)
+	c.level = level
 	return c
 }
 
--- a/pkg/kgo/compression.go
+++ b/pkg/kgo/compression.go
@@ -113,8 +113,8 @@ out:
 		case codecGzip:
 			level := gzip.DefaultCompression
 			if codec.level != 0 {
-				if _, err := gzip.NewWriterLevel(nil, int(codec.level)); err != nil {
-					level = int(codec.level)
+				if _, err := gzip.NewWriterLevel(nil, codec.level); err != nil {
+					level = codec.level
 				}
 			}
 			c.gzPool = sync.Pool{New: func() any { c, _ := gzip.NewWriterLevel(nil, level); return c }}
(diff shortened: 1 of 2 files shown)


twmb:
Thanks!

twmb (dismissed):

asg0451:
@twmb am i good to merge here?

twmb:
I want to get this in the next patch release but can't figure out why #785 is failing (which is important to get in, probably the most important bug at the moment)

twmb:
Typo, I mean #786. If you're able to figure out the CI problem, that'd be great :D

asg0451:
ok, i'll hold off. that pr looks like ci is passing to me though?

twmb:
Wow, typo twice. I'm going to delete my prior comments. #787 is the one that's failing.

asg0451:
gotcha. not sure if i really have the bandwidth to debug that at the moment ;)

btw if i may ask, do these new race-y bugs affect the integration i just merged into cockroachdb? will we need to upgrade to fix this (once you fix it of course) before it makes it into a new release? we produce concurrently with max buffered records set, not bytes.

twmb:
The data race is largely non impactful (it's actually existed for a really long time, it's really hard to trigger, and at the architecture level, you're not going to see a partial write).

The #787 is worrying, but it only affects people that use MaxBufferedBytes. I don't think you do, it's a fairly new option and I don't expect most people to use it.

asg0451:
(if you want to move this discussion to another venue that's fine, lmk)

> at the architecture level, you're not going to see a partial write
can you expand on this? what would the result be both downstream and in the running process

twmb:
The data race is a read/write problem on a boolean. The standard problem with read/write races is that the reader experiences a partial read -- e.g. I'm writing "foobar" and the underlying data says "hacked", the reader reads "fooked" (half of foobar, half of hacked). No CPU architecture actually does a partial write of a boolean sized field.

The other problem is that compilers and CPUs can reorder where the read happens, but Go uses seqcst for any atomic operation, and there are atomic operations before and after the reading of this bool. So, this bool isn't being reordered to a place that it really matters.

The last worry here is that _what_ this field is for isn't working properly. This field is meant to guard failing records if we don't know it's safe. There's a related open issue about this (not the bug, but "why was this hanging") where you can see what safe is, here: https://github.com/twmb/franz-go/issues/769#issuecomment-2235707417

So, if a race happens and what the bool is guarding isn't effectively guarded, what you'll see in your application is OutOfOrderSequenceNumber errors and you'll need to restart your client. This chunk of code has existed for 3yr and so far nobody has reported unexpected OOOSN errors.

asg0451:
gotcha. thanks for the context and the responsiveness!

PR #782, opened 2024-07-18
Add Conduit to list of users

hariso:
@twmb A huge thanks for this project!

We use it in Conduit, but also a few related but separate projects (our Kafka connector and our own schema registry). 

By the way, what do you think about removing Benthos, given that it's now Redpanda Connect?


twmb:
I'd rather replace Benthos with Redpanda Connect, since it's a pretty "different" thing, though maybe I'm just favoring keeping it because I like more names :D

twmb:
(also ty for the addition here!)

twmb (approved):

Issue #777, opened 2024-07-16
Deadlock in Produce() / TryProduce() when kgo.MaxBufferedBytes() is configured


pracucci:
In Mimir we found a deadlock in `Produce()` and `TryProduce()` when `kgo.MaxBufferedBytes()` is configured and the limit hit.

The gist of the root cause of such deadlock is that there's no guarantee that all goroutines waiting for `p.waitBuffer` will be released, because there are conditions under which the number of messages sent to `p.waitBuffer` channel is different than the number of goroutines waiting for it.

## Details

`Produce()` and `TryProduce()` wait for `p.waitBuffer` when the buffer limit is reached:
https://github.com/twmb/franz-go/blob/a5f2b710830e32fc3f90374be47ec59849807342/pkg/kgo/producer.go#L439-L449

A message is published to `p.waitBuffer` each time a record completes, if the client detect that such record was over the bytes limit:
https://github.com/twmb/franz-go/blob/a5f2b710830e32fc3f90374be47ec59849807342/pkg/kgo/producer.go#L520-L534

In case of "max buffered records" (which does not suffer this issue) each record has the same "weight": each record accounts for +1 in the `p.bufferedRecords` accumulator. However, in the case of "max buffered bytes" limit, different records don't have the same "weight", because each record has a different `userSize()`. This means that the computation of `wasOverMaxBytes` is screwed up, depending on which record gets completed first.

### Example

Let's assume we set `kgo.MaxBufferedBytes()` to 100 bytes and the following workflow happen in order:

1. `Produce()` record A (100 bytes)
2. `Produce()` record B (50 bytes), waiting for buffer to free
3. `Produce()` record C (50 bytes), waiting for buffer to free
4. Record A is produced, `finishRecordPromise()` gets called, detects it was over the limit so publish 1 message to `waitBuffer`
5. Record B is unlocked, `finishRecordPromise()` gets called, does **not** detect it was over the limit (only 50 bytes), so record C is never unblocked and will wait indefinitely on `waitBuffer`

### Reproduction test

I've written a test reproducing the deadlock:
https://github.com/twmb/franz-go/compare/pracucci:reproduce-produce-deadlock

twmb:
Good find. Offhand, I think the fix is to convert this to a sync.Cond and have every waiter be notified and inspect the conditions, but I'll have to stare a bit at how to do this.

twmb:
[...]

PR #787, opened 2024-07-22
kgo: fix deadlock in Produce when using MaxBufferedBytes 

twmb:
Copying from the issue,
"""
1) Produce() record A (100 bytes)
2) Produce() record B (50 bytes), waiting for buffer to free
3) Produce() record C (50 bytes), waiting for buffer to free
4) Record A is produced, finishRecordPromise() gets called, detects it was over the limit so publish 1 message to waitBuffer
5) Record B is unlocked, finishRecordPromise() gets called, does not detect it was over the limit (only 50 bytes), so record C is never unblocked and will wait indefinitely on waitBuffer
"""

The fix requires adding a lock while producing. This reuses the existing
lock on the `producer` type. This can lead to a few more spurious
wakeups in other functions that use this same mutex, but that's fine.

The prior algorithm counted anything to produce immediately into the
buffered records and bytes fields; the fix for https://github.com/twmb/franz-go/issues/777 could not really be
possible unless we avoid counting the "buffered" aspect right away.
Specifically, we need to have a goroutine looping with a sync.Cond that
checks *IF* we add the record, will we still be blocked? This allows us
to wake up all blocked goroutines always (unlike one at a time, the
problem this issue points out), and each goroutine can check under a
lock if they still do not fit.

This also fixes an unreported bug where, if a record WOULD be blocked
but fails early due to no topic / not in a transaction while in a
transactional client, the serial promise finishing goroutine would
deadlock.

Closes https://github.com/twmb/franz-go/issues/777.

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -31,7 +31,6 @@ jobs:
 
   integration-test-kafka:
     if: github.repository == 'twmb/franz-go'
-    needs: golangci
     runs-on: ubuntu-latest
     name: "integration test kafka"
     container: golang:latest
@@ -55,7 +54,7 @@ jobs:
           KAFKA_KRAFT_CLUSTER_ID: XkpGZQ27R3eTl3OdTm2LYA # 16 byte base64-encoded UUID
     steps:
       - uses: actions/checkout@v4
-      - run: go test ./...
+      - run: go test -timeout 5m ./...
         env:
           KGO_TEST_RF: 1
           KGO_SEEDS: kafka:9092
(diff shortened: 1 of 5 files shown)


twmb:
I've loop tested this locally,
* ~15 times against a 3 node Kafka cluster
* ~4 times against a 3 node Kafka cluster with `-race`
* ~15 times against a 1 node Bitnami cluster (to mirror CI)
* ~5 times so far against a 1 node Bitnami cluster with `-race`

So far I am unable to recreate what fails in CI at all.

asg0451:
gotcha. not sure if i really have the bandwidth to debug that at the moment ;)

btw if i may ask, do these new race-y bugs affect the integration i just merged into cockroachdb? will we need to upgrade to fix this (once you fix it of course) before it makes it into a new release? we produce concurrently with max buffered records set, not bytes.

twmb:
Turns out the code was correct. The test needed to use `newTestClient` instead of `NewClient`.

I did find a small "optimization" at least, in all this looking.

Issue #784, opened 2024-07-19
Question: Is it safe to poll after receiving ErrGroupSession on poll when a broker node restarts in the cluster?


sbuliarca:
Hi!

We have the following situation, and I would really appreaciate your guidance and knowledge insides on it:
- we're consuming under a consumer group using the PollMaxRecords 
- when a broker restarts, some of the apps that were consuming from that broker, receive in the kgo.Fetches the error: `unable to join group session: unable to dial: dial tcp broker_ip:broker_port: connect: connection refused` which I identified being added [here in the code](https://github.com/twmb/franz-go/blob/master/pkg/kgo/consumer_group.go#L433)

The question is:  is it safe to still use the client and retry the Poll method in this case? Will the client retry to join the group session safely, since it seems it also triggered a metadata refresh from the code? Or do we need to close and recreate the client at this point?

Thank you so much

twmb:
This is safe to recover from, although you may experience duplicate messages -- your group consumer died (potentially after processing some messages, and potentially before committing). The client will eventually rejoin the group.

PR #789, opened 2024-07-24
kgo: export the wrapped error from ErrGroupSession

sbuliarca:
We want to test in a unit test the behaviour exposed in this question: https://github.com/twmb/franz-go/issues/784, where we rely on this error type. So we need to create this kind of error, with another wrapped one inside it in the tests.

Would this change be ok? The only side efect of it is that clients of the lib would be able to create it too.

### DIFF
--- a/pkg/kgo/errors.go
+++ b/pkg/kgo/errors.go
@@ -298,11 +298,11 @@ func (e *errUnknownCoordinator) Error() string {
 // consumer group member was kicked from the group or was never able to join
 // the group.
 type ErrGroupSession struct {
-	err error
+	Err error
 }
 
 func (e *ErrGroupSession) Error() string {
-	return fmt.Sprintf("unable to join group session: %v", e.err)
+	return fmt.Sprintf("unable to join group session: %v", e.Err)
 }
 
-func (e *ErrGroupSession) Unwrap() error { return e.err }
+func (e *ErrGroupSession) Unwrap() error { return e.Err }


twmb:
Why not use `Unwrap` to get the inner error, rather than exposing the error?

sbuliarca:
> Why not use `Unwrap` to get the inner error, rather than exposing the error?

Our problem is that we want to create this error ourselves, in the test, to simulate a kgo.Fetches that contains this error. Something like:
```
egs := kgo.ErrGroupSession{Err: syscall.ECONNREFUSED}
fetch := kgo.Fetch{
			Topics: []kgo.FetchTopic{{
				Topic: "dummyTopic",
				Partitions: []kgo.FetchPartition{{
					Partition: 0,
					Err:       err,
				}},
			}}}
```
We currently can not do it, as the err is not exported in kgo.ErrGroupSession.

sbuliarca:
Hey @twmb! Just a friendly reminder of the existence of this PR. Thanks!

twmb:
I haven't forgotten -- I'm primarily waiting until this weekend or next (or next) to set aside time to implement the three new KIPs that were released in Kafka 3.8.

That said -- I'm still not sure what value a test on this particular error provides? Stakes here are pretty low though, so I'm not really _opposed_ either.

sbuliarca:
>I'm still not sure what value a test on this particular error provides?

For us, it allows to unit test the exact scenario described in the question https://github.com/twmb/franz-go/issues/784 , that we're not exiting the poll loop when we get such an error.

twmb:
Bit of a delay on my side... releasing this evening.

PR #794, opened 2024-07-29
kgo: add TopicID to the FetchTopic type

### LINKED ISSUE #790: fetch using topic id
Hello,

I'm trying to use the franz go kafka client to consume records from a source cluster. I'm using a direct consumer, because the workload is to consume a list of arbitrary (topic id, partition id, offset) periodically.

I'm trying to achieve this by keeping a single client alive. When it receives the list of (topic id, partition id, offset) to consume from, it performs remove consume topic partitions using `RemoveConsumePartitions` or even `PurgeTopicsFromClient`, and adds the new consume topic partitions using `AddConsumePartitions`.

However, the above APIs use the topic name string and not the topic ids. I think I can make my use case of ensuring I'm fetching from the correct topic work if the `FetchTopic` struct also has the topic id in it and it's returned from the `PollRecords` function with the topic ids also set.

I think it's possible to do this by:
1. Adding topic id field to `FetchTopic`.
2. Adding a config to indicate that the topic id should be set in `FetchTopic`.
3. Setting the topic id in the FetchTopic upon fetches(I think this is possible for request versions >= 13?)

What do you think? If you're open to this, then I can implement this.

twmb:
Closes #790.

### DIFF
--- a/pkg/kgo/record_and_fetch.go
+++ b/pkg/kgo/record_and_fetch.go
@@ -274,6 +274,9 @@ func (p *FetchPartition) EachRecord(fn func(*Record)) {
 type FetchTopic struct {
 	// Topic is the topic this is for.
 	Topic string
+	// TopicID is the ID of the topic, if your cluster supports returning
+	// topic IDs in fetch responses (Kafka 3.1+).
+	TopicID [16]byte
 	// Partitions contains individual partitions in the topic that were
 	// fetched.
 	Partitions []FetchPartition
@@ -560,6 +563,7 @@ func (fs Fetches) EachTopic(fn func(FetchTopic)) {
 	for topic, partitions := range topics {
 		fn(FetchTopic{
 			topic,
+			[16]byte{},
 			partitions,
 		})
 	}
--- a/pkg/kgo/source.go
+++ b/pkg/kgo/source.go
@@ -1041,6 +1041,7 @@ func (s *source) handleReqResp(br *broker, req *fetchRequest, resp *kmsg.FetchRe
 
 		fetchTopic := FetchTopic{
 			Topic:      topic,
+			TopicID:    rt.TopicID,
 			Partitions: make([]FetchPartition, 0, len(rt.Partitions)),
 		}
 


pkg/kgo/source.go
```diff
@@ -1041,6 +1041,7 @@ func (s *source) handleReqResp(br *broker, req *fetchRequest, resp *kmsg.FetchRe
 
 		fetchTopic := FetchTopic{
 			Topic:      topic,
+			TopicID:    rt.TopicID,
```
arjunnair1997, line 1044: Nice, so if version >= 13, I should be assured that this is set?

twmb: It's set if the broker returns it -- it's a new field in fetch response v13, yes -- and this will be all 0s if it is unset.

arjunnair1997: Gotcha, I assume all 0s is an invalid topic id.

arjunnair1997 (approved):

pkg/kgo/record_and_fetch.go
```diff
@@ -274,6 +274,9 @@ func (p *FetchPartition) EachRecord(fn func(*Record)) {
 type FetchTopic struct {
 	// Topic is the topic this is for.
 	Topic string
+	// TopicID is the ID of the topic, if your cluster supports returning
+	// topic IDs in fetch responses (Kafka 3.1+).
```
arjunnair1997, line 278: nit: maybe add fetch request version? Will make it easier for someone using the API, I think.

Imran-imtiaz48 (commented):
the addition of the TopicID field enhances the FetchTopic struct, aligning it with newer Kafka versions. However, the initialization of this field with an empty value in EachTopic may need reconsideration based on its actual usage and availability. Ensuring proper initialization and documentation will improve the code’s robustness and clarity.

twmb:
Topics do not have all 0 topic IDs, if the cluster supports topic IDs. The user can be sure that a topic ID is present based on the fact it is not all 0s. I use this same comparison myself:
https://github.com/twmb/franz-go/blob/6b61d17383b1e0c766143118efa39f1f320a3b2b/pkg/kgo/metadata.go#L795-L803
https://github.com/twmb/franz-go/blob/6b61d17383b1e0c766143118efa39f1f320a3b2b/pkg/kgo/source.go#L1895-L1898

pkg/kgo/record_and_fetch.go
twmb: End users generally don't actually know what version of Kafka introduced what version of a request. It's undocumented anywhere -- a person has to first know the version a field was introduced in, then figure out which KIP introduced that version, then figure out which version of Kafka the KIP was actually released in.

Mentioning the version of Kafka that introduced topic IDs in the fetch response circumvents that process and gets directly to the answer, IMO.

twmb:
Bit of a delay on my side ... releasing this evening.

PR #802, opened 2024-08-08
Clarify bytes metrics description

pracucci:
Earlier today I was wondering where exactly `write_bytes_total` and `read_bytes_total` are tracked, and I realized it counts the bytes we write and read from the wire.

What do you think if we clarify it in the metric description, including a specific mention with regards to compression?


twmb:
Sure thang, ty

Issue #815, opened 2024-08-28
kprom: Support for multiple clients


lukasrynt:
Hi,

I'm trying to use the kprom plugin with two producers running in the same program to expose separate metrics for each producer. I expected the `WithClientLabel` configuration option to allow this, but when I register both producers via the hooks option to the same metric, only the most recent producer gets registered. Here's an example:

```go
metrics := kprom.NewMetrics("test", kprom.WithClientLabel())

// Prepare producers
primary, err := kgo.NewClient(
		kgo.SeedBrokers(brokers...),
		kgo.ClientID("primary"),
		kgo.WithHooks(metrics),
	)
...
secondary, err := kgo.NewClient(
		kgo.SeedBrokers(brokers...),
		kgo.ClientID("secondary"),
		kgo.WithHooks(metrics),
	)
...
mux := nethttp.NewServeMux()
mux.Handle("/metrics", metrics.Handler())

// Run the actual HTTP server with metrics
...
// Only metrics about secondary client are seen (although I suspect it's sum of both)
```

However, I believe that supporting separate metrics for multiple consumers/producers within a single program is a valid use case for this plugin, especially since `kgo` allows multiple client instances in the same application. I would love to see this use case supported, I can even create a pull request for it if you'll find this issue valid.

twmb:
I'm open to this, what's the proposed API?

lukasrynt:
Thanks for the swift response. I created a [PR](https://github.com/twmb/franz-go/pull/820) for this. I would retain the API, feel free to correct me if you don't like it.

mehranmeidani:
@lukasrynt Why do you want to pass the same `kprom.NewMetrics` instance to both client? Why not create a new one and register both to a single prometheus registry and serve metrics from there? We are doing this in our application with multiple clients and it's working very well.

lukasrynt:
@mehranmeidani Thanks for the suggestion. I already tried passing the metrics to one common Prometheus registry, however the solution felt a bit hacky. Since kprom already creates its own registry, one would have to create an extra structure that implements the `Collector` interface, that another `prometheus.Registry` can register.

I might be missing something though. Would you be so kind as to provide some example code with usage of two kprom metrics instances? Thanks a lot!

lukasrynt:
[...]

PR #820, opened 2024-09-17
kprom: Multiple clients

lukasrynt:
# Issue
The main goal of this PR is to add the ability for `kprom` plugin to differentiate between multiple clients. 

The current implementation of the `WithClientLabel()`  option in the kprom package assigns metrics to the most recently registered kgo client, which leads to metrics being grouped under a static client_id. The goal is to keep the existing API but make the `client_id` label dynamic, allowing it to reflect the actual client making the request rather than being statically set at first registration (see [twmb#815](https://github.com/twmb/franz-go/issues/815)).

Only a subset of metrics will be affected by this change:
- `write_bytes_total`
- `write_errors_total`
- `write_wait_seconds`
- `write_time_seconds`
- `read_bytes_total`
- `read_errors_total`
- `read_time_seconds`
- `request_duration_e2e_seconds`
- `request_throttled_seconds`
- `produce_compressed_bytes_total`
- `produce_(uncompressed_)bytes_total`
- `produce_batches_total`
- `produce_records_total`
- `fetch_compressed_bytes_total`
- `fetch_(uncompressed_)bytes_total`
- `fetch_batches_total`
- `fetch_records_total`

Metrics that will continue to have a constant `client_id`:
- `connects_total`
- `connect_errors_total`
- `disconnects_total`
- `buffered_fetch_records_total`
- `buffered_produce_records_total`

# Testing
I tested the updated metrics locally using a combination of consumers and producers. When scraping Prometheus, the metrics are now properly separated by the client_id label, even when writing to and reading from the same topic using different client IDs for two producers and consumers:

```
# HELP test_buffered_fetch_records_total Total number of records buffered within the client ready to be consumed
# TYPE test_buffered_fetch_records_total gauge
test_buffered_fetch_records_total{client_id="primary"} 0
# HELP test_buffered_produce_records_total Total number of records buffered within the client ready to be produced
# TYPE test_buffered_produce_records_total gauge
test_buffered_produce_records_total{client_id="primary"} 0
# HELP test_connects_total Total number of connections opened
# TYPE test_connects_total counter
test_connects_total{client_id="primary",node_id="1"} 8
test_connects_total{client_id="primary",node_id="seed_0"} 4
# HELP test_fetch_bytes_total Total number of uncompressed bytes fetched
# TYPE test_fetch_bytes_total counter
test_fetch_bytes_total{client_id="primary",node_id="1",topic="primary"} 800
test_fetch_bytes_total{client_id="secondary",node_id="1",topic="primary"} 800
# HELP test_produce_bytes_total Total number of uncompressed bytes produced
# TYPE test_produce_bytes_total counter
test_produce_bytes_total{client_id="primary",node_id="1",topic="primary"} 400
test_produce_bytes_total{client_id="secondary",node_id="1",topic="primary"} 400
# HELP test_read_bytes_total Total number of bytes read
# TYPE test_read_bytes_total counter
test_read_bytes_total{client_id="primary",node_id="1"} 14029
test_read_bytes_total{client_id="primary",node_id="seed_0"} 1166
test_read_bytes_total{client_id="secondary",node_id="1"} 13876
test_read_bytes_total{client_id="secondary",node_id="seed_0"} 1166
# HELP test_write_bytes_total Total number of bytes written
# TYPE test_write_bytes_total counter
test_write_bytes_total{client_id="primary",node_id="1"} 10080
test_write_bytes_total{client_id="primary",node_id="seed_0"} 172
test_write_bytes_total{client_id="secondary",node_id="1"} 10072
test_write_bytes_total{client_id="secondary",node_id="seed_0"} 180
```

# Implementation
Since `kgo` doesn’t export `client_id` in a way that supports dynamic labeling in hooks, I had to make some modifications to the core library. My goal was to minimize changes to the kgo package itself and to avoid breaking any existing APIs (such as hooks). To that end, I chose not to add dynamic labels to the connection and buffered metrics, keeping those labels static.

### DIFF
--- a/pkg/kgo/hooks.go
+++ b/pkg/kgo/hooks.go
@@ -135,7 +135,7 @@ type BrokerE2E struct {
 	// ReadErr is any error encountered during reading.
 	ReadErr error
 
-	// ClientID is pointer to ID of the client that made the request
+	// ClientID is the ID of the client that made the request
 	ClientID string
 }
 
@@ -243,7 +243,7 @@ type ProduceBatchMetrics struct {
 	// zstd.
 	CompressionType uint8
 
-	// ClientID is pointer to ID of the client that made the request
+	// ClientID is the ID of the client that made the request
 	ClientID string
 }
 
@@ -295,7 +295,7 @@ type FetchBatchMetrics struct {
 	// zstd.
 	CompressionType uint8
 
-	// ClientID is pointer to ID of the client that made the request
+	// ClientID is the ID of the client that made the request
 	ClientID string
 }
 
--- a/pkg/kgo/hooks.go
+++ b/pkg/kgo/hooks.go
@@ -134,6 +134,9 @@ type BrokerE2E struct {
 	WriteErr error
 	// ReadErr is any error encountered during reading.
 	ReadErr error
+
+	// ClientID is pointer to ID of the client that made the request
+	ClientID string
 }
 
 // DurationE2E returns the e2e time from the start of when a request is written
@@ -239,6 +242,9 @@ type ProduceBatchMetrics struct {
 	// 0 is no compression, 1 is gzip, 2 is snappy, 3 is lz4, and 4 is
 	// zstd.
 	CompressionType uint8
+
+	// ClientID is pointer to ID of the client that made the request
+	ClientID string
 }
 
 // HookProduceBatchWritten is called whenever a batch is known to be
@@ -288,6 +294,9 @@ type FetchBatchMetrics struct {
 	// 0 is no compression, 1 is gzip, 2 is snappy, 3 is lz4, and 4 is
 	// zstd.
 	CompressionType uint8
+
+	// ClientID is pointer to ID of the client that made the request
+	ClientID string
 }
 
 // HookFetchBatchRead is called whenever a batch if read within the client.
@@ -417,4 +426,12 @@ func implementsAnyHook(h Hook) bool {
 		return true
 	}
 	return false
+}
+
+func (cl *Client) clientIDString() string {
+	resolved := "kgo"
+	if cl.cfg.id != nil {
+		resolved = *cl.cfg.id
+	}
+	return resolved
 }
--- a/pkg/kgo/sink.go
+++ b/pkg/kgo/sink.go
@@ -1738,6 +1738,7 @@ func (p produceMetrics) hook(cfg *cfg, br *broker) {
 		for _, h := range hooks {
 			for topic, partitions := range p {
 				for partition, metrics := range partitions {
+					metrics.ClientID = br.cl.clientIDString()
 					h.OnProduceBatchWritten(br.meta, topic, partition, metrics)
 				}
 			}
--- a/plugin/kprom/config.go
+++ b/plugin/kprom/config.go
@@ -96,9 +96,12 @@ func HandlerOpts(opts promhttp.HandlerOpts) Opt {
 	return opt{func(c *cfg) { c.handlerOpts = opts }}
 }
 
-// WithClientLabel adds a "cliend_id" label to all metrics.
+// WithClientLabel adds a "client_id" label to all metrics.
 func WithClientLabel() Opt {
-	return opt{func(c *cfg) { c.withClientLabel = true }}
+	return opt{func(c *cfg) {
+		c.withClientLabel = true
+		c.fetchProduceOpts.labels = append(c.fetchProduceOpts.labels, "client_id")
+	}}
 }
 
 // Subsystem sets the subsystem for the kprom metrics, overriding the default
@@ -182,6 +185,7 @@ type Detail uint8
 const (
 	ByNode            Detail = iota // Include label "node_id" for fetch and produce metrics.
 	ByTopic                         // Include label "topic" for fetch and produce metrics.
+	ByClient                        // Include label "client_id" for fetch and produce metrics
 	Batches                         // Report number of fetched and produced batches.
 	Records                         // Report the number of fetched and produced records.
 	CompressedBytes                 // Report the number of fetched and produced compressed bytes.
@@ -211,6 +215,8 @@ func FetchAndProduceDetail(details ...Detail) Opt {
 					labelsDeduped[ByTopic] = "topic"
 				case ByNode:
 					labelsDeduped[ByNode] = "node_id"
+				case ByClient:
+					labelsDeduped[ByClient] = "client_id"
 				case Batches:
 					c.fetchProduceOpts.batches = true
 				case Records:
--- a/plugin/kprom/config.go
+++ b/plugin/kprom/config.go
@@ -100,7 +100,6 @@ func HandlerOpts(opts promhttp.HandlerOpts) Opt {
 func WithClientLabel() Opt {
 	return opt{func(c *cfg) {
 		c.withClientLabel = true
-		c.fetchProduceOpts.labels = append(c.fetchProduceOpts.labels, "client_id")
 	}}
 }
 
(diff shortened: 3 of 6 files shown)


twmb:
I forgot this included changes to kgo when prepping for the release I just cut. I'm aiming to work on the next batch of requests (caching improvements largely) for 1.19 over the next month, but that's going to delay my merge of this. Really sorry about missing that.

In the meantime, my first comment is `ClientID is pointer to ID of the client that made the request` => these aren't pointers (maybe you started with a pointer); the comment can be touched up. I'll review the rest of this more closely soon.

The main thing I'll want to ensure is that the default metrics do _not_ change for anybody (no breaking change); any changes here will be opt in.

vtolstov:
may be create additional method to create hooks with some predefined labels? like NewHook(labels ...string) , so we can create hook that already populate labels for all metrics ? And default behaviour - fill client_id label like now?

pkg/kgo/hooks.go
```diff
@@ -134,6 +134,9 @@ type BrokerE2E struct {
 	WriteErr error
 	// ReadErr is any error encountered during reading.
 	ReadErr error
+
+	// ClientID is pointer to ID of the client that made the request
```
twmb, line 138: I suspect you originally had this a pointer (this doc and other docs can be updated)

plugin/kprom/kprom.go
```diff
@@ -125,14 +125,21 @@ func (m *Metrics) Handler() http.Handler {
...
 	var (
-		factory     = promauto.With(m.cfg.reg)
-		namespace   = m.cfg.namespace
-		subsystem   = m.cfg.subsystem
-		constLabels prometheus.Labels
+		factory   = promauto.With(m.cfg.reg)
+		namespace = m.cfg.namespace
+		subsystem = m.cfg.subsystem
 	)
+	var constLabels prometheus.Labels
```
twmb, line 132: No reason to move this down two lines? Or did you want it to be next to dynamicLabels?

plugin/kprom/kprom.go
```diff
@@ -146,195 +153,174 @@ func (m *Metrics) OnNewClient(client *kgo.Client) {
...
-		Namespace:   namespace,
-		Subsystem:   subsystem,
-		ConstLabels: constLabels,
-		Name:        "fetch_records_total",
-		Help:        "Total number of records fetched",
+		Namespace: namespace,
+		Subsystem: subsystem,
+		Name:      "fetch_records_total",
+		Help:      "Total number of records fetched",
 	}, m.cfg.fetchProduceOpts.labels)
```
twmb, line 324: Is there a reason not to add this to the two gauges below?

plugin/kprom/config.go
```diff
@@ -182,6 +185,7 @@ type Detail uint8
 const (
 	ByNode            Detail = iota // Include label "node_id" for fetch and produce metrics.
 	ByTopic                         // Include label "topic" for fetch and produce metrics.
+	ByClient                        // Include label "client_id" for fetch and produce metrics
```
twmb, line 188: Should WithClientLabel be deprecated in favor of this?

plugin/kprom/config.go
```diff
@@ -96,9 +96,12 @@ func HandlerOpts(opts promhttp.HandlerOpts) Opt {
 	return opt{func(c *cfg) { c.handlerOpts = opts }}
 }
 
-// WithClientLabel adds a "cliend_id" label to all metrics.
+// WithClientLabel adds a "client_id" label to all metrics.
 func WithClientLabel() Opt {
-	return opt{func(c *cfg) { c.withClientLabel = true }}
+	return opt{func(c *cfg) {
+		c.withClientLabel = true
+		c.fetchProduceOpts.labels = append(c.fetchProduceOpts.labels, "client_id")
```
twmb, line 103: Can this avoid the append if the client_id label is already present?

twmb (commented):
I'm finally gearing up to release the next minor version. If you're disinterested in this, I can take over the PR -- I myself have been frustrated having to wait an inordinately long time for a review (my case was >1yr). I can take over if you'd wish.

This lgtm generally, although I have a few small requests.

To confirm, this does _not_ change the formatting at all for existing uses?

lukasrynt:
Thank you for making a time for this PR!

Sorry for not responding sooner, but to answer your question. I designed this PR to not make any breaking changes to users of this library. So although one uses the `WithClientLabel` option it only splits the metrics for individual clients as one would expect. No changes to kgo hooks that would disrupt their usage were made, I only used already present metadata structures to pass in the required client ID.

I made four new commits to separate the content for review's sake. I plan to squash these before merging. One small addition that wasn't mentioned in the comment is returning constLabels to connection metrics. I accidentally removed these and forgot to return them back.

pkg/kgo/sink.go
```diff
@@ -1738,6 +1738,7 @@ func (p produceMetrics) hook(cfg *cfg, br *broker) {
 		for _, h := range hooks {
 			for topic, partitions := range p {
 				for partition, metrics := range partitions {
+					metrics.ClientID = br.cl.clientIDString()
```
twmb, line 1741: This only needs to be done once outside the loop, rather than per iteration

plugin/kprom/kprom.go
```diff
@@ -130,9 +131,19 @@ func (m *Metrics) OnNewClient(client *kgo.Client) {
...
 		constLabels = make(prometheus.Labels)
 		constLabels["client_id"] = client.OptValue(kgo.ClientID).(string)
+		dynamicLabels = append(dynamicLabels, "client_id")
+		if !slices.Contains(m.cfg.fetchProduceOpts.labels, "client_id") {
+			m.cfg.fetchProduceOpts.labels = append(m.cfg.fetchProduceOpts.labels, "client_id")
+		}
+	}
+
+	// Skip metrics definitions if hook was already called
+	if m.connConnectsTotal != nil {
```
twmb, line 145: This isn't concurrent safe -- do you intend to initialize clients concurrently?

plugin/kprom/kprom.go
twmb: I see the limitation. So, this PR is adding support to using the same kprom.Metrics in multiple clients, _but_ with the caveat that the _first_ client initialized sets a permanent client_id for these three metrics as well as the connect metrics above?

twmb: It looks like unfortunately a related Prometheus issue was closed with no comment and no resolution: https://github.com/prometheus/client_golang/issues/923

It doesn't look like it'd take _too_ much to add in a custom value / gauge func that actually allows custom labels, but this gets into guts that are a bit too detailed for me at the moment: https://github.com/prometheus/client_golang/blob/main/prometheus/value.go#L79

HOWEVER, one option could be to register a gauge func with a ConstLabels _per client id_ -- the promauto library has a similar example here: https://pkg.go.dev/github.com/prometheus/client_golang@v1.20.5/prometheus#example-GaugeFunc-ConstLabels where Namespace `mysql` & Name `connections_open` is registered twice, with the same label `destination` but the label value being different.

plugin/kprom/kprom.go
twmb: For what it's worth, if so, the logic can be extracted to a guarded initialization (and other hooks can have a quick check to see if initialization has completed). Let me know if you'd like me to add that.

plugin/kprom/kprom.go
twmb: I can make the change if you're interested in testing it?

lukasrynt:
@twmb returning back to your review I actually found out, that my testing setup was flawed. The approach where one would initiate the metrics separately per client instead of reusing the same object (while using the `kprom.Registerer`) would work (example below).

```
primaryCl := newClient(kgo.SeedBrokers(brokers...), kgo.ClientID("primary"))
secondaryCl := newClient(kgo.SeedBrokers(brokers...), kgo.ClientID("secondary"))
...

func newClient(opts ...kgo.Opt) *kgo.Client {
	metrics := kprom.NewMetrics("kafka_stats",
		kprom.Registerer(prometheus.DefaultRegisterer),
		kprom.WithClientLabel(),
	)
	opts = append(opts, kgo.WithHooks(metrics))
	cl, err := kgo.NewClient(opts...)
	if err != nil {
		panic(err)
	}
	return cl
}
```

Based on this, I suggest closing this PR and related issue. I'm terriblly sorry for wasting your time but I do really appreciate your review and insights, it gave me quite a lot. I hope to contribute in a more meaningful way next time. Thank you!

lukasrynt:
Yes, the comments are no longer relevant, thank you.

PR #826, opened 2024-09-20
Don't set nil config when seeding topics in kfake cluster

colega:
Setting the configs to `nil` causes it to panic later when trying to alter the topic configs, as [it only checks for entry in the map not being present, not for it not being nil](https://github.com/colega/franz-go/blob/f1cded5cc755374d6cac290c23da247ce5eb4e7a/pkg/kfake/data.go#L259-L261).


colega:
I don't see any tests here, I'd be happy to add one, but then I'd have to import `kgo` and `kadm` from `kfake`, is that ok?

twmb:
Looks good, sorry I missed this. No tag necessary since kfake is not tagged yet. Tests eh............. we'll leave that question for later (kfake is largely used for tests at the moment)

Issue #823, opened 2024-09-19
high number of allocations in `kgo.recordToRecord` function


ortuman:
As a result of the large volume of records generated, we’ve observed in one of our consumers a high number of allocations originated in the `kgo.recordToRecord` function (`~65%` of total allocations according to the attached screenshot, concretely in [this line](https://github.com/twmb/franz-go/blame/master/pkg/kgo/source.go#L1765)). This results in a performance degradation caused by GC overhead.

![Screenshot 2024-09-19 at 09 59 14](https://github.com/user-attachments/assets/f5e13d71-aebe-49ef-a726-7fbdc49c596b)

Has the possibility of reusing the generated records through a pool ever been considered in order to minimize this effect? Perhaps it could be offered as an optional parameter? I could propose a PR for your review if you consider it.

ortuman:
In https://github.com/twmb/franz-go/pull/827 I've proposed a possible solution to the problem.

twmb:
I see you took your proposal a good bit further in https://github.com/grafana/franz-go/pull/3. I used to have a `sync.Pool` for records internally, but removed it due to due to it being completely ineffective. Specifically with my prior usage internally, I'd pull thousands of records out of the pool (allocating them all) and then put them back in at a different independent time. GC would run, all records would be collected anyway, and the pool just wasn't useful.

PR 3 there I think is _sound_ but is tricky to follow. I think it's trying to solve a few goals:
1. Reuse the temporary byte slice that is created while decompressing
2. Reuse the temporary `[]kmsg.Record` slice that is created while processing a fetch response
3. Reuse the end-user `*kgo.Record`

I think the implementation does the job, but the code is pretty sketchy to analyze. As well, I think it's not working 100% _as intended_ in the case where you are consuming uncompressed data. Currently, if consuming uncompressed data, nothing is being acquired from the `DecompressBufferPool`, _but_ `reuse` still releases back into the pool. If consuming compressed _and_ uncompressed data, this byte slice that was never acquired originally (via non-compressed batches) is being acquired for decompressing batches, and the slice is persisting randomly around. If this is what's actually happening (i.e. my reading comprehension is working right now), it's _fine_, but certainly not the intent.

[...]

PR #827, opened 2024-09-23
`*kgo.Record` pooling support

ortuman:
This PR resolves https://github.com/twmb/franz-go/issues/823 by allowing to reuse consumed records via `kgo.(*Record).Reuse` method.

### DIFF
--- a/pkg/kgo/config.go
+++ b/pkg/kgo/config.go
@@ -151,6 +151,8 @@ type cfg struct {
 	partitions map[string]map[int32]Offset // partitions to directly consume from
 	regex      bool
 
+	recordsPool recordsPool
+
 	////////////////////////////
 	// CONSUMER GROUP SECTION //
 	////////////////////////////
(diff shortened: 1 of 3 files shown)


pkg/kgo/config.go
```diff
@@ -1347,6 +1349,18 @@ func ConsumeRegex() ConsumerOpt {
...
 
+// EnableRecordsPool sets the client to obtain the *kgo.Record objects from a pool,
+// in order to minimize the number of allocations.
+//
+// By enabling this option, the records returned by PollFetches/PollRecords
+// can be sent back to the pool via Reuse method in order to be recycled.
+//
+// This option is particularly useful for use cases where the volume of generated records is very high,
+// as it can negatively impact performance due to the extra GC overhead.
+func EnableRecordsPool() ConsumerOpt {
```
ortuman, line 1360: Truth is, I'm not 100% convinced about adding this option. For simplicity, we could always fetch the records directly from the pool, since if the user doesn't explicitly invoke the `Reuse` method, these will essentially always end up being allocated from the heap.

vtolstov:
so , i don't understand - records already get from pool or this is not planned?

PR #829, opened 2024-09-26
Add Opts method for sr.Client

andrewstucki:
This adds the equivalent of `kgo.Client.Opts()` to the `sr.Client` implementation. The use-case would be if you want to initialize a Schema Registry client with an existing Schema Registry client, but say do something like change the credentials being used for authentication.


RafalKorepta (approved):
LGTM

PR #836, opened 2024-10-11
kgo: add AssumeConsumersRequireStable()

### LINKED ISSUE #754: Use EOS sessions with BlockRebalanceOnPoll
Hi,
With the current session implementation, it seems like sessions will always abort the next transaction, regardless of if an rebalance actually happened after a poll or session.Begin(). A rebalance can happen completely out of the consume-produce loop, and still the next session.End() will fail. I suspect this is because of this `revoked` flag that is set on a revoke and only reverted after an session.End() call: https://github.com/twmb/franz-go/blob/master/pkg/kgo/txn.go#L104

Is there a way to make EOS sessions work with BlockRebalanceOnPoll and not abort on the next session.End()? I would like to prevent aborted transactions for example during controlled deployment rollouts.

Thanks!

twmb:
This allows group transact sessions to NOT internally force an abort if a revoke happens.

Closes #754.

### DIFF
--- a/pkg/kgo/config.go
+++ b/pkg/kgo/config.go
@@ -160,10 +160,11 @@ type cfg struct {
 	balancers  []GroupBalancer // balancers we can use
 	protocol   string          // "consumer" by default, expected to never be overridden
 
-	sessionTimeout    time.Duration
-	rebalanceTimeout  time.Duration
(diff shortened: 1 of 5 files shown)


twmb:
See comment on #754. This is hard to get correct.

PR #849, opened 2024-10-30
Add `CreateSchemaWithIDAndVersion()` to the sr package

mihaitodor:
This would come in handy when someone is trying to migrate schemas between two Schema Registry instances and they wish to retain the original IDs to avoid schema ID translation when migrating Kafka messages.

### DIFF
--- a/pkg/sr/api.go
+++ b/pkg/sr/api.go
@@ -93,13 +93,13 @@ type (
 		// a Kafka topic, and whether this is for a key or value. For example,
 		// "foo-key" would be the subject for the foo topic for serializing the
 		// key field of a record.
-		Subject string `json:"subject"`
+		Subject string `json:"subject,omitempty"`
 
 		// Version is the version of this subject.
-		Version int `json:"version"`
+		Version int `json:"version,omitempty"`
 
 		// ID is the globally unique ID of the schema.
-		ID int `json:"id"`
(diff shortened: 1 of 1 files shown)


pkg/sr/api.go
```diff
@@ -352,18 +352,32 @@ func (cl *Client) Schemas(ctx context.Context, subject string) ([]SubjectSchema,
...
 		ID int `json:"id"`
 	}
-	if err := cl.post(ctx, path, s, &id); err != nil {
-		return SubjectSchema{}, err
+	if id == -1 {
+		if err := cl.post(ctx, path, s, &into); err != nil {
+			return SubjectSchema{}, err
+		}
+	} else {
+		if err := cl.post(ctx, path, SubjectSchema{Schema: s, ID: id}, &into); err != nil {
```
twmb, line 375: Is this a RP specific thing, or the API in general?

mihaitodor: I found it in the Confluent docs here: https://docs.confluent.io/platform/current/schema-registry/installation/migrate.html#migrate-individual-schemas. See how they POST the schema ID. It’s not mentioned in the API reference though…

mihaitodor:
As per the Slack chat, I updated this new API to include the version ID and set the `omitempty` tags accordingly. I think it should work OK now.

PR #850, opened 2024-10-30
Specify the type for `sr.SoftDelete` and `sr.HardDelete`

mihaitodor:
Note: This is a breaking change, but I guess the impact is minimal.


twmb (approved):
Technically breaking, but non-existent impact.

PR #862, opened 2024-11-25
Fix grammar on InvalidRecord error

rodaine:


twmb:
The wording here is a copy of what's in Kafka itself, but it's fine to update.

PR #867, opened 2024-12-03
Add PreReq option sr API

mihaitodor:


pkg/sr/clientopt.go
```diff
@@ -85,6 +85,16 @@ func BearerToken(token string) ClientOpt {
 	}}
 }
 
+// PreReqFunc is a hook func which mutates a request.
+type PreReqFunc func(req *http.Request) error
+
+// PreReq sets a hook func to call before every request is sent.
+func PreReq(preReq PreReqFunc) ClientOpt {
```
twmb, line 92: I'd remove the defined type for the function and just have this be an inline function, wdyt?

mihaitodor: Good call! I'm not sure now why I left that in there, but it's not really useful so I removed it. Thanks!

twmb (approved):

Issue #875, opened 2024-12-14
discussion: a metrics plugin using github.com/VictoriaMetrics/metrics


vrischmann:
Hi,

this is not really an issue, I'm opening a discussion.

We use [VictoriaMetrics](https://victoriametrics.com) at work and its [metrics](https://github.com/VictoriaMetrics/metrics) library to expose metrics from Go programs. It's a lightweight alternative to the official prometheus client.
It's not necessary to use VictoriaMetrics, it's just another, more efficient way to expose metrics in the Prometheus text format (and its histogram is [easier to use](https://valyala.medium.com/improving-histogram-usability-for-prometheus-and-grafana-bc7e5df0e350)).

I'm wondering if you would be open to include a plugin like kvictoria or something that would expose metrics using the `metrics` library, that way we could benefit from the advantages of the `metrics` library.

I'm willing to work on this. My plan would be to essentially copy and adapt the kprom plugin.

Let me know what you think.

vtolstov:
[metrics](https://github.com/VictoriaMetrics/metrics) library does not suffort prometheus histograms

vrischmann:
> [metrics](https://github.com/VictoriaMetrics/metrics) library does not support prometheus histograms

Do you mean that Prometheus (the database) wouldn't know what to do with a Histogram from the metrics library ?
That's probably true but this plugin would be for users of VictoriaMetrics (the database).

vtolstov:
i'm use metrics library in some of my projects, but without prometheus like histograms metrics can't be collected by prometheus server. So this plugin can be useful only with victoriametrics server that collects metrcis

vrischmann:
Yes. That's my use case and it might also benefit other users of VictoriaMetrics server so I'd argue a plugin is still useful.

vtolstov:
nice to hear!

twmb:
I'm open to this, and will review/tag a release if you open a PR. Please let me know in the PR message if this involves changes to kgo because that'll need to be merged/tagged independently.

vrischmann:
Hi, thanks for the feedback. 

I've started working on this and I realized that I'll have the same problem that is addressed in #820, in fact I've had this problem at work because we sometimes use multiple clients per application.

As far as I can tell the changes in #820 will work for me too.

twmb:
Released as plugin/kvictoria/v1.0.0

PR #881, opened 2025-01-05
add the kvictoria plug-in to provide metrics

vrischmann:
This adds a plug-in to provide metrics using the [VictoriaMetrics/metrics](https://github.com/VictoriaMetrics/metrics) library as discussed in #875 

It's based on the `kprom` plug-in although I decided to not provide any option other than the possibility to set the subsystem: in my experience, every metrics available in the hooks are useful and I see no point in not including them (and if a user doesn't want some metrics they can [drop them during scraping](https://docs.victoriametrics.com/relabeling/#how-to-drop-metrics-during-scrape)).

There's no change needed in `kgo` but like I mentioned in #875, the PR #820 will be useful for this plug-in too because I'll be able to use the client id in more metrics; right now I can only use the client id in the various gauges tracking what's buffered.

I tested this locally with a small test program, a VictoriaMetrics instance that scrape it and a grafana dashboard to see if the metrics make sense. I can provide that if needed; as far as I can tell everything looks good.

Let me know if you have any question.


twmb:
Is this forward compatible if somebody _does_ want to add options to filter within the package in the future?
Should this wait for #820?

vrischmann:
> Is this forward compatible if somebody _does_ want to add options to filter within the package in the future? Should this wait for #820?

Yes, if in the future someone adds these options it won't have any impact on current users since they'll have to be explicitly added in the `NewMetrics` call by the user.

It can wait for #820, in that case you wouldn't have to do two reviews (this one and a future one that uses the newly added ClientID). Personally I'm fine either way.

twmb:
Looks like #820 can be closed (see latest comment). I was planning to get to this after #820, since that'll close, I'll get to this very soon. You may want to pull the client ID changes into this PR -- though that'll force this to wait for the next 1.19 release (and I'm working on some large things for it so it'll take a bit). If you're OK to merge without ClientID, I can prioritize this sooner.

vrischmann:
I'm OK merging as is, I'll look into opening another PR for the ClientID changes at another time.

plugin/kvictoria/kvictoria.go
```diff
@@ -0,0 +1,243 @@
...
+
+// OnGroupManageError implements the [kgo.HookBrokerThrottle] interface for metrics gathering.
+// This method is meant to be called by the hook system and not by the user.
+func (m *Metrics) OnGroupManageError(err error) {
+	labels := make(map[string]string)
+
+	var kerr *kerr.Error
+	if errors.As(err, &kerr) {
+		labels["error_message"] = kerr.Message
+	}
```
twmb, line 166:
```go
} else {
    labels["error_message"] = err.Error()
```
?

plugin/kvictoria/kvictoria.go
```diff
@@ -0,0 +1,243 @@
...
+		sets: make(map[string]*vm.Set),
+	}
+}
+
+// OnNewClient implements the [kgo.HookNewClient] interface for metrics gathering.
+// This method is meant to be called by the hook system and not by the user.
+func (m *Metrics) OnNewClient(client *kgo.Client) {
+	clientID := client.OptValue(kgo.ClientID).(string)
+
+	set, ok := m.sets[clientID]
```
twmb, line 55: This reads as if it's meant to support the ability for the same `Metrics` struct to be used in multiple `kgo.Client`'s. If that's the case, should this be guarded with a sync.Mutex, here and in OnClientClosed?

plugin/kvictoria/kvictoria.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+// OnClientClosed implements the [kgo.HookClientClosed] interface for metrics gathering.
+// This method is meant to be called by the hook system and not by the user.
+//
+// This will unregister all metrics that are scoped to the client id of the client provided.
+func (m *Metrics) OnClientClosed(client *kgo.Client) {
+	clientID := client.OptValue(kgo.ClientID).(string)
+
+	set, ok := m.sets[clientID]
```
twmb, line 86: It's common for people to use the same clientID in many clients -- going off my comment above, I wonder if this package should document that a new `Metrics` should be created per client?

Otherwise, here, this may unregister a set that is used by multiple clients.

plugin/kvictoria/kvictoria.go
```diff
@@ -0,0 +1,243 @@
...
+
+	if m.cfg.namespace != "" {
+		builder.WriteString(m.cfg.namespace + "_")
+	}
+	if m.cfg.subsystem != "" {
+		builder.WriteString(m.cfg.subsystem + "_")
+	}
+	builder.WriteString(name)
+
+	// Note: can't use maps.Keys yet because it needs Go 1.23+
```
twmb, line 220: You can enforce this in the go.mod to allow the import (I usually leave kgo with the oldest Go support, but other packages or plugins I allow being higher).

twmb (commented):
A few minor comments, sorry for the delay -- a lot going on in the last few months (and upcoming two).

plugin/kvictoria/kvictoria.go
vrischmann: good idea

plugin/kvictoria/kvictoria.go
vrischmann: Ok, rethinking about this more I think you're right. In that case I can also just drop the `sets` map and store a single set.

vrischmann:
@twmb thanks for the review, I made some changes that address your feedback if you can take a look.

Making `Metrics` explicitly per-client makes it easy to support adding the client_id to every metrics since I can just store it in the object, so the last commit does just that. No need for another follow up PR.

plugin/kvictoria/kvictoria.go
```diff
@@ -0,0 +1,311 @@
...
+//
+// [VictoriaMetrics/metrics]: https://github.com/VictoriaMetrics/metrics
+type Metrics struct {
+	cfg cfg
+
+	clientID string
+	set      *vm.Set
+}
+
+// NewMetrics returns a new Metrics that tracks metrics under the given namespace.
```
twmb, line 104: Definitely please document that a Metrics will panic if used for two kgo clients.

plugin/kvictoria/kvictoria.go
```diff
@@ -0,0 +1,311 @@
...
+	labels := map[string]string{
+		"client_id": m.clientID,
+		"node_id":   kgo.NodeName(meta.NodeID),
+	}
+
+	m.set.GetOrCreateCounter(m.buildName("disconnects_total", labels)).Inc()
+}
+
+// OnBrokerWrite is a noop implementation of [kgo.HookBrokerWrite], logic moved to OnBrokerE2E
+func (m *Metrics) OnBrokerWrite(meta kgo.BrokerMetadata, _ int16, bytesWritten int, _, _ time.Duration, err error) {
```
twmb, line 180: You can probably leave the no-op functions out, fwiw. There's no impact to not having them if they're no-ops anyway.

twmb (approved):
Two minor requests, lgtm otherwise. It looks like a fair amount of allocs happen per hook (the label map and building names), which is unfortunate, but I'm not really digging into the Victoria Metrics library to see how to avoid it.

vrischmann:
The two last commits address your requests.

And yeah, it does allocate often but the Victoria Metrics library doesn't help us here, it's barebones and expects the user to build the metric name.
We could make something better if performance ends up being bad.

twmb (approved):

PR #898, opened 2025-01-23
add missing group metadata to commit offsets call

imle:
Brokers requiring member id fail without this


twmb:
How are you using kadm? Generation and MemberID are required when a group is alive and has active members; both fields are -1 when the group has no active members. Admin commits (external to the group) are really only meant to be allowed when a group is not actively running. An admin commit to an active group doesn't persist -- the members that you are trying to commit for will do their own separate commit and overwrite whatever you just did externally.

PR #904, opened 2025-02-07
kgo: allow compressor & decompressors to be entirely user driven; allow users to inject memory pools

twmb:
* Exports the internal Compressor and Decompressor
* Adds WithCompressor and WithDecompressor options to allow end-user overrides
* Improves # of allocs when reading fetch responses by performing slab allocations
* Exports the function that processes fetch response partitions, allowing end-users to manually process fetch responses
* Adds a WithPools option and exposes three pools a user can provide to reuse memory while processing fetch responses
* Adds a DisableFetchCRCValidation consumer option, for if you really need this escape hatch (for #909)

### DIFF
--- a/pkg/kgo/compression.go
+++ b/pkg/kgo/compression.go
@@ -89,6 +89,9 @@ type Compressor interface {
 type Decompressor interface {
 	// Decompress decompresses src, which is compressed with codecType,
 	// and returns the decompressed data or an error.
+	//
+	// If the decompression codec type is CodecNone, this should return
+	// the input slice.
 	Decompress(src []byte, codecType CompressionCodecType) ([]byte, error)
 }
 
@@ -275,10 +278,13 @@ type decompressor struct {
 	ungzPool   sync.Pool
 	unlz4Pool  sync.Pool
 	unzstdPool sync.Pool
+	pools      pools
 }
 
 // DefaultDecompressor returns the default decompressor used by clients.
-func DefaultDecompressor() Decompressor {
+// The first pool provided, if any, that implements PoolDecompressBytes
+// will be used.
+func DefaultDecompressor(pools ...Pool) Decompressor {
 	d := &decompressor{
 		ungzPool: sync.Pool{
 			New: func() any { return new(gzip.Reader) },
@@ -299,6 +305,7 @@ func DefaultDecompressor() Decompressor {
 				return r
 			},
 		},
+		pools: pools,
 	}
 	return d
 }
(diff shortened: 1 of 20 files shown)


pkg/kgo/compression.go
```diff
@@ -233,9 +278,13 @@ type decompressor struct {
...
 	unlz4Pool  sync.Pool
 	unzstdPool sync.Pool
+	pools      pools
 }
 
-func newDecompressor() *decompressor {
+// DefaultDecompressor returns the default decompressor used by clients.
+// If a pool is provided and implements PoolDecompressBytes, it will be
+// used.
+func DefaultDecompressor(pools ...Pool) Decompressor {
```
dimitarvdimitrov, line 287: is the intention here that only one pool is passed (and the function takes a variadic because that pool is optional)? It could be just a language thing, but that wasn't immediately clear to me.

pkg/kgo/pools.go
```diff
@@ -0,0 +1,207 @@
...
+type pools []Pool
+
+func (ps pools) each(fn func(Pool)) {
+	for _, p := range ps {
+		fn(p)
+	}
+}
+
+/*TODO
+type PoolRequestBuffer interface {
```
dimitarvdimitrov, line 38: did you mean to implement this in this PR or only later on?

pkg/kgo/pools.go
```diff
@@ -0,0 +1,207 @@
...
+	if &ps.recs[idx] != r {
+		panic("sanity check pointer comparison index for recycling record is not the same!")
+	}
+
+	rem := ps.n.Add(-1)
+	if rem > 0 {
+		return
+	}
+
+	// Reset the capacity of slices to max to ensure we zero things and
```
dimitarvdimitrov, line 146: nitpick on the comment: this only resets the length of the slices, their capacity is still the same. Also the sentence in the comment isn't complete

dimitarvdimitrov (approved):
SGTM, I left a few small questions

pkg/kgo/compression.go
twmb: The first pool that implements PoolDecompressBytes will be used; the intention here is to allow the end user to use the same pools here that they use in the config option `WithPools` -- so, they can be a bit lazy if they have multiple separate pools rather than figure out which pool they should use.

pkg/kgo/pools.go
twmb: This will be implemented in this PR soon :)

pkg/kgo/compression.go
dimitarvdimitrov: ahaaa. Is it the first or the last? Currently the loop in `decompressor.Decompress` takes a slice for all implementations and keeps only the last one. 

Maybe mention this behaviour in a brief comment here as this is a public function (or a validation function and not rely on humans/AI reading comments),

twmb: Actually you're pointing out a bug -- the first pool that returns a slice should cause the `each` function to return early. I'm going to change `each` arg to `func(Pool) bool`, and if it returns true, stop iterating.

pkg/kgo/pools.go
twmb: Actually, I'm reverting the implementation. There are not many ways a person can really implement this buffer pool and the benefit of user configurability here seems low at the moment. If this is requested in the future, it can be added.

MasslessParticle:
Super excited for this PR. I tested the pooling on Friday and it works really well

twmb:
Good to hear @MasslessParticle, also thank you for testing it out. Let me know if you have any suggestions for API improvements; currently I'm relatively happy with the PR. I'm spending quite an bit of time now trying to address everything for 1.19; it'll be a large release, at least behind the scenes.

twmb:
@MasslessParticle would you be open to contributing an example? I could but it's ...likely very low on my priority list.

rockwotj:
nice

PR #907, opened 2025-02-14
kgo: add ctx field to client cfg

o-mago:
Adds `ctx` field to allow custom contexts to be used by the client
For example, it allows the custom `onAssigned` func, added using `OnPartitionsAssigned` opt, to access a custom context and any propagated value inside (like tracers)

### DIFF
--- a/pkg/kgo/client.go
+++ b/pkg/kgo/client.go
@@ -454,7 +454,13 @@ func NewClient(opts ...Opt) (*Client, error) {
 		}
 	}
 
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx := context.Background()
+
+	if cfg.ctx != nil {
+		ctx = cfg.ctx
+	}
+
+	ctx, cancel := context.WithCancel(ctx)
 
 	cl := &Client{
 		cfg:       cfg,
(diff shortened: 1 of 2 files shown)


o-mago:
could you review it @twmb 😊?

twmb:
Looks good, can you update the `Client.OptValue` method to return this context?

o-mago:
done @twmb

twmb:
Is it worth it to add a `Context() context.Context` function to the client, so that the client can return the internal context (which will be this context, wrapped with a cancellation)?

o-mago:
> Is it worth it to add a `Context() context.Context` function to the client, so that the client can return the internal context (which will be this context, wrapped with a cancellation)?

Makes sense, just added @twmb

pkg/kgo/client.go
```diff
@@ -529,6 +537,10 @@ func (cl *Client) Opts() []Opt {
 	return cl.opts
 }
```
twmb, line 539:
```go
// Context returns the internal context used wherever possible in the client.
// By default this is context.WithCancel(context.Background()). You may
// override the background context with your own via [WithContext].
// The context is occasionally wrapped further internally in client subsystems.
```

o-mago: nice, comment added

PR #917, opened 2025-02-25
Fix produce mutex to allow variables to be updated in blocking condition

erenboz:
We have hit this bug franz-go leaking whole lot of goroutines where we call produce without cancel from a spawned goroutine for each. Looking through the code, having mutex open prevents other goroutines to update variables the wait condition is reading on. Cause both `quit, overMaxRecs, overBytes` all depends on `p.mu`. Also looks like condition Wait also missing locks required by its semantics. 

```
func (c *Cond) Wait() {
	c.checker.check()
	t := runtime_notifyListAdd(&c.notify)
	c.L.Unlock()
	runtime_notifyListWait(&c.notify, t)
	c.L.Lock()
}
```

However, now this concurrency code looks fairly unnatural, might be better to re-write it eventually. It's likely hard to make a test to reproduce it in github actions environment that might have very limited actual cpu parallelism, I tried setting a high GOMAXPROCS but it still is passing with/without changes.

Issue https://github.com/twmb/franz-go/issues/918

### DIFF
--- a/pkg/kgo/producer.go
+++ b/pkg/kgo/producer.go
@@ -472,7 +472,9 @@ func (cl *Client) produce(
 			p.mu.Lock()
 			calcNums()
 			for !quit && (overMaxRecs || overMaxBytes) {
+				p.mu.Unlock()
 				p.c.Wait()
+				p.mu.Lock()
 				calcNums()
 			}
 			p.blocked.Add(-1)
@@ -481,7 +483,7 @@ func (cl *Client) produce(
 
 		drainBuffered := func(err error) {
 			// The expected case here is that a context was
-			// canceled while we we waiting for space, so we are
+			// canceled while we waiting for space, so we are
 			// exiting and need to kill the goro above.
 			//
 			// However, it is possible that the goro above has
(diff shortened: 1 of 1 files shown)


erenboz:
Oh I missed 
```
	p.c = sync.NewCond(&p.mu)
```
That ensures that bit is correct, so deadlock must come from somewhere else.

> What's your usage look like, and the stack dump of what leads you to believe Produce is leaking?
I was sifting through past deadlock titled issues and that bit of code read weird and decreasing concurrency in the service and increasing maxbufferedrecords from default 10k to 50k seemed to help. That suggest it should be still around produce blocking condition even if not directly there.

Our use-case was producing into 5-6 big topics with varying amount of partitions up to like 700-800. And our code was mistakenly calling goroutines from uncancellable dedicated goroutines for individual produce calls.

erenboz:
We'll try to set something reproducing in prod to find whereabouts

erenboz:
> We'll try to set something reproducing in prod to find whereabouts

Commented on the [issue](https://github.com/twmb/franz-go/issues/918#issuecomment-2684939970)

twmb:
Closing pending issue investigation (that said, I need to work in that comment fix someday)

PR #920, opened 2025-03-03
kgo: fix race condition in ProduceSync

daedric:
While reading the code I stumbled upon a potential race condition.
Here is my suggestion for a fix.


twmb:
Luckily not a race; Produce finishes all promises in order in a dedicated goroutine https://github.com/twmb/franz-go/blob/f30c518d6b727b9169a90b8c10e2127301822a3a/pkg/kgo/producer.go#L573

daedric:
Oh thanks, that was subtle :)

PR #921, opened 2025-03-04
kgo: fix cancellation of a fetch in manageFetchConcurrency

iimos:
I discovered a "bug" in the code responsible for handling the cancellation of a fetch request. Fortunately, it doesn’t seem to cause any immediate issues (if I’ve correctly understood the relationship between `manageFetchConcurrency` and `fetchLoop`). However I think it’s better to fix it now to prevent potential problems in the future when changes are made.

In this snippet, instead of deleting the `i`-th element, the last element is being removed:
```go
_ = append(wantFetch[i:], wantFetch[i+1:]...)
wantFetch = wantFetch[:len(wantFetch)-1]
```

I also attempted to write a test to cover this case, but it would require a significant redesign of `manageFetchConcurrency` so I’ve set this idea aside.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,5 +4,6 @@ dev
 origin
 TESTING
 generate/generate
+.idea
 go.work
 go.work.sum
--- a/pkg/kgo/consumer.go
+++ b/pkg/kgo/consumer.go
@@ -1541,9 +1541,9 @@ func (s *consumerSession) manageFetchConcurrency() {
 			var found bool
 			for i, want := range wantFetch {
 				if want == cancel {
-					_ = append(wantFetch[i:], wantFetch[i+1:]...)
-					wantFetch = wantFetch[:len(wantFetch)-1]
+					wantFetch = append(wantFetch[:i], wantFetch[i+1:]...)
 					found = true
+					break
 				}
 			}
 			// If we did not find the channel, then we have already


twmb:
Actually, looking more closely, your form is a more succinct version of the existing code. The current code shifts the end of the slice down by one and trims off the end. Your code follows the more standard pattern of appending out the element you want to delete which automatically shifts the remaining slice down and trims off the end. The net new, besides cleaner & more idiomatic code, here is a `break`.

Overall, +1 thank you for the cleanup :)

iimos:
> your form is a more succinct version of the existing code

This is false. Here is the reproducer: https://go.dev/play/p/lMNhSm7abox


> The current code shifts the end of the slice down by one and trims off the end.
Look carefully at:
```go
_ = append(wantFetch[i:], wantFetch[i+1:]...)
```
Here is the **tail is appended to the tail**. To shift the tail down by one, two letters should be swapped: `[i:]` -> `[:i]`
This doesn't cause an issue only because it results in the recreation of the entire `wantFetch` slice, which is then discarded due to the `_ =`

iimos:
The linter is complaining that `.golangci.yml` doesn't match json schema. Should i fix it in this PR or will it be handled without my involvement?

twmb:
> Here is the tail is appended to the tail. To shift the tail down by one, two letters should be swapped: [i:] -> [:i]

Ah yes I do have a typo here, you're right. _Without_ the type, it'd shift down in place and not recreate the slice. _With_ the typo, it's a complete bug. I'm going to stare at this a bit to try to understand why it doesn't affect things today, and then merge it into the next release.

We can ignore the linter. I disabled it in a [separate PR](https://github.com/twmb/franz-go/pull/912/commits/4cb5c926b921e1bf9c0baf86a8e741841256018b) because it's annoyingly breaking here.

Issue #926, opened 2025-03-10
impossible to create two clients with metrics


gstarikov:
Hi, 
It is impossible to create two clients with different options  and with metrics with different labels.

here ( https://github.com/twmb/franz-go/blob/f30c518d6b727b9169a90b8c10e2127301822a3a/plugin/kprom/kprom.go#L128 ) is used promauto.With and there is an internal MustRegister call, so there is no way to create 2 clients with metrics, even with the same labels, etc.

I suggest adding an option to add a custom static label. In this case, clients will be distinguished at the metrics level and registration will take place ( https://github.com/prometheus/client_golang/blob/45edd8a6b61330c6cc4e525ab165868902a2bb9e/prometheus/registry.go#L297 )

stampy88:
@gstarikov I have created multiple clients with different metrics in the same app doing the following:
```
m := kprom.NewMetrics(
	mf.Namespace(),
	kprom.Registry(mf),
	kprom.Subsystem(c.subsystem),
	kprom.FetchAndProduceDetail(kprom.Batches, kprom.Records, kprom.UncompressedBytes, kprom.ByTopic, kprom.ByNode, kprom.ConsistentNaming),
)

franzOpts := []kgo.Opt{
	kgo.SeedBrokers(c.brokers...),
	kgo.WithHooks(m),
}
```
`mf` is an abstraction over a prometheus Registry.

PR #927, opened 2025-03-10
added static label

gstarikov:
https://github.com/twmb/franz-go/issues/926

### DIFF
--- a/plugin/kprom/config.go
+++ b/plugin/kprom/config.go
@@ -3,6 +3,7 @@ package kprom
 import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"maps"
 )
 
 type cfg struct {
@@ -13,6 +14,7 @@ type cfg struct {
 	gatherer prometheus.Gatherer
 
 	withClientLabel  bool
+	withConstLabels  prometheus.Labels
 	histograms       map[Histogram][]float64
 	defBuckets       []float64
 	fetchProduceOpts fetchProduceOpts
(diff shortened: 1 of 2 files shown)


twmb:
Not opposed, but is this necessary or can the approach [here](https://github.com/twmb/franz-go/pull/820#issuecomment-2642259196) suffice / do well enough?

gstarikov:
I'm not an expert in the Kafka protocol, but as I see it, client_id goes somewhere into the protocol and seems to be used on the Kafka side for identification. i.e. it should be unique at least within the consumer group.

as i googled

> The sole purpose of this is to be able to track the source of requests beyond just ip and port by allowing a logical application name to be included in Kafka logs and monitoring aggregates.

In our code, there may be several clients configured for different tasks with different parameters (DQL, reader, writer & etc) and it is not uniq at all. but yes, if we don't look at the transport level, but only at the labels, then we can use client_id as done in https://github.com/twmb/franz-go/pull/820 ticket.

I'm not sure, but I wouldn't want to mix the label for metrics and id at the transport level.
and all our libraries use the "alias" label to filter metrics... I wouldn't want to switch to client_id in just one place...

plugin/kprom/config.go
```diff
@@ -101,6 +103,14 @@ func WithClientLabel() Opt {
 	return opt{func(c *cfg) { c.withClientLabel = true }}
 }
 
+// WithStaticLabel adds a static label to all metrics.
+func WithStaticLabel(label, value string) Opt {
```
twmb, line 107: Is there a use case to a person wanting multiple static labels, such that this should be `labelvalues ...string)`?

twmb:
+1 and I see no downside to this PR anyway.

plugin/kprom/config.go
```diff
@@ -101,6 +103,13 @@ func WithClientLabel() Opt {
 	return opt{func(c *cfg) { c.withClientLabel = true }}
 }
 
+// WithStaticLabel adds a static label to all metrics.
```
twmb, line 106: Sorry one last nit -- is it worth noting that the `client_id` const label is always unconditionally set? Should we allow the end user to be able to override it, or should the current behavior be kept?

(separately: good idea on just allowing the user to pass in prometheus.Labels, that's way better than my keyvals suggestion)

gstarikov: Yes, the collision of metrics is unpleasant. But I would leave the current behavior, otherwise suddenly someone's  something will break from this change. And according to Occam's razor, new entities should not be invented. And if someone has better ideas, then he can come with an issue or mr

twmb:
Merging, but awaiting for comments on #936 before releasing.

twmb:
Released in plugin/kprom/v1.2.0

PR #928, opened 2025-03-10
kgo: minor improvements while attempting to work around a Kafka hanging transaction bug

twmb:
What this adds:
* After EndTxn, we send ListTransactions to verify the transaction has actually completed
* If the transaction hangs for 10s, we log and set internally to reinitialize the producer ID.
* We log a lot

Some logging changes will be kept; the ListTransactions verification will be deleted, as well as the hard-resetting of the producerID.

### DIFF
--- a/pkg/kgo/group_test.go
+++ b/pkg/kgo/group_test.go
@@ -25,7 +25,7 @@ func TestGroupETL(t *testing.T) {
 	t.Parallel()
 
 	topic1, topic1Cleanup := tmpTopic(t)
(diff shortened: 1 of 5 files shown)


twmb:
ref: https://issues.apache.org/jira/browse/KAFKA-18957

twmb:
Closing, none of this is worth keeping in light of EndTxn v5.

twmb:
Actually, some of this is worth keeping -- reopening and paring down.

Issue #914, opened 2025-02-21
discussion: kprom improvement, following collector pattern


loginakhil:
Hello there,

Thank you for your hard work in implementing such an awesome library.

Upon exploring [kprom](https://github.com/twmb/franz-go/tree/master/plugin/kprom), I noticed that this module implements the Prometheus collector differently. 
Instead of implementing a Prometheus collector like below:

```go
// NewCustomCollector creates a new CustomCollector
func NewCustomCollector() *CustomCollector {
	counter := prometheus.NewCounterVec(
		prometheus.CounterOpts{
			Name: "my_custom_counter",
			Help: "A custom counter for demonstration purposes",
		},
		[]string{"label"},
	)

	prometheus.MustRegister(counter)

	return &CustomCollector{
		counter: counter,
	}
}

// Collect implements the Collector interface
func (c *CustomCollector) Collect(ch chan<- prometheus.Metric) {
	// Simulate some metrics
	c.counter.WithLabelValues("example").Inc()
	c.counter.WithLabelValues("example2").Inc()

	// Collect the metrics
	c.counter.Collect(ch)
}

// Describe implements the Collector interface
func (c *CustomCollector) Describe(ch chan<- *prometheus.Desc) {
	c.counter.Describe(ch)
}
```

Is there any specific reason behind this implementation style, or would you be open to approving a pull request for this modification?

twmb:
What's the benefit of a change?

loginakhil:
This will allow applications that use `kprom` to register multiple collectors to the same registry by inversing the dependency.
We will then be following the recommended approach of collector implementation 
- https://github.com/prometheus/client_golang/blob/main/prometheus/collector.go#L16-L18

twmb:
I haven't seen a collector implemented like that yet, I've only ever seen people register metrics with an existing collector (i.e., what kprom currently does).

That said, is it possible to add the APIs you're suggesting in a non-breaking way?

loginakhil:
@twmb I have created the following PR that adds this change in a backward compatible manner.

- https://github.com/twmb/franz-go/pull/936

Could you please review the changes 🙏

PR #936, opened 2025-03-19
feat: collector pattern for kprom

loginakhil:
Fixes https://github.com/twmb/franz-go/issues/914

### DIFF
--- a/plugin/kprom/config.go
+++ b/plugin/kprom/config.go
@@ -2,6 +2,7 @@ package kprom
 
 import (
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/collectors"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
@@ -12,10 +13,11 @@ type cfg struct {
 	reg      prometheus.Registerer
 	gatherer prometheus.Gatherer
 
-	withClientLabel  bool
-	histograms       map[Histogram][]float64
-	defBuckets       []float64
-	fetchProduceOpts fetchProduceOpts
+	withSkipRegisterer bool
+	withClientLabel    bool
+	histograms         map[Histogram][]float64
+	defBuckets         []float64
+	fetchProduceOpts   fetchProduceOpts
 
 	handlerOpts  promhttp.HandlerOpts
 	goCollectors bool
@@ -40,8 +42,8 @@ func newCfg(namespace string, opts ...Opt) cfg {
 	}
 
 	if cfg.goCollectors {
-		cfg.reg.MustRegister(prometheus.NewProcessCollector(prometheus.ProcessCollectorOpts{}))
-		cfg.reg.MustRegister(prometheus.NewGoCollector())
+		cfg.reg.MustRegister(collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}))
+		cfg.reg.MustRegister(collectors.NewGoCollector())
 	}
 
 	return cfg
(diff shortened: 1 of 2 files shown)


plugin/kprom/kprom.go
```diff
@@ -360,33 +366,42 @@ func (m *Metrics) OnNewClient(client *kgo.Client) {
 		},
 		func() float64 { return float64(client.BufferedFetchRecords()) },
 	)
+
+	m.allMetricCollectors = append(m.allMetricCollectors,
```
twmb, line 370: Is there a reason to not keep the old code, gated behind the new conditional below? This looks to add a slice that has the same amount of lines and requires the same updates whenever a new metric is added.

plugin/kprom/config.go
```diff
@@ -81,8 +83,15 @@ func Gatherer(gatherer prometheus.Gatherer) Opt {
 	return opt{func(c *cfg) { c.gatherer = gatherer }}
 }
 
+// SkipRegisterer allows registering an external registry,
+// exposing kprom as a custom prometheus collector
+// This option is mutually exclusive with Registerer & GoCollectors
+func SkipRegisterer() Opt {
```
twmb, line 89: Would it be better to change the `Register` option to accept `nil` as a way to disable the registerer?

plugin/kprom/kprom.go
```diff
@@ -496,6 +511,18 @@ func (m *Metrics) OnBrokerE2E(meta kgo.BrokerMetadata, _ int16, e2e kgo.BrokerE2
 	}
 }
 
+func (m *Metrics) Collect(ch chan<- prometheus.Metric) {
```
twmb, line 514: Might be worth documenting that Collect and Desc exist to implement ... whatever the interface is, if the person wants to use a `Metrics` that way.

twmb:
@loginakhil ping

plugin/kprom/kprom.go
loginakhil: I introduced this flag to keep the existing behaviour untouched.
But your suggestion of `nil` check on `Register` also will work.
I have updated the code in the latest commit.

plugin/kprom/config.go
loginakhil: Makes sense, pushed new changes ✅

plugin/kprom/kprom.go
loginakhil: Added the documentation part ✅

loginakhil:
@twmb Thank you for the review. 🙌 

I have applied your suggestions and also fixed some typos I made in the the first commit.

Could you please take another look 🙏

plugin/kprom/config.go
```diff
@@ -64,10 +67,18 @@ type RegistererGatherer interface {
...
 // new registry. Use this option if you want to configure both Gatherer and
 // Registerer with the same object.
+// Additionally, passing a `nil`  RegistererGatherer allows registering an external registry,
+// exposing kprom as a custom prometheus collector, this option is mutually
+// exclusive with GoCollectors
 func Registry(rg RegistererGatherer) Opt {
 	return opt{func(c *cfg) {
-		c.reg = rg
-		c.gatherer = rg
+		if rg == nil || (reflect.ValueOf(rg).Kind() == reflect.Ptr && reflect.ValueOf(rg).IsNil()) {
```
twmb, line 75: The latter half of this conditional is a bit paranoid -- I've never seen this -- but it doesn't seem to hurt, so ok 👍

twmb (approved):

twmb:
Released in plugin/kprom/v1.2.0

Issue #925, opened 2025-03-10
ListTopics is actually a DescribeTopics


ward-eric:
I have a use case where I'm managing multiple Kafka clusters and I need to fetch the list of topics from each cluster, but the `ListTopics` call is very slow (10 seconds per cluster in franz-go vs 2 seconds with the java client or sarama). Upon further inspection the franz-go library doesn't seem to have a true `ListTopics` as it performs the metadata lookup that the `DescribeTopics` call does in other libraries.

Can we get a new method that just returns the topic listings? Failing that could you expose an option to disable `IncludeTopicAuthorizedOperations` in the existing call? That appears to be the main offender here. Though now that I say that I guess what I actually need is a `DescribeTopics`, but with the option to disabled the auth lookup like the java library gives us: https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/admin/DescribeTopicsOptions.java#L41

twmb:
A true `ListTopics` doesn't exist, that only exists via `Metadata`.

Are you specifically referring to this?
```go
$ rg IncludeTopicAuthorizedOperations
kadm/metadata.go
233:	req.IncludeTopicAuthorizedOperations = true
```

I'm surprised including auth options takes a while, what time difference are you seeing with that being false vs. true? Is it, well, 8s as hinted in the parentheses? I only added that recently because I figured it was super cheap, I'm open to it being switched to default false and having a kadm.Client method that swaps it to default true for every request issued.

ward-eric:
I don't have a strong opinion other than to agree with you that I likes options one and three much more than two. Three seems like it is more flexible if people want to conditionally fetch authorized ops within the same runtime.

twmb:
Sorry for the delay; I'm going to release this in pkg/kadm/v1.16.0 now. I wanted to get a kadm Lag calculation improvement in at the same time, which turned out to be an easier commit than I initially thought.

I was hoping to add a few more features in, but I'm moving those to v1.17.0. The next release is blocked on kgo 1.19, which I hope to get in soon but we'll see.

twmb:
Closing, however if you see this, question: in your PR, why are we still requesting cluster authorized options by default? Are cluster auth ops not a problem?

PR #939, opened 2025-03-24
Disable metadata fetch of authorized operations by default

ward-eric:
Based on the conversation here: https://github.com/twmb/franz-go/issues/925

### DIFF
--- a/pkg/kadm/metadata.go
+++ b/pkg/kadm/metadata.go
@@ -230,7 +230,7 @@ func (cl *Client) Metadata(
 func (cl *Client) metadata(ctx context.Context, noTopics bool, topics []string) (Metadata, error) {
 	req := kmsg.NewPtrMetadataRequest()
 	req.IncludeClusterAuthorizedOperations = true
-	req.IncludeTopicAuthorizedOperations = true
+	req.IncludeTopicAuthorizedOperations = ctx.Value(&includeAuthOps) != nil
 	for _, t := range topics {
 		rt := kmsg.NewMetadataRequestTopic()
 		rt.Topic = kmsg.StringPtr(t)
--- a/pkg/kadm/topics.go
+++ b/pkg/kadm/topics.go
@@ -9,6 +9,8 @@ import (
 	"github.com/twmb/franz-go/pkg/kmsg"
 )
 
+var includeAuthOps = "include_auth_ops"
+
 // ListTopics issues a metadata request and returns TopicDetails. Specific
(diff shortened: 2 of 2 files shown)


pkg/kadm/topics.go
```diff
@@ -9,6 +9,10 @@ import (
 	"github.com/twmb/franz-go/pkg/kmsg"
 )
 
+const (
+	IncludeTopicAuthorizedOperations = "IncludeTopicAuthorizedOperations"
```
twmb, line 13: This can be private, no need to expose to end users.

```go
var includeAuthOps = "include_auth_ops"
```

and then `context.WithValue(&includeAuthOps, true)`

The reason to use a pointer here (which forces it to not be a const) is that stuffing a pointer into an interface doesn't cause an allocation -- and it is _certainly_ critical to performance in these code paths.

twmb:
Thanks, sorry for my delay -- investigating the CI failure on #912 caused me a huge delay and then I had / have some big personal life interruptions. Left a minor comment but I can merge & release once addressed.

pkg/kadm/topics.go
ward-eric: ah, thanks! I've been in Java land for like a decade now and only really picked golang in the last year or so. I appreciate the pointers (no pun intended 😃 )

pkg/kadm/metadata.go
```diff
@@ -227,10 +227,19 @@ func (cl *Client) Metadata(
 	return cl.metadata(ctx, false, topics)
 }
 
+func shouldIncludeClusterAuthorizedOperations(ctx context.Context) bool {
+	if value := ctx.Value(&includeAuthOps); value != nil {
+		if includeTopicAuthorizedOperations, ok := value.(bool); ok {
+			return includeTopicAuthorizedOperations
+		}
```
twmb, line 234: The most minor nit that doesn't need to be pushed -- one thing you could do is to have the value be an empty `struct{}`. The presence of the key is indication enough that the option was opted into. The conditional wrapping this return can be eliminated, too -- if the value exists, then the option was set and we can return true here.

pkg/kadm/topics.go
twmb: I should have noticed with a function name as long as `shouldIncludeClusterAuthorizedOperations`

Kidding ;) looks great now, ty for the push. Let me know if you want to address the latest nit otherwise lgtm.

twmb (approved):

Issue #953, opened 2025-04-09
kgo#Client.Produce context.WithTimeout


randomizedcoder:
G'day,

I hope you are doing well.

I really like the franz-go client, and it works very well.

However, the behavior of calling Produce ( https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#Client.Produce ) with a context that has a timeout results in "err:context canceled", which surprises me.  This may be my misunderstanding about what a timeout should mean when calling Produce, so please forgive me for the miss understanding.  If it is intended, then it might be great to add some words about why in the docs.

My expectation is that you could call Produce with a short timeout, like maybe 1-3 seconds, so that if this single produce call somehow resulted in an issue, then my application code could decide what to do next.

e.g. The calling pattern I would expect is something like
```
// wrap my current ctx with a short timeout
ctxP, cancelP = context.WithTimeout(ctx, 1 * time.Second )
defer cancelP()
	g.kClient.Produce(
		ctxP,
		kgoRecord,
		func(kgoRecord *kgo.Record, err error) {
...
```

Doing this will result in the a Produce error, where the context is cancelled ( oh and I suppose there's a typo missing the 2nd "l" char )
```
2025/04/09 20:29:56.543454 destinations.go:90: destKafka 0.000509s Produce err:context canceled
2025/04/09 20:29:56.543483 destinations.go:90: destKafka 0.000780s Produce err:context canceled
2025/04/09 20:29:56.543531 destinations.go:90: destKafka 0.000658s Produce err:context canceled
2025/04/09 20:29:56.543552 destinations.go:90: destKafka 0.000692s Produce err:context canceled
2025/04/09 20:29:56.543597 destinations.go:90: destKafka 0.000633s Produce err:context canceled
```
Within the franz-go examples and tests, all cases of Produce() are using background, so I don't know if this behavior is well tested.

Throughout our go code, we make extensive use of contexts and we usually have timeouts, so unfortunately, when using franz-go we need to change the normal pattern.   ( Normally, if I see context.Background() in any of the code coming through our code review channel, it's a bug. There are x2 cases that it's not a bug. #1 tests, #2 and func main can have a context.Background().  After that everything should receive and pass a context.  It is normal to wrap the passed context with an even shorter context. e.g. Incoming GRPC request should have a timeout, and then we will wrap in the coming ctx in a shorter timeout, and this will bubble its way through the code to result in database timeouts, etc.  It all works pretty nicely. If you don't do this, things are really bad. )

[...]

PR #957, opened 2025-04-10
Produce with ctx

randomizedcoder:
This small example demonstrates a per message context with a timeout.

We _do_ expect franz-go kgo.Client to fully support context deadlines.

Clarifying: https://github.com/twmb/franz-go/issues/953


twmb:
awesome, thanks!

PR #963, opened 2025-04-11
sr: add `Client.OptValue` and `Client.OptValues` functions

### LINKED ISSUE #954: Add `Client.OptValue` and `Client.OptValues` to `sr` package
Similar to the `Client` type in the `kgo` package, it would be useful to have `Client.OptValue` and `Client.OptValues` functions for the `Client` type in the `sr` package. This would allow to retrieve values used in options, which can be useful for testing or option value retrieval in existing client objects.

If you agree with this request, I can give it a try to implement it and send a PR. 👍

hhromic:
This PR adds `Client:OptValue` and `Client.OptValues` functions to the `sr` package.
These are heavily based on the functions of the same name in the `kgo` package.

~~Note that I did not add a case and unit test for the `DialTLSConfig` option because this option creates a new `http.Client` object with TLS preconfigured instead of storing the TLS config in the client struct.~~ EDIT: Added now after merged #978.

Fixes #954

### DIFF
--- a/pkg/sr/client.go
+++ b/pkg/sr/client.go
@@ -26,6 +26,9 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"reflect"
+	"runtime"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


twmb (commented):
+1. Do you think DialTLSConfig should be switched to store the config -- and then initialize the client in NewClient?

hhromic:
The missing OptValue for DialTLSConfig for sure felt like a sore thumb :')
I was thinking that indeed that option should be switched to store the config, however there is some "magic" done in the DIalTLSConfig option setter that I believe needs some discussion.

Given that you seem open to improve on this, I will open an issue to discuss that further as I think it would be material for another PR separate from this one. Do you agree?

twmb:
@hhromic this can be updated now with the recent httpclient PR merge. There's also #979 but I'm not sure on the final API for that yet, so that can be updated separately later (though a `sr` release will likely wait for 979).

hhromic:
> @hhromic this can be updated now with the recent httpclient PR merge. There's also #979 but I'm not sure on the final API for that yet, so that can be updated separately later (though a `sr` release will likely wait for 979).

@twmb updated now after #978 👍 
We can merge this PR and I can open a new PR after #979 is merged.

twmb (approved):

pkg/sr/client.go
```diff
@@ -115,13 +118,98 @@ func NewClient(opts ...ClientOpt) (*Client, error) {
...
+	case namefn(HTTPClient):
+		return []any{cl.httpcl}
+	case namefn(UserAgent):
+		return []any{cl.ua}
+	case namefn(URLs):
+		return []any{cl.urls}
+	case namefn(DialTLSConfig):
+		return []any{cl.dialTLS}
+	case namefn(BasicAuth):
+		return []any{cl.basicAuth}
```
twmb, line 201: Shouldn't this return the user and pass as two separate arguments?

pkg/sr/client.go
```diff
@@ -115,13 +118,98 @@ func NewClient(opts ...ClientOpt) (*Client, error) {
...
+func (cl *Client) OptValue(opt any) any {
+	vs := cl.OptValues(opt)
+	if len(vs) > 0 {
+		return vs[0]
+	}
+	return nil
+}
+
+// OptValues returns all values for options. This method is useful for
+// options that have multiple inputs (notably, URLs). This is also useful
```
twmb, line 169: notably basic auth here?

pkg/sr/client.go
hhromic: Ah yeah, good call! Will do.
I understood this better after working more in the code base in the other PR :)

hhromic:
@twmb thanks for reviewing. I now switched to returning both values for the `BasicAuth` option.
I also implemented a unit test for this case and reworded the new functions documentation.
Let me know if you think something else is needed.

pkg/sr/client.go
```diff
@@ -115,13 +118,99 @@ func NewClient(opts ...ClientOpt) (*Client, error) {
...
+	vs := cl.OptValues(opt)
+	if len(vs) > 0 {
+		return vs[0]
+	}
+	return nil
+}
+
+// OptValues returns all values for options. This method is useful for
+// options that have multiple inputs (notably, BasicAuth). This is also useful
+// for options that are internally stored as a pointer -- tThis function will
```
twmb, line 170: `tThis` => this

twmb (approved):
one comment typo otherwise +1

hhromic:
Doh, sorry for the typo. Fixed now! 👍

twmb (approved):

PR #973, opened 2025-04-15
kmsg: properly annotate DescribeGroups.Group.ErrorMessage as v6+

### LINKED ISSUE #972: Generated code for DescribeGroupsResponse appears incorrect
It appears the generated code for `DescribeGroupsResponse` is incorrect for versions < 6: https://github.com/twmb/franz-go/blob/faf3af8ab001ca1b6c7fd8624d9ca55664bc88e5/pkg/kmsg/generated.go#L15814-L15821

In the Kafka DescribeGroupsResponse, the ErrorMessage is added after the ErrorCode, but it should be parsed only if the version is >= 6: https://github.com/apache/kafka/blob/fb2ce76b495338a26af0542e21ff15e983c7e558/clients/src/main/resources/common/message/DescribeGroupsResponse.json#L40-L41. It appears to be parsed unconditionally, which causes a failure reading responses for prior versions.

twmb:
Closes #972.

### DIFF
--- a/generate/definitions/15_describe_groups
+++ b/generate/definitions/15_describe_groups
@@ -30,7 +30,7 @@ DescribeGroupsResponse =>
     // GROUP_ID_NOT_FOUND is returned on v6+ if the group ID is not found (KIP-1043).
     ErrorCode: int16
     // ErrorMessage is an optional message with more detail for the error code.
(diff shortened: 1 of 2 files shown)


chrisseto (approved):
LGTM

PR #978, opened 2025-04-17
sr: refactor DialTLSConfig option so it is stored in client

### LINKED ISSUE #971: sr: refactor DialTLSConfig option so it is stored in struct
At the moment, the DialTLSConfig option setter creates and sets a custom `http.Client` instance with the TLS config applied:
https://github.com/twmb/franz-go/blob/faf3af8ab001ca1b6c7fd8624d9ca55664bc88e5/pkg/sr/clientopt.go#L47-L68

The provided TLS config is not stored anywhere else and becomes a part of the HTTP Client itself, through the HTTP Transport object.

The proposal is to refactor this so the TLS config is stored in `sr.Client` instead, similar to how `kgo.Client` does it.
This will allow for `OptValue` and `OptValues` to be able to return a consistent TLS configuration object.

However, there are some issues to discuss first:
* The `HTTPClient` option setter overrides the current HTTP client, including any set with `DialTLSConfig`.
* Similarly, the `DialTLSConfig` option setter also overrides the current HTTP client, including any set with `HTTPClient`.
* The `DialTLSConfig` option setter creates an HTTP client with a lot of (hidden) hard-coded magic parameter values. This means that "just adding a TLS config" has side-effects over the configured HTTP client compared to setting a TLS config on a custom HTTP client and setting it via the `HTTPClient` option setter.
  * A critical side-effect of this is the `http.ProxyFromEnvironment` setting for the HTTP client, which is not in the default HTTP client, which is very simple (only configures a timeout).
* The order of precedence of which option setter "wins", depends on the order they were provided to `NewClient`.

In order to proceed, we need to decide on what to do about the above. Some questions to start:

* Is the magic HTTP client config done in `DialTLSConfig` specific to TLS connections? Should all of that be used for the default HTTP client as well? All of the magic should be documented as well, given the amount of changes done to a pristine HTTP client.
* Should the `DialTLSConfig` option setter only configure the provided TLS config in the configured HTTP client and not override it completely? Or even better, should `NewClient` do all of this instead of the setter, and the setter only storing the config?

Looking forward for your thoughts. I'm happy to implement the resulting outcome of this.

hhromic:
This PR refactors the `DialTLSConfig` option setter so the TLS config is stored in `sr.Client` and applied in `sr.NewClient` instead of the option setter.

Note that using `DialTLSConfig()` overrides any client set with `HTTPClient()`. This is the same behaviour as before.

After this PR is merged, I will add the missing `OptValues` case for the `DialTLSConfig` option to #963.

Fixes #971

### DIFF
--- a/pkg/sr/client.go
+++ b/pkg/sr/client.go
@@ -19,6 +19,7 @@ package sr
 import (
 	"bytes"
 	"context"
+	"crypto/tls"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -55,6 +56,7 @@ func (e *ResponseError) Error() string {
 type Client struct {
 	urls      []string
 	httpcl    *http.Client
+	dialTLS   *tls.Config
 	ua        string
 	defParams Param
 	opts      []ClientOpt
@@ -82,6 +84,37 @@ func NewClient(opts ...ClientOpt) (*Client, error) {
 
 	if len(cl.urls) == 0 {
 		return nil, errors.New("unable to create client with no URLs")
+	}
+
+	if cl.dialTLS != nil {
+		if cl.httpcl != nil {
+			// Sanity check some invariants: if the user is using DialTLSConfig
+			// and a custom http client, the client must be compatible with us
+			// setting the one needed field (and no others should be set).
+			if cl.httpcl.Transport != nil {
+				tr, ok := cl.httpcl.Transport.(*http.Transport)
+				if !ok {
+					return nil, errors.New("unable to use DialTLSConfig with a custom http.Client that does not use an *http.Transport")
+				}
+				if tr.Dial != nil || tr.DialContext != nil || tr.DialTLS != nil || tr.DialTLSContext != nil {
+					return nil, errors.New("unable to use DialTLSConfig with an http.Client that has a custom dial function")
+				}
+				tr = tr.Clone()
+				tr.TLSClientConfig = cl.dialTLS
+				cl.httpcl.Transport = tr
+			} else {
+				tr := http.DefaultTransport.(*http.Transport).Clone()
+				tr.TLSClientConfig = cl.dialTLS
+				cl.httpcl.Transport = tr
+			}
+		} else {
+			tr := http.DefaultTransport.(*http.Transport).Clone()
+			tr.TLSClientConfig = cl.dialTLS
+			cl.httpcl = &http.Client{
+				Timeout:   5 * time.Second,
+				Transport: tr,
+			}
+		}
 	}
 
 	return cl, nil
--- a/pkg/sr/client.go
+++ b/pkg/sr/client.go
@@ -86,34 +86,29 @@ func NewClient(opts ...ClientOpt) (*Client, error) {
 		return nil, errors.New("unable to create client with no URLs")
 	}
 
+	if cl.httpcl == nil {
+		return nil, errors.New("unable to create client with no http client")
+	}
+
 	if cl.dialTLS != nil {
-		if cl.httpcl != nil {
-			// Sanity check some invariants: if the user is using DialTLSConfig
-			// and a custom http client, the client must be compatible with us
-			// setting the one needed field (and no others should be set).
-			if cl.httpcl.Transport != nil {
-				tr, ok := cl.httpcl.Transport.(*http.Transport)
-				if !ok {
-					return nil, errors.New("unable to use DialTLSConfig with a custom http.Client that does not use an *http.Transport")
-				}
-				if tr.Dial != nil || tr.DialContext != nil || tr.DialTLS != nil || tr.DialTLSContext != nil {
-					return nil, errors.New("unable to use DialTLSConfig with an http.Client that has a custom dial function")
-				}
-				tr = tr.Clone()
-				tr.TLSClientConfig = cl.dialTLS
-				cl.httpcl.Transport = tr
-			} else {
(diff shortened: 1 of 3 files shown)


pkg/sr/client.go
```diff
@@ -84,6 +87,29 @@ func NewClient(opts ...ClientOpt) (*Client, error) {
...
+			}).DialContext(ctx, network, host)
+		}
+
+		tr.DialTLSContext = dialFn
+
+		cl.httpcl = &http.Client{
+			Timeout:   5 * time.Second,
+			Transport: tr,
+		}
+	}
```
twmb, line 111: Rather than unconditionally overriding the client, what do you think about this?
```go
        if cl.dialTLS != nil {
                if cl.httpcl != nil {
                        // Sanity check some invariants: if the user is using DialTLSConfig
                        // and a custom http client, the client must be compatible with us
                        // setting the one needed field (and no others should be set).
                        if cl.httpcl.Transport != nil {
                                tr, ok := cl.httpcl.Transport.(*http.Transport)
                                if !ok {
                                        return nil, errors.New("unable to use DialTLSConfig with a custom http.Client that does not use an *http.Transport")
                                }
                                if tr.Dial != nil || tr.DialContext != nil || tr.DialTLS != nil || tr.DialTLSContext != nil {
                                        return nil, errors.New("unable to use DialTLSConfig with an http.Client that has a custom dial function")
                                }
                                tr = tr.Clone()
                                tr.TLSClientConfig = cl.dialTLS
                        } else {
                                tr := http.DefaultTransport.(*http.Transport).Clone()
                                tr.TLSClientConfig = cl.dialTLS
                                cl.httpcl.Transport = tr
                        }
                } else {
                        tr := http.DefaultTransport.(*http.Transport).Clone()
                        tr.TLSClientConfig = cl.dialTLS
                        cl.httpcl = &http.Client{
                                Timeout:   5 * time.Second,
                                Transport: tr,
                        }
                }
        }
```
I expect the most common reason to use a custom HTTP client is to provide your own timeouts; the chunk of code in this comment allows custom clients with DialTLSConfig (minus if you provide your own transport or dial function)

twmb:
In kgo: if DialTLSConfig and Dialer are both present, an error is returned.

In this package, the old code kept whatever option was used last: if you had DialTLSConfig before HTTPClient, the final HTTPClient would be kept -- and vise versa. The new code in this PR unconditionally keeps DialTLSConfig.

I added a comment, wondering what you think, that allows people to use a custom client while opting into DialTLSConfig.

hhromic:
> In this package, the old code kept whatever option was used last: if you had DialTLSConfig before HTTPClient, the final HTTPClient would be kept -- and vise versa. The new code in this PR unconditionally keeps DialTLSConfig.

Ah you are right, I misread the existing `NewClient` code. Indeed the optoin order determines precedence.

Thanks for the feedback! Appreciated.
I am okay with the comment you left and left a another question for you.

pkg/sr/client.go
twmb: The same code that does not opt-in if TLSClientConfig is non-nil also does not opt in if there is a custom dialer,
```go
 361 func (t *Transport) hasCustomTLSDialer() bool {
 362         return t.DialTLS != nil || t.DialTLSContext != nil
 363 }
...
 396         if !t.ForceAttemptHTTP2 && (t.TLSClientConfig != nil || t.Dial != nil || t.DialContext != nil || t.hasCustomTLSDialer()) {
 397                 // Be conservative and don't automatically enable
 398                 // http2 if they've specified a custom TLS config or
 399                 // custom dialers. Let them opt-in themselves via
 400                 // http2.ConfigureTransport so we don't surprise them
 401                 // by modifying their tls.Config. Issue 14275.
 402                 // However, if ForceAttemptHTTP2 is true, it overrides the above checks.
 403                 return
 404         }
```

AT most, I'd document that using DialTLSConfig may opt out of HTTP2.

hhromic: Ah I see. I didn't check the source code of `net/http` indeed. Good call!

Then I agree with you 100%, let's just document this and mirror the behavior in `net/http`. 👍 
I will update the PR accordingly shortly.

twmb: Lmk your path forward & resolve this comment when you're done -- I'm aiming to get through `sr` issues and PRs after the incoming kgo release.

hhromic: Hey @twmb , I added your suggested implementation now, and also added some unit tests for it.

I tried to simplify your code to avoid a bit the repetitions, but turned out harder to follow. Therefore I decided to just keep your version, aka _be clear, not clever_ 👍.

hhromic:
Looking again at the code, I realize that `Client.httpcl` should actually never be set to `nil` by `HTTPClient`.
This is currently not enforced, but if a user does it, the registry client becomes unusable anyway.

That being said, I think the check for `httpcl != nil` and its code-path could be removed and simplified.
However, in that case I would add a simple check for `httpcl != nil` to `NewClient` and give a clear error.

What do you think?

hhromic:
@twmb sorry for the noise.
I kept thinking about my previous comment and I'm convinced that checking `cl.httpcl` for `nil` is the consistent and right thing to do, similar to how `len(cl.urls)` is checked for `> 0`. That being said, I went ahead and implemented this check, simplified the code for setting the TLS config and also simplified the unit tests.

Let me know if you have additional comments.

twmb:
Looks great, ty!

PR #979, opened 2025-04-17
sr: add a logger option to schema registry client

### LINKED ISSUE #910: Schema registry: allow a hook for logging?

r-vasquez:
This includes a new logger configuration option for the SR client, it also exports a Basic logger that showcases a basic implementation of the logger interface.

Initially, I'm adding just 2 logs in the client requests. 

Fixes #910

### DIFF
--- a/pkg/sr/client.go
+++ b/pkg/sr/client.go
@@ -58,6 +58,7 @@ type Client struct {
 	ua        string
 	defParams Param
 	opts      []ClientOpt
+	logger    Logger
 
 	basicAuth *struct {
 		user string
@@ -74,6 +75,7 @@ func NewClient(opts ...ClientOpt) (*Client, error) {
 		httpcl: &http.Client{Timeout: 5 * time.Second},
 		ua:     "franz-go",
 		opts:   opts,
+		logger: new(nopLogger),
 	}
 
 	for _, opt := range opts {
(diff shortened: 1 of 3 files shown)


pkg/sr/client.go
```diff
@@ -150,12 +152,13 @@ start:
 			return fmt.Errorf("pre-request hook failed for %s %q: %w", method, reqURL, err)
 		}
 	}
-
+	cl.logger.Log(LogLevelDebug, "sending request", "method", method, "URL", reqURL, "bearer", cl.bearerToken != "", "basicAuth", cl.basicAuth != nil)
```
twmb, line 155: Wdyt about `has_bearer_token` and `has_basic_auth`
Also wdyt about switching on the log level: if info, log without has_{basic|bearer}_stuff; if debug, log with?

pkg/sr/logger.go
```diff
@@ -0,0 +1,69 @@
...
+type nopLogger struct{}
+
+func (*nopLogger) Log(LogLevel, string, ...any) {}
+
+type basicLogger struct {
+	dst   io.Writer
+	level LogLevel
+}
+
+// BasicLogger returns a logger that writes newline delimited messages to dst.
```
twmb, line 59: I think since this duplicates the kgo BasicLogger, I should be able to _use_ a kgo.BasicLogger here as well, right?

r-vasquez: I don't think so, as the `kgo.BasicLogger` uses kgo.LogLevel and here is `sr.LogLevel`.

pkg/sr/client.go
r-vasquez: I like it, I added the `Level()` method to the interface and changed the name to `has_{basic|bearer}`

r-vasquez:
Closing in favor of https://github.com/twmb/franz-go/pull/1015

PR #994, opened 2025-04-29
Kfake inclusive timequery

twmb:
Copying from #897,

This commit updates the docs for ListOffsetsAfterMilli to be explicit
that timestamps returned are inclusive. Both Redpanda and Kafka use
inclusive search conditions.

Kafka: https://github.com/apache/kafka/blob/084fcbd3275ffd22ec0e1841b54de87d986afe0c/core/src/main/scala/kafka/log/UnifiedLog.scala#L1222-L1223
Redpanda: https://github.com/redpanda-data/redpanda/blob/997dd5a76b16b78e249b85843b00d591d2f91adc/src/v/storage/log_reader.cc#L605

### DIFF
--- a/pkg/kadm/metadata.go
+++ b/pkg/kadm/metadata.go
@@ -420,7 +420,7 @@ func (cl *Client) ListCommittedOffsets(ctx context.Context, topics ...string) (L
 	return cl.listOffsets(ctx, 1, -1, topics)
 }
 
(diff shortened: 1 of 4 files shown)


rockwotj:
I fixed up the tests (and cleaned up the imports) in https://github.com/twmb/franz-go/pull/999 so I think this is obsolete now

PR #1016, opened 2025-05-14
Use gotils strconv to convert between bytes to string efficiently and safe

peczenyj:
Hello

Both `reflect.StringHeader` and `reflect.SliceHeader` are considere deprecated see https://pkg.go.dev/reflect#StringHeader

the new way, since go 1.20 is to use `unsafe` package

https://pkg.go.dev/unsafe#StringData
https://pkg.go.dev/unsafe#SliceData

to be possible be backward compatible with go 1.19 (if needed) or if you want to keep the code updated for future changes, I advice to use the library `github.com/savsgio/gotils`:

https://github.com/savsgio/gotils/tree/master/strconv

as you can see it provides the two ways to convert between slice of bytes and string via build flags.

Another option is to `copy` the same code everywhere. however you may need to keep track on future updates and the purpose of this library is to be transparent.

based on this, I made this pull request and add this external dependency as suggestion. You dont need to use.

### DIFF
--- a/generate/main.go
+++ b/generate/main.go
@@ -419,7 +419,7 @@ func validateMaxVersions(top, on *Struct) {
 	}
 }
 
-//go:generate sh -c "go run . | gofumpt | gofumpt -lang 1.19 -extra > ../pkg/kmsg/generated.go"
+//go:generate sh -c "go run . | gofumpt | gofumpt -lang go1.19 -extra > ../pkg/kmsg/generated.go"
 func main() {
 	const dir = "definitions"
 	const enums = "enums"
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.23.8
 require (
 	github.com/klauspost/compress v1.18.0
 	github.com/pierrec/lz4/v4 v4.1.22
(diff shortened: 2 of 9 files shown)


twmb:
The usage of these falls under unsafe rules (1) and (6).

https://pkg.go.dev/unsafe

unsafe.StringData, unsafe.String, and unsafe.SliceData were from my proposal:

https://github.com/golang/go/issues/53003

They could not be used in this package because of https://github.com/golang/go/issues/58554

The library you linked is doing the same thing I raise in issue 58554 -- which I created because I tried introducing unsafe.StringData and unsafe.SliceData to this package.

The original issue is likely fixed now due to https://github.com/golang/go/issues/59033, albeit they can't backport that to 1.20 and prior. Since this library requires a minimum version of 1.23 now anyway, I'd rather convert this library to the unsafe.StringData and unsafe.SliceData APIs, rather than introduce a library.

Deprecated != not working and not supported, for what it's worth. The main benefit of switching is to switch to APIs that are documented to not change.

twmb:
Let me know if you're open to switching the franz-go library itself to using unsafe directly, rather than an external library. I'll do it myself otherwise, and am leaning to close this.

PR #1022, opened 2025-05-16
EventHub Readme update

stampy88:
After upgrading to the v1.19.x, I was no longer able to connect to EventHub with the default settings. The problem stems from the fact that MS doesn't support greater than 3.8. The problem appears to be support for the v4 API which was added in v3.9. Commented this out allows v3.9 to work as well:
```
// KAFKA-17011 ede289db93f
now.incmax(18, 4) // 4 api versions
```

The following error occurs when using v3.9 and later:
```
[DEBUG] connection opened to broker; addr: XXX.servicebus.windows.net:9093, broker: seed_0
[DEBUG] issuing api versions request; broker: seed_0, version: 4
[DEBUG] wrote ApiVersions v4; broker: seed_0, bytes_written: 31, write_wait: 44.391µs, time_to_write: 21.115µs, err: <nil>
[DEBUG] read ApiVersions v4; broker: seed_0, bytes_read: 0, read_wait: 25.502µs, time_to_read: 62.984188ms, err: read tcp 192.168.0.36:34240->X.X.X.X:9093: read: connection reset by peer
[DEBUG] connection initialization failed; addr: XXX.servicebus.windows.net:9093, broker: seed_0, err: read tcp 192.168.0.36:34240->X.X.X.X:9093: read: connection reset by peer

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,15 @@ This library attempts to provide an intuitive API while interacting with Kafka t
 - **Kafka**: the original Java project
 - **Confluent Platform**
 - **Microsoft Event Hubs**
+  - As of May 16, 2025, Event Hubs supports only Kafka version 3.8. You **must** set the `kgo.MaxVersions(kversion.V3_8_0())` 
+    option; otherwise, you'll encounter the following error:
+    ```
+    [DEBUG] connection opened to broker; addr: XXX.servicebus.windows.net:9093, broker: seed_0
+    [DEBUG] issuing api versions request; broker: seed_0, version: 4
+    [DEBUG] wrote ApiVersions v4; broker: seed_0, bytes_written: 31, write_wait: 45.291µs, time_to_write: 22.107µs, err: <nil>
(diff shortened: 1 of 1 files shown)


twmb:
Another thing I can do -- since I have to create a patch anyway -- is detect cxn resets, try twice, and then downgrade to v3.

stampy88:
Oh, that would definitely work, too. Up to you.

twmb:
Yes, they're supposed to return UNSUPPORTED_VERSION, and a single api key: the min & max version of ApiVersions they support. This is called out in KIP-511 [in this section](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=125309609#KIP511:CollectandExposeClient'sNameandVersionintheBrokers-ApiVersionsRequest/ResponseHandling).

I myself was previously handling this incorrectly and thought all keys returned -- that was fixed [here](https://github.com/twmb/franz-go/commit/50aa74f17c940f676a364fc952ef92bffb0c9d8a).

twmb:
If you're open to trying a PR, please let me know if #1026 fixes this for you.

PR #1038, opened 2025-06-03
sr: add new `Client.RegisterSchema` function

hhromic:
First of all, apologies for submitting a PR with a new feature without starting a discussion about it first.
I thought this feature is simple enough and justified to just submit it for inclusion.

We recently migrated our Kafka cluster from Confluent 6.2.x to 7.8.x. In our setup, we have the ACL authorizer plugin configured in the Schema Registry service for implementing security. Until version 6.2.x, the `sr` package worked like a charm, however in version 7.8.x we hit access denied issues without any code changes.

After deeper investigation, we discovered that the new Schema Registry version does not apply ACL authorization correctly for the `GET /schemas/ids/{id}/versions` endpoint. Therefore, when calling `Client.CreateSchema`, we get 403 errors when the function tries to obtain the subject/version data for the newly created schema after calling the `POST /subjects/{subject}/versions` endpoint.

This does not happen when using a "superuser" with access to everything. Also, after configuring more relaxed ACLs for the "regular user", the call to the `GET /schemas/ids/{id}/versions` endpoint succeeds but returns empty arrays. Therefore I believe the `sr` package is not doing anything wrong. This is 100% a bug in the Confluent Schema Registry implementation of the ACL authorizer.

Because getting Confluent to fix this in the Schema Registry can be tedious and long (we still will try to get them to fix it), I thought that a faster solution would be to decouple these two calls in `Client.CreateSchemaWithIDAndVersion`. For this, I propose to create a new `Client.RegisterSchema` function (~~not 100% convinced on the name but is the best I got after several minutes thinking about it~~) that only calls the `POST /subjects/{subject}/versions` endpoint and is re-used by `Client.CreateSchemaWithIDAndVersion`.

Our use case actually only requires the assigned ID of the schema anyway, so we can skip the call to the `GET /schemas/ids/{id}/versions` for now until Confluent some day fixes this bug.

I hope this feature is well received and hopefully a new version of the `sr` package can be released with it.
Thanks in advance for your consideration and very open to your comments.

~~EDIT: Another candidate for the function name was `Client.RegisterSchema`, but I thought it may be confusing for users to differentiate between that and `Client.CreateSchema`.~~

### DIFF
--- a/pkg/sr/api.go
+++ b/pkg/sr/api.go
@@ -348,7 +348,43 @@ func (cl *Client) Schemas(ctx context.Context, subject string) ([]SubjectSchema,
 	return schemas, firstErr
 }
 
+// RegisterSchema attempts to create a schema with a fixed ID and version in the given subject, and
+// returns the globally unique identifier assigned by the registry. If the id is set to -1 or 0,
+// the registry will try to allocate a new identifier or reuse an existing one. If the version is
+// set to -1 or 0, the registry will interpret it as "latest".
+//
+// If the schema is already registered, the original ID will be returned and a new version will not be created.
+//
+// If you want to register a schema and receive more details back, you can use [CreateSchema] or
+// [CreateSchemaWithIDAndVersion] at the expense of more HTTP requests to the registry.
+//
+// This supports param [Normalize].
+func (cl *Client) RegisterSchema(ctx context.Context, subject string, s Schema, id, version int) (int, error) {
+	// POST /subjects/{subject}/versions => returns ID
+	path := pathSubjectWithVersion(subject)
+	var into struct {
+		ID int `json:"id"`
+	}
+	if id == -1 {
+		if err := cl.post(ctx, path, s, &into); err != nil {
+			return 0, err
+		}
+	} else {
+		ss := SubjectSchema{Schema: s, ID: id}
+		if version != -1 {
+			ss.Version = version
+		}
+		if err := cl.post(ctx, path, ss, &into); err != nil {
+			return 0, err
+		}
+	}
+	return into.ID, nil
+}
+
 // CreateSchema attempts to create a schema in the given subject.
+//
+// If you want to register or create a schema with a specific ID and/or version,
+// you can use [RegisterSchema] or [CreateSchemaWithIDAndVersion].
 //
 // This supports param [Normalize].
 func (cl *Client) CreateSchema(ctx context.Context, subject string, s Schema) (SubjectSchema, error) {
@@ -356,34 +392,23 @@ func (cl *Client) CreateSchema(ctx context.Context, subject string, s Schema) (S
 }
 
 // CreateSchemaWithIDAndVersion attempts to create a schema with a fixed ID and
-// version ID in the given subject. If the id is set to -1 or 0, this method is
-// equivalent to CreateSchema(). If the versionID is set to -1 or 0, it will be
-// omitted when creating the schema.
+// version in the given subject. If the id is set to -1 or 0, this method is
+// equivalent to [CreateSchema]. If the version is set to -1 or 0, the registry
+// will interpret it as "latest".
+//
+// This function first registers the schema and then looks up all details.
+// If you want to register a schema, only receive the registered ID back, and avoid further
+// HTTP requests to the registry, you can use [RegisterSchema] instead.
 //
 // This supports param [Normalize].
-func (cl *Client) CreateSchemaWithIDAndVersion(ctx context.Context, subject string, s Schema, id, versionID int) (SubjectSchema, error) {
-	// POST /subjects/{subject}/versions => returns ID
+func (cl *Client) CreateSchemaWithIDAndVersion(ctx context.Context, subject string, s Schema, id, version int) (SubjectSchema, error) {
 	// Newer SR returns the full SubjectSchema, but old does not, so we
 	// re-request to find the full information.
-	path := pathSubjectWithVersion(subject)
-	var into struct {
-		ID int `json:"id"`
+	sid, err := cl.RegisterSchema(ctx, subject, s, id, version)
+	if err != nil {
+		return SubjectSchema{}, err
 	}
-	if id == -1 {
-		if err := cl.post(ctx, path, s, &into); err != nil {
-			return SubjectSchema{}, err
-		}
-	} else {
-		ss := SubjectSchema{Schema: s, ID: id}
-		if versionID != -1 {
-			ss.Version = versionID
-		}
-		if err := cl.post(ctx, path, ss, &into); err != nil {
-			return SubjectSchema{}, err
-		}
-	}
-
-	usages, err := cl.SchemaUsagesByID(ctx, into.ID)
+	usages, err := cl.SchemaUsagesByID(ctx, sid)
 	if err != nil {
 		return SubjectSchema{}, err
 	}
@@ -392,7 +417,7 @@ func (cl *Client) CreateSchemaWithIDAndVersion(ctx context.Context, subject stri
 			return usage, nil
 		}
 	}
-	return SubjectSchema{}, fmt.Errorf("created schema under id %d, but unable to find SubjectSchema", into.ID)
+	return SubjectSchema{}, fmt.Errorf("created schema under id %d, but unable to find SubjectSchema", sid)
 }
 
 // LookupSchema checks to see if a schema is already registered and if so,
--- a/pkg/sr/client_test.go
+++ b/pkg/sr/client_test.go
@@ -188,6 +188,18 @@ func TestSchemaRegistryAPI(t *testing.T) {
 			expected: `[{"subject":"foo","version":1,"id":1,"schema":"{\"name\":\"foo\", \"type\": \"record\", \"fields\":[{\"name\":\"str\", \"type\": \"string\"}]}"}]`,
 		},
 		{
+			name:     "register schema",
+			fn:       func() (any, error) { return c.RegisterSchema(ctx, dummySchema.Subject, dummySchema.Schema, -1, -1) },
+			expected: `1`,
+		},
+		{
+			name: "register schema with ID and version",
+			fn: func() (any, error) {
+				return c.RegisterSchema(ctx, dummySchema.Subject, dummySchema.Schema, dummySchema.ID, dummySchema.Version)
+			},
+			expected: `1`,
+		},
+		{
 			name:     "create schema",
 			fn:       func() (any, error) { return c.CreateSchema(ctx, dummySchema.Subject, dummySchema.Schema) },
 			expected: `{"subject":"foo","version":1,"id":1,"schema":"{\"name\":\"foo\", \"type\": \"record\", \"fields\":[{\"name\":\"str\", \"type\": \"string\"}]}"}`,


hhromic:
We just figured out an "acceptable" (but not perfect) workaround with ACL rules to make the `sr` package work again as-is with our Schema Registry service at version 7.8.x. Turns out the principal needs both `GLOBAL_READ` and `SUBJECT_COMPATIBILITY_READ` permissions for it work in this version.

This means that this PR is not strictly necesary anymore, if you don't feel like adding it. Feel free to just close it in that case.

EDIT: For the record, the above is _not_ properly documented in the Confluent [documentation](https://docs.confluent.io/platform/current/confluent-security-plugins/schema-registry/authorization/index.html#confluentsecurityplugins-sr-operations). Not surprised :P.

pkg/sr/api.go
```diff
@@ -348,42 +348,59 @@ func (cl *Client) Schemas(ctx context.Context, subject string) ([]SubjectSchema,
 	return schemas, firstErr
 }
 
-// CreateSchema attempts to create a schema in the given subject.
+// CreateSchemaID attempts to create a schema with a fixed ID and version ID in the given subject,
+// and returns the globally unique identifier assigned by the registry. If the id is set to -1 or 0,
+// the registry will try to allocate a new identifier or reuse an existing one (see below).
+// If the versionID is set to -1 or 0, it will be omitted when creating the schema.
```
twmb: Please change `versionID` to just `version` (and change the other location as well). I missed this in the original PR.

pkg/sr/api.go
```diff
@@ -348,42 +348,59 @@ func (cl *Client) Schemas(ctx context.Context, subject string) ([]SubjectSchema,
...
-// This supports param [Normalize].
-func (cl *Client) CreateSchema(ctx context.Context, subject string, s Schema) (SubjectSchema, error) {
-	return cl.CreateSchemaWithIDAndVersion(ctx, subject, s, -1, -1)
-}
-
-// CreateSchemaWithIDAndVersion attempts to create a schema with a fixed ID and
-// version ID in the given subject. If the id is set to -1 or 0, this method is
-// equivalent to CreateSchema(). If the versionID is set to -1 or 0, it will be
-// omitted when creating the schema.
+// If the same schema is registered under a different subject, the same identifier will be returned.
```
twmb: `the same identifier will be returned` -> `the original identifier will be returned` does this make more sense? These latter three lines could perhaps be a bit clearer. Also, isn't this line identical to the next?

```
If the schema is already registered, the original ID will be returned and a new version will not be created.
```

Does that cover it all?

Also, what happens if the schema does NOT exist but the id & version are already in use?

pkg/sr/api.go
```diff
@@ -348,42 +348,59 @@ func (cl *Client) Schemas(ctx context.Context, subject string) ([]SubjectSchema,
 	return schemas, firstErr
 }
 
-// CreateSchema attempts to create a schema in the given subject.
+// CreateSchemaID attempts to create a schema with a fixed ID and version ID in the given subject,
+// and returns the globally unique identifier assigned by the registry. If the id is set to -1 or 0,
```
twmb: `If the id is <= 0, the registry...` (and change the logic in the code below where it checks -1 only to handle any negative)

Same thing for the next line: `if the version is <= 0`

I'm trying to make the doc a bit more succinct with these stray details because the functions are getting complicated

pkg/sr/api.go
```diff
@@ -348,42 +348,59 @@ func (cl *Client) Schemas(ctx context.Context, subject string) ([]SubjectSchema,
...
-	usages, err := cl.SchemaUsagesByID(ctx, into.ID)
+// CreateSchema attempts to create a schema in the given subject.
+//
+// This supports param [Normalize].
+func (cl *Client) CreateSchema(ctx context.Context, subject string, s Schema) (SubjectSchema, error) {
+	return cl.CreateSchemaWithIDAndVersion(ctx, subject, s, -1, -1)
+}
+
+// CreateSchemaWithIDAndVersion attempts to create a schema with a fixed ID and
+// version ID in the given subject. If the id is set to -1 or 0, this method is
```
twmb: Sorry to add comments to not your code but... while you're in here, can you change this to `If the id is <= 0`, change `versionID` to version, and fixup the next line as well? Also, can you change `CreateSchema()` to `[CreateSchema]`?

pkg/sr/api.go
```diff
@@ -348,42 +348,59 @@ func (cl *Client) Schemas(ctx context.Context, subject string) ([]SubjectSchema,
...
+// equivalent to CreateSchema(). If the versionID is set to -1 or 0, it will be
+// omitted when creating the schema.
```
twmb: I would add:
```
//
// This function first registers the schema and then looks up all details.
// If you want to register a schema, only receive the registered ID back,
// and avoid another HTTP request, you can use [RegisterSchema]
```

pkg/sr/api.go
```diff
@@ -348,42 +348,59 @@ func (cl *Client) Schemas(ctx context.Context, subject string) ([]SubjectSchema,
 	return schemas, firstErr
 }
 
-// CreateSchema attempts to create a schema in the given subject.
+// CreateSchemaID attempts to create a schema with a fixed ID and version ID in the given subject,
```
twmb: Yeah I'm totally unconvinced on the name. I like your proposal of RegisterSchema. I was thinking PostSchema schema but Register makes more sense for what the return value is.

pkg/sr/api.go
```diff
@@ -348,42 +348,59 @@ func (cl *Client) Schemas(ctx context.Context, subject string) ([]SubjectSchema,
...
-	return cl.CreateSchemaWithIDAndVersion(ctx, subject, s, -1, -1)
-}
-
-// CreateSchemaWithIDAndVersion attempts to create a schema with a fixed ID and
-// version ID in the given subject. If the id is set to -1 or 0, this method is
-// equivalent to CreateSchema(). If the versionID is set to -1 or 0, it will be
-// omitted when creating the schema.
+// If the same schema is registered under a different subject, the same identifier will be returned.
+// If the schema is identical to an existing schema in the registry, a new version will not be created.
+// Instead, the registry will return the identifier of the existing schema.
```
twmb: I'd add some doc line like I mentioned below on create with id and version,

I would add:
```
//
// If you want to register a schema and receive more details back,
// you can use [CreateSchema] or [CreateSchemaWithIDAndVersion]
// at the expense of another internal HTTP request.
```

pkg/sr/api.go
```diff
@@ -348,42 +348,59 @@ func (cl *Client) Schemas(ctx context.Context, subject string) ([]SubjectSchema,
...
 		if err := cl.post(ctx, path, ss, &into); err != nil {
-			return SubjectSchema{}, err
+			return 0, err
 		}
 	}
+	return into.ID, nil
+}
 
-	usages, err := cl.SchemaUsagesByID(ctx, into.ID)
+// CreateSchema attempts to create a schema in the given subject.
```
twmb, line 384: Also add (since we're clarifying all functions)
```
// If you want to register or create a function with a specific ID and/or version,
// you can use [RegisterSchema] or [CreateSchemaWithIDAndVersion].
```

twmb (commented):
You said you worked around it so I put this on the back burner. Anyway, I'm +1 to merge but I think since there are now three functions to create a schema, I'm commenting a lot on docs here to try to make it as minimally confusing as possible.

pkg/sr/api.go
hhromic: In the [official documentation](https://docs.confluent.io/platform/current/schema-registry/develop/api.html) it is stated that:

> version (versionId) – Version of the schema to be returned. Valid values for versionId are between [1,2^31-1] or the string latest, which returns the last registered schema under the specified subject. The value -1 is equivalent to latest.

So, technically it does not handle _any negative_ number but specifically the range `[1,2^31-1]` or the value `-1`.

In my opinion, the library should not add any extra semantics on top of the SR API's own semantics, for simplicity. We could add parameter validation or let the API itself error back if a wrong value was sent.  What do you think?

pkg/sr/api.go
hhromic: No problem adding more changes to this PR that are not directly related. Happy to help.

pkg/sr/api.go
hhromic: > Also, what happens if the schema does NOT exist but the id & version are already in use?

Just tested this with our running Confluent Schema Registry 7.8.2 instance.
I tried to register a new dummy (but valid) schema under an existing schema ID and version number.
The Schema Registry server rejected the request with this error:
```
Subject mytopic-value is not in import mode
```

This seems to be part of [SR migrations](https://docs.confluent.io/platform/current/schema-registry/installation/migrate.html), which are valid use cases for this scenario.

pkg/sr/api.go
twmb: Eh I think fine as is, I forgot this was from the official handling.

pkg/sr/api.go
```diff
@@ -348,42 +348,59 @@ func (cl *Client) Schemas(ctx context.Context, subject string) ([]SubjectSchema,
...
+//
+// This supports param [Normalize].
+func (cl *Client) CreateSchema(ctx context.Context, subject string, s Schema) (SubjectSchema, error) {
+	return cl.CreateSchemaWithIDAndVersion(ctx, subject, s, -1, -1)
+}
+
+// CreateSchemaWithIDAndVersion attempts to create a schema with a fixed ID and
+// version ID in the given subject. If the id is set to -1 or 0, this method is
+// equivalent to CreateSchema(). If the versionID is set to -1 or 0, it will be
```
twmb: `CreateSchema()` => `[CreateSchema]`, for godoc.

pkg/sr/api.go
twmb: missed the push here?

pkg/sr/api.go
```diff
@@ -348,42 +348,59 @@ func (cl *Client) Schemas(ctx context.Context, subject string) ([]SubjectSchema,
...
-// version ID in the given subject. If the id is set to -1 or 0, this method is
-// equivalent to CreateSchema(). If the versionID is set to -1 or 0, it will be
-// omitted when creating the schema.
+// If the same schema is registered under a different subject, the same identifier will be returned.
+// If the schema is identical to an existing schema in the registry, a new version will not be created.
+// Instead, the registry will return the identifier of the existing schema.
 //
 // This supports param [Normalize].
-func (cl *Client) CreateSchemaWithIDAndVersion(ctx context.Context, subject string, s Schema, id, versionID int) (SubjectSchema, error) {
+func (cl *Client) CreateSchemaID(ctx context.Context, subject string, s Schema, id, versionID int) (int, error) {
```
twmb: Rename to RegisterSchema?

pkg/sr/api.go
hhromic: Apologies, I forgot to comment that I made all the suggested changes locally but have not pushed yet until I heard back from you regarding further questions I made. This because GitHub sometimes is fussy about showing comments in new changes, specially after force-pushing.

Now that you clarified all the rest, I'll be pushing today. Thanks for the feedback!

pkg/sr/api.go
hhromic: I did this locally but have not pushed yet. 👍

pkg/sr/api.go
hhromic: Done locally too, will push soon with all these changes. 👍

hhromic:
@twmb I pushed now all the changes you proposed for you to review. Hope all is good now.
I slightly reworded some sentences in the godoc comments to be more succint and simpler.
Let me know if anything else is needed 👍.

twmb:
Thank you!

twmb:
I'll release `sr` soon now that all open SR issues are closed and PRs merged

PR #1039, opened 2025-06-04
Add buffered fetch/produce bytes to prometheus metrics

cleroux:
I noticed `buffered_fetch_bytes_total` and `buffered_produce_bytes_total` were not exported by the kprom metrics plugin so I've added them in this PR.

The [kvictoria](https://github.com/twmb/franz-go/blob/master/plugin/kvictoria/kvictoria.go#L126-L128) plugin already exports these metrics.  However, I noticed these metrics are not exported by kotel and kgmetrics and I'm not sure whether they should be or not.  

In any case, I think adding these metrics to kprom would be really useful so I decided to start with that.  Thanks!


twmb:
Thanks! I'll release ...soon...

twmb:
Pushed as plugin/kprom/v1.3.0, thanks again!

PR #1048, opened 2025-06-17
kfake: create brokers with net.Listener interface

kmrgirish:
This change introduces a new option to provide custom `net.Listener` implementations to `kfake`.
By using the `Listeners(...net.Listener)` option, users can now inject their own `net.Listener` (e.g., in-memory listeners), enabling more flexible testing scenarios.
Existing port-based configuration remains supported.

### DIFF
--- a/pkg/kfake/cluster.go
+++ b/pkg/kfake/cluster.go
@@ -105,6 +105,9 @@ func NewCluster(opts ...Opt) (*Cluster, error) {
 	if len(cfg.ports) > 0 {
 		cfg.nbrokers = len(cfg.ports)
 	}
+	if len(cfg.listeners) > 0 {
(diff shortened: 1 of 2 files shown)


twmb:
What do you think about instead providing a `ListenFn(network, address string)`?

kmrgirish:
> What do you think about instead providing a ListenFn(network, address string)?

@twmb done!

twmb:
Awesome thank you!

kmrgirish:
@twmb could you please make a release for kfake? Thanks

twmb:
kfake currently has no tagged release, it is untagged, so a `go get -u` or `go get ...@<sha>` will pull in your patch

PR #1049, opened 2025-06-20
sr: expose client.Do method

r-vasquez:
This allows users to extend the client for non-standard schema registry endpoints while reusing the client's configuration.


twmb:
Thanks!

PR #1050, opened 2025-06-25
Panic fix in kadm.Lag function

man0xff:
Hello! First of all, I want to say that I really appreciate your brilliant work! 🙏

We've got a panic using Lag function and I decided to investigate the reason.

DescribeGroups makes ListGroups request if there was no groups passed explicitly. We can get errors from ListGroups and in that case this code will panic.

I'm not sure wether this code is absolutely correct because we simply skip errors and nobody will be aware about some possibly skipped groups.

### DIFF
--- a/pkg/kadm/groups.go
+++ b/pkg/kadm/groups.go
@@ -1485,7 +1485,12 @@ func (cl *Client) Lag(ctx context.Context, groups ...string) (DescribedGroupLags
 		return nil, err
 	case errors.As(err, &se) && !se.AllFailed:
 		for _, se := range se.Errs {
-			for _, g := range se.Req.(*kmsg.DescribeGroupsRequest).Groups {
+			// can be ListGroupsRequest as well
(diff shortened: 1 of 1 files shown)


twmb:
The three options are either:
* Fail entirely and return the error (similar to auth errors)
* Do what you're doing -- we can't attribute the error to any group
* Keep the error, and return it later _alongside_ any group lags that were actually described

I do the third in some places -- notably, DescribeGroups returns all failed list requests, which is causing the panic you're running into. This PR chooses the second option, whereas the third option would keep the error and return it later at all returns (unless FetchManyOffsets has an AuthError, the only other spot an error is returned today).

I doubt anybody will proceed to handle the lag _if_ an error is returned (i.e. if route 3 is chosen). Giv en that, I think this patch handles it decently.

twmb:
Pushed as pkg/kadm/v1.16.1, thanks for the fix!

twmb:
I totally forgot I also added a few new APIs that were unreleased, so this patch release is ... very marginally bigger and should be a minor.

Issue #1041, opened 2025-06-12
Max Produce requests is limit to 9, regardless the configured MaxProduceRequestsInflightPerBroker


pracucci:
We produce records with idempotency **disabled** and we configure `kgo.MaxProduceRequestsInflightPerBroker(20)`. While testing how the Kafka client behaves in case of a Produce requests, I noticed that the maximum number of in-flight Produce requests per connection is effectively limited to 9, regardless the configured `kgo.MaxProduceRequestsInflightPerBroker()`.

Based on some tests I've done, my understanding is that:
- `sink.drain()` is blocked on `sink.produce()` call ([code](https://github.com/grafana/franz-go/blob/7ebb25d1527a5782af7b620260b5ea636330beb9/pkg/kgo/sink.go#L247))
- `sink.produce()` is blocked on `sink.doSequenced()` ([code](https://github.com/grafana/franz-go/blob/7ebb25d1527a5782af7b620260b5ea636330beb9/pkg/kgo/sink.go#L464))
- `sink.doSequenced()` is blocked on `sink.seqResps.push()` ([code](https://github.com/grafana/franz-go/blob/7ebb25d1527a5782af7b620260b5ea636330beb9/pkg/kgo/sink.go#L500))
- `sink.seqResps.push()` is blocked because there are already 8 sequences pushed to the ring buffer, and the ring buffer is limited to 8. This means we have 9 in-flight Produce requests (8 tracked by the ring buffer + 1 which is the one we want to `push()`). No more drain will happen on the connection, until a response for a previous Produce request is received, and so an entry is freed from the ring buffer

pracucci:
After some more digging, I think we also need to make `ringReq` ring buffer "unlimited" to support an arbitrary number of `MaxProduceRequestsInflightPerBroker`.

pracucci:
As a quick test, I've implemented "unlimited buffers" (so they're not a ring anymore, and they're less efficient). In my test, now `MaxProduceRequestsInflightPerBroker` gets honored with 👇 

https://github.com/twmb/franz-go/compare/pracucci:unlimited-buffers

twmb:
You're right. This could be unbounded, but I'm not sure there's much benefit. I'm also not convinced the ring optimizations as they are currently implemented have much benefit anyway, though. I'm pretty sure at once time in the past I actually implemented the ring somewhere to have a fixed small size of 8, then an overflow slice if overflow was needed, but I can't find that implementation on any laptop anymore.

[...]

PR #1057, opened 2025-07-08
Use unlimited ring buffers in the Produce path, to allow in-flight Produce requests per broker to go above 9 (when configured to a higher value)

pracucci:
In this PR I propose a refactoring to ring buffers to address the issue https://github.com/twmb/franz-go/issues/1041. In details:

1. Replace type-specific implementation with generics
2. Have two ring buffer implementations, both supporting `die()`:
  - `fixedRing`: this is the fixed size ring (8 elements). `die()` was already implemented in the places where this type is used.
  - `unlimitedRing`: this is an unlimited size ring, backed by a fixed ring (fast path) used if the number of elements is < 8, and a dynamic slice if the number of elements in the ring grow above 8. This implementation is the older `ringBatchPromise`, with two additions: support for `die()` (here it's lightweight, given there's no waiting being unlimited) and the auto-shrink of the `overflow` slice if the current size is much smaller than its capacity

Fixes https://github.com/twmb/franz-go/issues/1041

**Draft** because I still have to test it in Mimir, to ensure it actually fix the in-flight Produce requests per broker limit, and doesn't introduce regressions.

### DIFF
--- a/pkg/kgo/ring.go
+++ b/pkg/kgo/ring.go
@@ -1,9 +1,14 @@
 package kgo
 
-import "sync"
+import (
+	"sync"
+)
 
-// The ring types below are fixed sized blocking MPSC ringbuffers. These
-// replace channels in a few places in this client. The *main* advantage they
+// The ring types below are based on fixed sized blocking MPSC ringbuffer.
+// One type is a fixed size ring, and the other type is an unlimited ring
+// that uses a fixed size ring if the number of elements is less than 8.
+//
+// These rings replace channels in a few places in this client. The *main* advantage they
 // provide is to allow loops that terminate.
 //
 // With channels, we always have to have a goroutine draining the channel.  We
@@ -40,11 +45,11 @@ const (
 	eight = mask7 + 1
 )
 
-type ringReq struct {
+type fixedRing[T any] struct {
 	mu sync.Mutex
 	c  *sync.Cond
 
-	elems [eight]promisedReq
+	elems [eight]T
 
 	head uint8
 	tail uint8
@@ -52,7 +57,7 @@ type ringReq struct {
 	dead bool
 }
 
-func (r *ringReq) die() {
+func (r *fixedRing[T]) die() {
 	r.mu.Lock()
 	defer r.mu.Unlock()
 
@@ -62,7 +67,7 @@ func (r *ringReq) die() {
 	}
 }
 
-func (r *ringReq) push(pr promisedReq) (first, dead bool) {
+func (r *fixedRing[T]) push(elem T) (first, dead bool) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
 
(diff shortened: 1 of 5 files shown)


pracucci:
As soon as I've tested it in Mimir, I've realised we also have to make `brokerCtx.resps` unlimited, otherwise `broker.handleReq()` is blocked on `cxn.waitResp()` and then it blocks more then 9 in-flight Produce requests to be sent. I've done the change in https://github.com/twmb/franz-go/pull/1057/commits/d19d8f9241c8e060f30a318050693248c4393b8f, but this means `fixedRing` is not used at all at this point.

If you agree with the direction of this PR, then I will remove `fixedRing` completely and rename `unlimitedRing` to simply `ring` (adjusting doc too).

pracucci:
It's running since 12h in a Mimir dev cluster and, so far, looks fine. Things I've checked:

- No higher CPU utilization (I've also looked at CPU profiles, and doesn't show up)
- No higher memory utilization (I've also looked at memory profiles, and doesn't show up)
- No panics

pkg/kgo/ring.go
```diff
@@ -40,19 +45,19 @@ const (
 	eight = mask7 + 1
 )
 
-type ringReq struct {
+type fixedRing[T any] struct {
```
twmb: I'm fine with fixed being deleted.

pkg/kgo/ring.go
```diff
@@ -101,169 +107,88 @@ func (r *ringReq) dropPeek() (next promisedReq, more, dead bool) {
...
+
+	// If the overflow slice is 8x the fixed ring size, and the overflow length
+	// is less than 1/8 of the capacity, then we do shrink the overflow slice
+	// to 2x the currently required capacity. This prevents edge cases where
+	// the overflow slice could grow indefinitely.
+	if cap(r.overflow) > 64 && len(r.overflow) < cap(r.overflow)/8 {
+		shrunk := make([]T, len(r.overflow), len(r.overflow)*2)
+		copy(shrunk, r.overflow)
+		r.overflow = shrunk
+	}
```
twmb: l178 shifts the start of overflow by one, which should eventually recycle the slice. Perhaps if the length is zero it can still be holding onto an old pointer, so, maybe it's worth nil'ing overflow if the len is 0 after shifting.

That said, I don't think this code will ever fire. The capacity is decremented by one every time we pull from overflow because we are shifting our starting position, not the end.

pkg/kgo/ring.go
pracucci: Done in https://github.com/twmb/franz-go/pull/1057/commits/f84c863fc500d95b8680220f7d4f93e547a10f44 and https://github.com/twmb/franz-go/pull/1057/commits/5839204b04294cd718afa74cd4fc373b8d949ee8

pkg/kgo/ring.go
pracucci: > That said, I don't think this code will ever fire. The capacity is decremented by one every time we pull from overflow because we are shifting our starting position, not the end.

You're right.

> l178 shifts the start of overflow by one, which should eventually recycle the slice

I think the actual recycling happens when future `append()` will re-allocate the slice memory. See the comment I've added to https://github.com/twmb/franz-go/pull/1057/commits/45707365807fcba4d74ca2b931d6ae9cf3c63764, where I've also removed this manual trimming logic. Makes sense to you?

pracucci:
Thanks @twmb. I've addressed your feedback. Can you take another look, please?

pkg/kgo/ring.go
twmb: Yeah I meant something along those lines but probably worded it poorly. Thanks!

twmb: It still may be worth adding
```
} else {
    r.overflow = nil
}
```

so that the overflow field is not hanging onto any empty but non-nil slice behind the scenes, replacing line 124 which just closes the conditional at the moment.

pracucci: Makes sense. I've done it here: https://github.com/twmb/franz-go/pull/1057/commits/1779a16a2dffe0952ff0ec8411ed97e018dcf309. WDYT?

twmb (approved):

PR #1062, opened 2025-07-15
sr/mock: add in-memory Schema Registry mock for testing

weeco:
## Summary

Adds an in-memory mock implementation of the Confluent Schema Registry REST API for unit and integration testing.

## Key Features

- **In-memory storage**: Fast tests with no external dependencies using `httptest.Server`
- **Core API coverage**: Implements schemas, subjects, config, and compatibility endpoints
- **Schema references**: Full support including circular dependency detection
- **Thread-safe**: Concurrent-safe operations with proper mutex locking
- **Testing utilities**: Seed helpers, interceptors, and reset functionality

## API Coverage

Implements the subset of Schema Registry endpoints exercised by franz-go:
- Schema operations: GET/POST for schemas and subjects
- Configuration: Global and subject-specific compatibility levels  
- Deletion: Soft and hard delete with proper reference checking
- Compatibility: Endpoint support (always returns compatible for simplicity)

## Usage

```go
reg := mock.New(mock.WithAuth("Bearer token"))
t.Cleanup(reg.Close)

reg.SeedSchema("user-value", 1, 1, userSchema)
client, _ := sr.NewClient(sr.URLs(reg.URL()))
```

This enables reliable testing of Schema Registry clients without requiring external infrastructure.

### DIFF
--- a/pkg/sr/mock/README.md
+++ b/pkg/sr/mock/README.md
@@ -1,0 +1,186 @@
+# Schema Registry Mock
+
+The `mock` package provides an in-memory, concurrency-safe implementation of the Confluent Schema Registry REST API for unit and integration testing.
+
+## Features
+
+- **In-memory storage**: Fast tests with no external dependencies
+- **Concurrency-safe**: All public methods are thread-safe
+- **HTTP test server**: Uses `httptest.Server` for realistic testing
+- **Interceptors**: Customize behavior for error testing
+
+## Quick Start
+
+```go
+import (
+    "github.com/twmb/franz-go/pkg/sr"
+    "github.com/twmb/franz-go/pkg/sr/mock"
+)
+
+func TestSchemaRegistry(t *testing.T) {
+    // Create mock registry
+    reg := mock.New(
+        mock.WithAuth("Bearer token"),
+        mock.WithGlobalCompat(sr.CompatBackward),
+    )
+    defer reg.Close()
+
+    // Seed with a schema
+    userSchema := sr.Schema{
+        Schema: `{"type":"record","name":"User","fields":[{"name":"id","type":"string"}]}`,
+        Type:   sr.TypeAvro,
+    }
+    reg.SeedSchema("user-value", 1, 1, userSchema)
+
+    // Create client and test
+    client, _ := sr.NewClient(sr.URLs(reg.URL()))
+    schema, err := client.SchemaByID(context.Background(), 1)
+    if err != nil {
+        t.Fatal(err)
+    }
+    
+    // Verify schema
+    if schema.Schema != userSchema.Schema {
+        t.Errorf("Schema mismatch")
+    }
+}
+```
+
+## Core Usage
+
+### Schema Registration
+
+```go
+// Register programmatically
+id, version, err := reg.RegisterSchema("subject", schema)
+
+// Or seed directly
+reg.SeedSchema("subject", version, id, schema)
+```
+
+### Schema References
+
+```go
+// Register base schema
+reg.SeedSchema("address-value", 1, 1, addressSchema)
+
+// Register schema with reference
+userSchema := sr.Schema{
+    Schema: `{"type":"record","name":"User","fields":[{"name":"address","type":"Address"}]}`,
+    Type:   sr.TypeAvro,
+    References: []sr.SchemaReference{
+        {Name: "Address", Subject: "address-value", Version: 1},
+    },
+}
+reg.RegisterSchema("user-value", userSchema)
+```
+
+## Interceptors
+
+Interceptors allow customizing the mock's behavior for testing error conditions. They receive HTTP requests first and can return custom responses.
+
+```go
+type Interceptor func(w http.ResponseWriter, r *http.Request) (handled bool)
+```
+
+### Example: Testing Schema Incompatibility
+
+```go
+func TestSchemaIncompatibility(t *testing.T) {
+    reg := mock.New()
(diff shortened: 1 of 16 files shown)


pkg/sr/mock/README.md
```diff
@@ -0,0 +1,186 @@
+# Schema Registry Mock
```
twmb: Package should be named `srfake` to both mirror the pattern by kfake and to avoid overlap with the extremely common package name and variable name `mock`.

pkg/sr/mock/README.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+### Managing Interceptors
+
+```go
+// Add interceptor
+reg.Intercept(interceptorFunc)
+
+// Clear all interceptors
+reg.ClearInterceptors()
```
twmb: What's the reason to clear interceptors?

pkg/sr/mock/errors.go
```diff
@@ -0,0 +1,92 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+
+	"github.com/twmb/franz-go/pkg/sr"
+)
+
+// Error constants for different error types
```
twmb: This will be godoc on ErrSubjectNotFound, right? If yes, probably should be deleted.

pkg/sr/mock/handlers.go
```diff
@@ -0,0 +1,497 @@
...
+}
+
+// handlePostSubjectVersion emulates POST /subjects/{subject}/versions to
+// register a new schema under a subject.
+func (r *Registry) handlePostSubjectVersion(w http.ResponseWriter, req *http.Request) {
+	req.Body = http.MaxBytesReader(w, req.Body, 1<<20)
+	subject := req.PathValue("subject")
+
+	var body sr.Schema
+	if err := json.NewDecoder(req.Body).Decode(&body); err != nil {
```
twmb: NewDecoder is usually not the best thing to use. IIRC, it doesn't detect if there's extra data after a valid json token.

That said, stakes are low here.

pkg/sr/mock/registry.go
```diff
@@ -0,0 +1,650 @@
...
+	}
+
+	for _, opt := range opts {
+		opt(r)
+	}
+
+	mux := http.NewServeMux()
+
+	// schema routes
+	mux.HandleFunc("GET /schemas/ids/{id}", r.handleGetSchemaByID)
```
twmb: I'm so glad Go supports variables in paths now. I was looking for the added dependency and ALAS, this is now native. yay!

pkg/sr/mock/registry.go
```diff
@@ -0,0 +1,650 @@
...
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	r.subjects = make(map[string]*subjectData)
+	r.schemasByID = make(map[int]sr.Schema)
+	r.nextID = 1
+	r.globalCompat = sr.CompatBackward
+}
+
+// getSchemaBySubjectVersionUnsafe is like getSchemaBySubjectVersion but assumes caller holds lock
```
twmb: I usually reserve unsafe for things that literally use `unsafe` and would swap Unsafe for Locked, but this works fine too.

pkg/sr/mock/registry.go
```diff
@@ -0,0 +1,650 @@
...
+					if err := r.detectCycle(refVersionData.schema.Schema, ref.Subject, ref.Version, visited, subjectStack); err != nil {
+						return err
+					}
+				}
+			}
+		}
+	}
+
+	// Remove from subject stack before returning
+	subjectStack[subject] = false
```
twmb: `delete` rather than setting false?

twmb (commented):
Only a few comments from a very fast pass; this is large and I'll be accepting this _relatively_ as is.

pkg/sr/mock/errors.go
weeco: Good point, looking into this now. 

I checked the sr package and I think we should add a constant for the various error codes that may be returned. I already added this - probably incomplete list to this mock package, but in a follow-up PR we can move a more complete version to the SR package so that clients can check for more specific errors on the client side:

```go
const (
	errCodeSubjectNotFound     = 40401
	errCodeVersionNotFound     = 40402
	errCodeSchemaNotFound      = 40403
	errCodeInvalidSchema       = 42201
	errCodeInvalidVersion      = 42202
	errCodeInvalidCompatLevel  = 42203
	errCodeBackendDataStore    = 50001
	errCodeOperationTimeout    = 50002
	errCodeForwardRequestError = 50003
)
```

weeco: I removed the error constants, removed the `Cause` property from `RegistryError` and also unexported the `RegistryError` which was really just meant to be used as an internal helper.

pkg/sr/mock/README.md
weeco: This is a handy function when using a single registry instance across multiple tests (possibly with seeded resources). You may want to apply different interceptors for specific test scenarios, and clearing interceptors allows you to reset the mock's behaviour between tests without recreating the entire registry.

pkg/sr/mock/handlers.go
weeco: With Go 1.25: The new `jsonv2` pkg with `UnmarshalRead` will make this easier I think. I added a `decodeJSONRequest` func.

Issue #1063, opened 2025-07-18
Poor batching and Produce performance


vmihailenco:
I have a somewhat busy topic (10 partitions) with approximately 20 million records per minute, and I've recently started collecting some statistics. It turns out that:

- The average batch size is 20 records or 18KB uncompressed (as reported by `messaging_kafka_produce_uncompressed_bytes`)
- kgo makes approximately 1 million produce requests per minute to Kafka
- There are much larger batches as well (20k)

What is even more surprising is that fetches are not performing any better:

- The average batch is 30 records or 5KB compressed (as reported by `messaging_kafka_fetch_compressed_bytes`)
- kgo makes approximately 160k fetch requests per minute to Kafka
- There are much larger batches as well (20k)

By adding `ProducerLinger(100 * time.Millisecond)`, I am able to reduce the number of requests by 10x and increase batch size by 10x as well, both for producing and fetching. This is still far less than I expected, given that I have `FetchMaxBytes=50MB` and `MaxBatchBytes=100MB`.

However, `ProducerLinger` does not improve performance (number of records per minute) for me because it appears that the bottleneck is in the `Produce` method. This method is fairly fast and only takes 10µs (10 microseconds), but when you need to produce 20 million records, that time adds up. It looks like I need to add more partitions just so I can produce to them fast enough.

I migrated a long time ago from segmentio/kafka-go, where `Produce` is essentially just sending records to a Go channel: `ch <- rec`. This is much cheaper than 10µs.

So my questions are:

1. Is batching working as expected, or have I misconfigured something?
2. More importantly, is it possible to optimize `Produce`? I've took a brief look and it is a fairly big method...

PS I must add that I have 5 Kafka brokers for 10 partitions which might explain why batching is not doing so well...

twmb:
Are you able to provide a reliable reproducer, hopefully minimal, for both segment's and my own client? 

This last came up in #929 and I spent some time investigating and in my own testing still see segment's client to be slower.

vmihailenco:
@twmb I don't particularly care about batching since it does not affect performance.

Do you need reproducer for 10us Produce performance?

twmb:
Closing this since our discussion is on #1064 (even though #1067 is where the repro was finally merged)

Issue #121, opened 2022-01-14
reuse single connection to publish/subscribe at any time


vtolstov:
does it possible to improve this package by adding this stuff
1. create some Conn to kafka broker
2. at any time when i'm want to publish i'm create NewProducer(c Conn, opts...) and do  Produce/ProduceSync 
3. at any time when i'm want to subscribe to topic i'm create NewConsumer(c Conn, opts...) and use it

why i want this - my use case to have in my framework kafka broker, user at any time can call subscribe to any topic, now i need to create separate client for each subscriber, if you create helper stuff i can effectively reuse the single client to receive them all.

twmb:
So if I'm reading this properly -- the thought is that you want to reuse the underlying connections between brokers? Will the connection be used in both clients at the same time, or will there be some separation between using the clients?

vtolstov:
i'm prefer to use single conn with all clients, as kafka can send records for many topics in single batch, also it supports as i understand publish to multiple topics (dont know how this works, does kafka client needs to combine batch with messages only to single topic or one batch may contain messages for multiple topics)

twmb:
The batching for producing occurs within a single client already; reusing a connection can't help batching. The client internally batches all topics / partitions into a single produce request.

Reusing connections _may_ be beneficial if you're using the clients at separate times. But, it wouldn't be sound to allow two clients to use the same connection at the same time, because there may be writer & reader synchronization issues. AFAICT the only way to use a single connection in two clients is to lock the connection around an entire request write and response read. Locking around the response read would not be so good, because Kafka replies to requests in order, but the clients do not know which order they wrote the request, so all clients would need to peek at the first bytes on the wire to see if the response coming back is the response they're looking for (if not, it may be for another client). This is actually what segment's kafka-go client does, which is part of what makes the concurrency aspect of connections a bit odd: https://github.com/segmentio/kafka-go/blob/d5d9116c820e66943605e2c599cceae11e9bb5c1/conn.go#L1335-L1371.

So, I'm not sure a pool to reuse connections in two _active_ clients at the same time is sound.

What I would like to do, though, is reuse the same connection between two clients when one client dies and then the other client starts being used. This would speed up connection initialization (so long as both clients use the same sasl).

twmb:
I've evaluated this a bit (and wrote a chunk of code for it). I'm not sure this is feasible at the moment. I think this issue can be reopened if a suitable API can be found. There are a few problems:

[...]

PR #1064, opened 2025-07-19
Poor batching performance reproducer

vmihailenco:
This is a reproducer for https://github.com/twmb/franz-go/issues/1063

With franz-go single topic single partition:

```
go run . -topic foo -compression=snappy      
99.83 MiB/s; 1046.82k records/s
104.14 MiB/s; 1091.97k records/s
99.89 MiB/s; 1047.47k records/s
100.27 MiB/s; 1051.36k records/s
102.20 MiB/s; 1071.63k records/s
102.28 MiB/s; 1072.50k records/s
103.64 MiB/s; 1086.77k records/s
99.16 MiB/s; 1039.79k records/s
100.10 MiB/s; 1049.67k records/s
99.75 MiB/s; 1045.91k records/s
101.58 MiB/s; 1065.16k records/s
```

With segmentio:

```
go run . -topic foo -compression=snappy -fast
198.36 MiB/s; 2080.00k records/s
214.58 MiB/s; 2250.00k records/s
216.31 MiB/s; 2268.20k records/s
206.17 MiB/s; 2161.80k records/s
207.90 MiB/s; 2180.00k records/s
212.67 MiB/s; 2230.00k records/s
220.30 MiB/s; 2310.00k records/s
206.95 MiB/s; 2170.00k records/s
210.76 MiB/s; 2210.00k records/s
214.58 MiB/s; 2250.00k records/s
206.95 MiB/s; 2170.00k records/s
```

I did not try particularly hard to optimize this, .e.g `make([]kafka.Message, len(batch))` could be optimized away but it is still twice as fast...


twmb:
Good push. If you're open, I'd request this change be moved into the examples/bench/compare/segment package. I tried adding to that package due to the last issue bringing up a performance delta, but I don't use segment's package and couldn't push a good comparison.

This library can't beat WriteMessages. I dug into the implementation due to this delta and, put simple, that function is really "raw" and I can't match it unless I remove a bunch of stuff that has been added over time due to feature requests and Kafka introducing KIPs that segment does not support, or does but not in this path.

WriteMessages:
* Calculates a partition for every message in one pass
* Builds batches out of these messages in another pass
* As batches are full, pushes those batches to be written immediately (as well as any final non-full batch once all messages are formatted)
* Waits for batches to be acknowledged

That's it. The only way I can tie or beat that is if I add another API call that bypasses all other logic in the produce codepaths and does something as simple.

Produce
* Grabs a global lock and calculates if too many records or bytes are buffered, for every message -- although this is a tightly scoped lock, it is a global lock that will affect users with many many cores all concurrently producing and is not present in WriteMessages
* Handles KIP-480 for every new batch -- allowing the partitioner to choose new partitions in advance (this is a very small bit of logic but is a bit of overhead in comparison)
* Does a bit of linger logic for every record

As well, the batch-creating&flushing goroutine runs concurrently with messages being buffered, whereas WriteMessages has no contention because WriteMessages is directly creating the batches before enqueuing them.

I tried deleting a bunch of code in the guts to see if locks / numbers overhead had any major impact, but the throughput did not increase at all. I strongly suspect the performance difference here can be almost entirely attributed to partition assigning & batch building being done in, well, batches. The segment client only needs to look up valid partitions once and only needs to lookup batches once. This client needs to do that for every single individual message.

I'm able to increase this library's throughput by 50% in your example by just adding a big linger (which was a bit surprising), but I can't top >160MiB/s locally against a local 3 node Kafka cluster. I'd be open to an API to produce batches of messages all at once in one go, but I'm not sure what that'd look like internally yet.

vmihailenco:
>I'd be open to an API to produce batches of messages all at once in one go, but I'm not sure what that'd look like internally yet.

I agree that this is the most promising idea and I actually assumed that `ProduceSync` already does this and my first implementation tried to used it, not kafka-go (that's why the flag is called -fast). Too bad ProduceSync is just range+Produce.

I assume such batch API will still need to do its own batching to account for partitioning so in theory:

- there will be a new ProduceBatch API
- then Produce could be changed to buffer messages using a channel and call ProduceBatch... which essentially what this PR is doing for kafka-go

That leaves:

- KIP-480 which hopefully can be put behind an opt-out flag if it is really the bottleneck
- a bit of linger logic is already done when accumulating batches in a channel so it can be safely removed

Another surprise for me was that the benchmark does not scale beyond a single connection, e.g. I can achieve 100MB/s with a single thread and a single kgo.Client, but I can't achieve 200MB/s by adding more goroutines and using the same kgo.Client. I can `go run` more benchmark processes and it will "scale" nicely so I am not bound by CPU/disk. To be fair kafka-go has the same behavior.

>I'm able to increase this library's throughput by 50% in your example by just adding a big linger 

Like I've reported in the issue, this library does 100k+ of requests/min with many small batches which can't be healthy. Even adding a small linger (100ms) reduces the number of requests by 10x, but it did not help with throughput in my case.

>I'd request this change be moved into the examples/bench/compare/segment package

Will do.

twmb:
I'd be open to switching ProduceSync to do what you expect, that makes more sense actually and no new API would be needed. The only downside (assuming all other logic can be maintained) is that it will flush partial batches that _could_ theoretically have a few more messages stuffed in.

> Produce could be changed to buffer messages using a channel and call ProduceBatch

I don't think this is the necessary part, but we'll see / it really depends on how the implementation turns out. I think KIP-480 and linger logic can both be maintained, they proved to add minimal or non-existent overhead when I outright deleted the logic that handled this stuff.

> benchmark does not scale beyond a single connection

This one I don't have an answer for, I noticed this as well. I ran 1, 2, 5, and 10 `Produce` loops concurrently. Two helped marginally, anything more actually lowered performance. I don't even know how to debug this because there are so many tiny nodes in the pprof graph and nothing really major that can be improved.

> adding a small linger (100ms) reduces the number of requests by 10x

I'm leaning to start a "poll" (open pinned issue) asking if anybody would mind switching to a default of 100ms linger and keep it open for 3-6mo.

vmihailenco:
>I don't even know how to debug this because there are so many tiny nodes in the pprof graph and nothing really major that can be improved.

I can't say if it is a good idea with Kafka, but normally Redis/Postgres/Clickhouse etc use a pool of connections so users don't have to worry about the details.

The problem with `*kgo.Client` is that it is bound by a single connection performance so. for example, I need to manually create a `*kgo.Client` for each partition. I can't use use a single kgo.Client for the whole topic with multiple partitions because writing simply does not scale / bound by a single connection performance.

That's fine. But when I create a separate kgo.Client per partition, then all this hard work you try to do to squeeze messages for different topic/partitions into a single request just won't have any chance to work.

So I essentially pay the performance penalty without actually using cross-topic/cross-partition batching...

So my plan for now is:

- Buffer records using Go channels to build batches
- Maintain a pool of kgo.Clients so I can scale producers

The problem is that now the franz-go API does not allow to check if kgo.Client is busy / can accept more messages, e.g. should I still use the kgo.Client #1 to build a larger batch or do I need to switch to a free kgo.Client #2. I guess round robin will have to do for now. But it does not look pretty...

twmb:
> normally Redis/Postgres/Clickhouse etc use a pool of connections so users don't have to worry about the details.

This is opposite of what is prescribed per Kafka clients. Kafka clients generally only use one, while opening a second (or third or fourth) depending on what you're doing. Kafka clients are super heavy weight internally and are _meant_ to do as much batching as possible and send as much as possible down a single connection. I did evaluate actually having a pool of connections in the past, but the thinking at the time was actually to _reduce_ the number of connections open to a broker: https://github.com/twmb/franz-go/issues/121#issuecomment-1055817531

Using multiple connections to _increase_ throughput would pretty quickly run into ordering semantics problems and would likely break idempotent producers.

> does not allow to check if kgo.Client is busy / can accept more messages

One thing you could do is opt into `ManualFlushing`, and then call `TryProduce` for everything. One `TryProduce` returns `ErrMaxBuffered`, you `Flush` and then switch to your second client. You only resume on the first client once Flush returns (since flush hangs until everything is flushed). You may need to work more clients into the rotation depending on how things play out; you could also cancel flushing by quitting the flush context. I'm not saying this is a good idea fwiw, just something to that may be worth trying.

In the mean time, I may have a coworker help me setting up a lot of AI stuff to see if I can have Claude do most of the rewriting ProduceRecords for me. It's a large and a bit tricky change and I'm not too keen to spend hours and hours of my spare time manually doing this for ............... a relatively niche amount of throughput 😅

vmihailenco:
> "are meant to do as much batching as possible and send as much as possible down a single connection"

I've observed this pattern with other clients as well, so there may be merit to this approach. However, the only scenario where this seems beneficial is when dealing with hundreds of topics/partitions with low message volumes and hundreds of clients. I question whether this constitutes good Kafka design in the first place.

When handling high message volumes, I need to create separate consumers per topic+partition combination to scale processing effectively. In this scenario, batching opportunities become more limited (since I'm always producing to the same topic across just a few partitions), and batching could be implemented more simply and efficiently.

> "I'm not saying this is a good idea fwiw, just something that may be worth trying."

I've ended up creating separate consumers for each topic+partition combination. This appears to be standard practice and typically doesn't require significant effort to implement.

Production is more challenging because I can't simply create separate producers per partition — I would need to handle message routing myself. This is where your suggestion of maintaining a pool of kgo.Client instances and using TryProduce could be valuable.

Anyway, thanks for your work and help. The client is really good and these limitations are more about how Kafka works, not about the client design.

twmb:
So, coming back to this: franz-go, for me, has the same performance as segment _without_ any changes. I've added to the `bench` utility to imitate the code you added to segment. If I create 10k recs _before_ calling `Produce` and then produce them all quickly, and if I do that in 5 goroutines (similar to your max-write-threads), I get the same performance.

That said, I also worked on ProduceSyncBatch which has a 50% performance increase over the above.

Results below, running against a local 3 node 4,0 KRaft cluster

```
[18:42:25]
twmb@h4x3r:~/src/twmb/franz-go/examples/bench
$ rpk topic create foo -p 3
TOPIC  STATUS
foo    OK

[18:42:37]
twmb@h4x3r:~/src/twmb/franz-go/examples/bench
$ go run . -topic foo -compression=snappy -pgoros 5 -batch-recs 10000 -psync-batch
300.41 MiB/s; 3150.00k records/s
307.08 MiB/s; 3220.00k records/s
301.36 MiB/s; 3160.00k records/s
308.04 MiB/s; 3230.00k records/s
309.94 MiB/s; 3250.00k records/s
313.76 MiB/s; 3290.00k records/s
^Csignal: interrupt

[19:29:54]
twmb@h4x3r:~/src/twmb/franz-go/examples/bench
$ go run . -topic foo -compression=snappy -pgoros 5 -batch-recs 10000  
170.95 MiB/s; 1792.52k records/s
174.24 MiB/s; 1827.01k records/s
174.54 MiB/s; 1830.19k records/s
172.67 MiB/s; 1810.53k records/s
174.02 MiB/s; 1824.77k records/s
173.56 MiB/s; 1819.88k records/s
^Csignal: interrupt

[19:30:05]
twmb@h4x3r:~/src/twmb/franz-go/examples/bench
$ cd compare/segment/

[19:30:07]
twmb@h4x3r:~/src/twmb/franz-go/examples/bench/compare/segment
$ go run . -topic foo -compression=snappy -max-write-threads 5
175.61 MiB/s; 1841.40k records/s
175.61 MiB/s; 1841.40k records/s
175.61 MiB/s; 1841.40k records/s
178.86 MiB/s; 1875.50k records/s
179.67 MiB/s; 1884.03k records/s
178.05 MiB/s; 1866.97k records/s
^Csignal: interrupt
```

twmb:
If you're open to giving my batchsync branch a try: https://github.com/twmb/franz-go/tree/batchsync

I have to add some dedicated tests for it (I've tested it, but by deleting ProduceSync and having the new API be ProduceSync; I need to add some tests for the dedicated API).

twmb:
I do recreate your original results without any batching, and adding goroutines without batching makes things worse:
```
[19:37:48]
twmb@h4x3r:~/src/twmb/franz-go/examples/bench
$ go run . -topic foo -compression=snappy -pgoros 5
85.82 MiB/s; 899.91k records/s
84.98 MiB/s; 891.07k records/s
85.60 MiB/s; 897.58k records/s
83.81 MiB/s; 878.82k records/s
^Csignal: interrupt

[19:37:54]
twmb@h4x3r:~/src/twmb/franz-go/examples/bench
$ go run . -topic foo -compression=snappy  
100.79 MiB/s; 1056.81k records/s
100.15 MiB/s; 1050.18k records/s
99.19 MiB/s; 1040.05k records/s
100.17 MiB/s; 1050.37k records/s
99.99 MiB/s; 1048.52k records/s
99.16 MiB/s; 1039.80k records/s
99.79 MiB/s; 1046.38k records/s
^Csignal: interrupt
```

vmihailenco:
Thanks for working on this! Kafka continues to be the bottleneck, so I'm still interested in improvements, though I've learned to mitigate it through:

1. Creating more partitions. Is is awkward, because I only need these partitions to be able to write/read data in time.
2. Hosting each partition on a separate broker
3. Using an intermediate buffer and calling `Produce` from a goroutine to avoid blocking the main  loop

I could switch to `ProduceSyncBatch`, but I'd need to re-implement all the rate-limiting logic that currently exists in the producer, which will take some time.

**Regarding the API design:**

The new API feels like a low-level building block that requires wrapping with a custom Producer, yet it's still tightly integrated with franz-go's internal mechanics and batching logic. This creates an awkward middle ground. It might work in practice as an interim solution, but ideally the existing `Produce` API would be refactored to address these architectural concerns:

**Client-wide buffering:** If I configure `MaxBufferedRecords(10000)`, it should actually buffer 10,000 records when the client can't keep up with load. [Currently](https://github.com/twmb/franz-go/issues/1081), the "buffer" is just a memory limit for franz-go rather than a true buffer as most developers would expect.

**Per-broker batching:** Each broker connection should have its own buffer for accumulating batches with separate timeout flushes. Currently, franz-go sends thousands of requests even with lingering enabled — it should be closer to 1-2 requests per second per broker.

**Async-first design:** The synchronous part of `Produce` should be minimal (e.g., just rate-limiting), with all heavy work happening asynchronously to avoid blocking the main produce
loop.

I've explored the source code to see if I could implement some of these ideas, but the current architecture is tightly coupled, making it difficult to know where to start untangling things. So this is mostly wishful thinking on my part—but hopefully useful feedback for future iterations.

Thanks again for your efforts on this!

twmb:
(1) and (2) can't really be replied to, since partitions are literally the unit of horizontal scaling and increasing throughput in Kafka.

For (3), that's what I added in the bench utility, and doing so shows franz-go can be as fast as segment without any changes to the library itself. I'm not 100% sure why producing a slice of messages all at once is faster than producing each message individually, my theory is below and relates to the scheduler, time, and contention.

> Regarding the API design

I mean, part of the original point was how WriteMessages was faster; ProduceSyncBatched as implemented behaves the same way. That said, I'm not too attached to the new API so will likely pull out any useful code and leave out the commit in a dead closed PR for resuscitation later if need be.

> Client-wide buffering

I'm still not sure how you expect this to behave: the linked issue basically points out there's enough blocking in the client in the stress test that the 10k limit isn't hit. Lowering the limit severely does forcefully hit the limit.

> Per-broker batching

This is the case -- there is one sink per broker. Sinks build batches and enqueue full batches to be written to the broker. `ProducerBatchMaxBytes` configures the size at which batches are cut, `BrokerMaxWriteBytes` configures the maximum size of a request that can be written to the wire, `ProducerLinger` configures how long to wait before cutting a batch, and `UniformBytesPartitioner` (least likely to be the problem` has a setting for when a new partition is picked. `sink` is an abstraction layer managing batching and handling producing, on top of `broker` which is a thin abstraction layer on raw connections.

There is definitely a bug in the current implementation of lingering that causes tons of small batches to be cut rather than waiting for the full linger period. I have that fixed locally (to be committed tomorrow).

> Async-first design

This is the case, as soon as it makes sense, IMO. `Produce` does three things: wait for space to be available, pick a partition, and then append the record to the end of the latest batch on the partition (on the sink).

Waiting for space to be available is a requirement.

You could argue that partition picking and appending could be async. The downside is that all produced messages would then need to send down a single channel to a single goroutine that is doing the partitioning logic. If Produce is highly concurrent, that'd just create contention on the one channel every `Produce` sends to in an attempt to be async. As implemented, partitioning _can_ be concurrent if you are producing to different topics from different goroutines. Contention has to exist if you produce to the same topic from separate goroutines because picking a partition is locked; partitioning has state.

You could then argue that you could buffer the record async after you've picked a partition, but that makes KIP-480 basically impossible to implement while also satisfying the guarantee that records produced in order are written in order. As well, buffering a record is pretty tiny -- it's basically just an append, and then triggering the sink "hey if you aren't draining, do so" or beginning lingering). If you _are_ lingering, well, that is better done inline rather than async because it's dependent on just how much is buffered (multiple batches? no lingering needed; one batch? if not full, then linger) -- which quickly gets back to KIP-480 logic, meaning buffering can't be async because it's needed for KIP-480 partitioning to complete.

The main benefit ProduceSyncBatch has over Produce is that batches are forcefully not eligible to be produced until all records are packed into batches and draining  -- it removes evaluating for every record "if you're not draining, do so (or begin lingering)" and removes any concurrent draining that is cutting batches early. With enough lingering, batches aren't really cut early, so this isn't a big factor (I have an uncommitted change to fix erroneous early cuts). As well, another big benefit (~33% improvement) is (I suspect) simply from creating batches of records ahead of time. High throughput `Produce` in many goroutines to the _same_ topic causes a lot of blocking.  By creating batches of records, you actually pause before producing and indirectly reduce contention in other goroutines. You can see this when running the `bench` program: if you run multiple goroutines producing to the same topic with to "batching", throughput actually decreases. If you add batching, throughput significantly increases. _But_, if you run multiple goroutines with no batching but each are producing to different topics, throughput linearly multiplies (two topics, two goroutines: double throughput; three topics, 3 goroutines: triple throughput).

> but the current architecture is tightly coupled

I'll concede on this part, for sure. The code is a pain to work with, especially now after so many years of feature additions and concurrency bug fixes / hardening.

The code has gotten much more complicated over time, and actually a big reason for a lot of the complication is because a lot of this is async and written to be as efficient as possible and handles so much behavior that other clients don't.

---

I don't claim the code is _looks_ perfect (features and debt over time!), but I do _mostly_ claim that the code is _mostly_ optimal in terms of throughput, async-ness, and features.

vmihailenco:
> (1) and (2) can't really be replied to, since partitions are literally the unit of horizontal scaling and increasing throughput in Kafka.

I don't dispute that, but the insight here is that **Kafka itself is not the bottleneck—the client is**. I can launch multiple clients and Kafka will happily accept 10x the messages to the same partition. For the client, what matters is the **number of brokers, not the number of partitions**. This is what initially confused me.

> the linked issue basically points out there's enough blocking in the client in the stress test that the 10k limit isn't hit. Lowering the limit severely does forcefully hit the limit.

Yes, and I believe blocking/contention may be a serious but somewhat orthogonal issue.

Regardless of contention levels, proper buffering must work because it needs to happen at the earliest stage before other factors interfere. The ideal flow should be:

1. Basic rate-limiting and admission policies
2. **Client-wide buffering** respecting `MaxBufferedRecords` and `MaxBufferedBytes`
3. Partitioning and broker selection
4. **Per-broker/connection buffering** respecting `ProducerBatchMaxBytes` and lingering (yes, this means the client may temporarily use more than `MaxBufferedBytes`)
5. Batch sending

From my understanding, franz-go doesn't buffer at stage (2) and only implements (4), which is overly complex because it tries to share limits across all brokers and partitions. While this sounds elegant in theory, in practice it:

- Creates hundreds of small batches per second
- Refuses to buffer data due to high contention spread across the codebase

That said, I may be overly pessimistic. Your response suggests the current batching flaw might be fixable without a complete redesign, which would be encouraging.

> it removes evaluating for every record "if you're not draining, do so (or begin lingering)" and removes any concurrent draining that is cutting batches early

Your buffering implementation is quite unusual. The "if you're not draining, do so (or begin lingering)" pattern scattered throughout the code initially struck me as inefficient. However, I recognize the code is very well optimized — perhaps this pattern performs better than it appears. But it just not necessary with a more simple buffering.

PR #1077, opened 2025-08-01
Allow topic exclusion when using regular expressions

mmatczuk:


pkg/kgo/config.go
```diff
@@ -1536,6 +1544,22 @@ func ConsumeRegex() ConsumerOpt {
 	return consumerOpt{func(cfg *cfg) { cfg.regex = true }}
 }
 
+// ConsumeExcludeTopics sets topics to exclude when using regex consumption.
+// This option only has effect when ConsumeRegex is enabled.
+//
+// Topics matching any of the provided regular expressions will be excluded from
+// consumption, even if they match patterns provided to ConsumeTopics.
+func ConsumeExcludeTopics(topics ...string) ConsumerOpt {
```
rockwotj, line 1555: Alternatively another API since kafka topics cannot have the character `!` in them you could adapt something like this in the existing pattern and interpret them as negative patterns:

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-including-and-excluding-branches

mmatczuk: Yes.

I think the way it is it's easier to understand for users, and is self documenting.

pkg/kgo/config.go
```diff
@@ -384,6 +385,13 @@ func (cfg *cfg) validate() error {
...
 			cfg.topics[re] = compiled
 		}
+		for re := range cfg.excludeTopics {
+			compiled, err := regexp.Compile(re)
+			if err != nil {
+				return fmt.Errorf("invalid regular expression %q", re)
+			}
+			cfg.excludeTopics[re] = compiled
+		}
 	}
```
twmb, line 397: Mind adding a `else if len(excludeTopics) > 0 { return errors.New("invalid use of ConsumeExcludeTopics when not using ConsumeRegex" }` (or some better error)

pkg/kgo/config.go
```diff
@@ -1536,6 +1544,22 @@ func ConsumeRegex() ConsumerOpt {
 	return consumerOpt{func(cfg *cfg) { cfg.regex = true }}
```
twmb, line 1547: Can you add a note on ConsumeRegex:

```
// ConsumeRegex sets the client to parse all topics passed to [ConsumeTopics] as
// regular expressions. You can further use [ConsumeExcludeTopics] to exclude
// topics that would match any ConsumeTopics regex.
```

(If you can think of better wording, feel free)

mmatczuk:
@twmb rebased and updated based on the feedback

twmb:
This will be released likely tomorrow

PR #1085, opened 2025-08-22
kgo: add Fetches.RecordsAll to return a Go native iterator of records in fetches

narqo:
The PR adds a new method `Fetches.RecordsApp`, to return an iterator over the records in the fetches.

The method is similar to the existing `Fetches.RecordIter` and `Fetches.EachRecord`, but it returns the `iter.Seq`, allowing the user to use the range-over-function.

### DIFF
--- a/pkg/kgo/record_and_fetch.go
+++ b/pkg/kgo/record_and_fetch.go
@@ -3,6 +3,7 @@ package kgo
 import (
 	"context"
 	"errors"
+	"iter"
 	"reflect"
 	"time"
 	"unsafe"
@@ -518,6 +519,19 @@ beforePartition:
 		i.pi++
 		i.ri = 0
 		goto beforePartition
+	}
+}
+
+// RecordsAll returns a Go native iterator that yields the records in a fetch.
+//
+// Similarly to RecordIter(), the errors should be inspected separately.
+func (fs Fetches) RecordsAll() iter.Seq[*Record] {
+	return func(yield func(*Record) bool) {
+		for iter := fs.RecordIter(); !iter.Done(); {
+			if !yield(iter.Next()) {
+				return
+			}
+		}
 	}
 }
 
--- a/pkg/kgo/group_test.go
+++ b/pkg/kgo/group_test.go
@@ -202,8 +202,7 @@ func (c *testConsumer) etl(etlsBeforeQuit int) {
 		}
 		netls++
 
-		for iter := fetches.RecordIter(); !iter.Done(); {
-			r := iter.Next()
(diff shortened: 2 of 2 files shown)


pkg/kgo/record_and_fetch.go
```diff
@@ -521,6 +522,19 @@ beforePartition:
 	}
 }
 
+// RecordsAll returns a Go native iterator that yields the records in a fetch.
+//
+// Similarly to RecordIter(), the errors should be inspected separately.
+func (fs Fetches) RecordsAll() iter.Seq[*Record] {
```
twmb, line 530: I just really wish I didn't create the `Iter` method years before `Iter` was introduced into Go :(

pkg/kgo/record_and_fetch.go
```diff
@@ -521,6 +522,19 @@ beforePartition:
 	}
 }
 
+// RecordsAll returns a Go native iterator that yields the records in a fetch.
+//
+// Similarly to RecordIter(), the errors should be inspected separately.
```
twmb: `RecordIter()` => `[RecordIter]`

pkg/kgo/record_and_fetch.go
twmb: Actually that method is named `RecordIter`. Regardless, same sentence.

pkg/kgo/record_and_fetch.go
narqo: 🆙

pkg/kgo/record_and_fetch.go
narqo: To me, it's more annoying that fetches already has the `Records` method, tbf. But there is, of course, a historical truth there.

I've noticed the naming convention in Go's own std uses `All`, `Things` and `SomethingSeq` names, with a very subtle difference in the meaning, e.g. 

* `maps.All`, `maps.Keys`
* `bytes.SplitSeq`

I've picked `RecordsAll` because this matches with the "current" [Go docs][1]

> The iterator method on a collection type is conventionally named All, because it iterates a sequence of all the values in the collection. 

[1]: https://pkg.go.dev/iter#hdr-Naming_Conventions

twmb: Ah, good justification, +1

twmb: It may be worth adding a reference to this method in the docs on Iter, as well (basically, reference in both directions):

```
// You can also use [RecordsAll] for a native Go iterator.
```
or something.

twmb:
This will be released likely tomorrow

PR #1086, opened 2025-08-25
Bumps minimum supported Go version to 1.24.0

### LINKED ISSUE #1075: Consider making `go` directive less restrictive
Hi there! I am one of the maintainers of the [OpenTelemetry Collector](https://opentelemetry.io/docs/collector/), a Go based application and set of libraries that depend on franz-go.

When used as a library, we want our users to be able to use a Go compiler from any of the last two releases (so, as of the writing of this message, any compiler version within 1.23.x and 1.24.x is supported by us).

However, franz-go declares `1.23.8` as the minimum version required, instead of `1.23` (any 1.23.x release, including pre-releases) or `1.23.0` (any 1.23.x stable release). Would it be possible to relax this requirement back to `1.23` or `1.23.0`?

I would be happy to open a PR to do this downgrade if accepted, though one small difficulty in ensuring this is that you publish modules that depend on each other (e.g. `pkg/kadm` depends on the main module), which make it harder to do the downgrade. We can either wait until you are ready to bump to 1.24.0 or I would also be happy to instead add local replaces so that the `go mod` command uses the `go` version in the local files as opposed to the published versions of the dependencies.

Thank you for your time :)

mx-psi:
Fixes #1075

### DIFF
--- a/bump-deps.sh
+++ b/bump-deps.sh
@@ -5,21 +5,21 @@
 
 set -euo pipefail
 
-minlang="1.23.8"
-maxlang="1.24.2"
+minlang="1.24.0"
+maxlang="1.25.0"
 for modfile in $(find . -name 'go.mod' -print0 | xargs -0)
 do
     moddir=$(dirname "$modfile")
-    cd "$moddir"
+    pushd "$moddir" > /dev/null
     echo "$moddir"
-    filelang="$(grep "^go " go.mod | tr -d 'go \n')"
-    lang=" -go=$minlang"
+    filelang="$(go mod edit -json | jq -r .Go)"
+    lang="$minlang"
     if [[ "$filelang" > "$minlang" ]]; then
-        lang=" -go=$filelang"
+        lang="$filelang"
     fi
     if [[ $(pwd) == *"/franz-go/examples/"* ]]; then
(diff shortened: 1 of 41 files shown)


twmb:
```
  Error: generate/gen.go:631:11: printf: non-constant format string in call to (*github.com/twmb/franz-go/generate.LineWriter).Write (govet)
  		l.Write(s.Comment)
  		        ^
  Error: generate/gen.go:828:11: printf: non-constant format string in call to (*github.com/twmb/franz-go/generate.LineWriter).Write (govet)
  		l.Write(e.Comment)
  		        ^
  Error: generate/gen.go:836:12: printf: non-constant format string in call to (*github.com/twmb/franz-go/generate.LineWriter).Write (govet)
  			l.Write(v.Comment)
  			        ^
  3 issues:
  * govet: 3
  
```
I feel like this has been flagged and then not flagged on and off for years. IMO this can be ignored.

mx-psi:
They are easy enough to address so I tried to do so in f4242ef201089331cf6946fb0acaedeeb9f97470. Let me know if that works

twmb:
Looks good. I'll try to fix the bitnami container deprecation removal thing in a separate PR. Thanks!

Issue #1089, opened 2025-08-28
Example on how to backoff when an error occurs?


grobinson-grafana:
I've been writing a consumer around franz-go and been getting confused how best to handle errors. I would like to backoff between consecutive calls to `PollFetches` if no records were fetched because one or more fetches returned an error. This is supposed to prevent a tight loop from occurring where:

1. `fetches.Empty()` is always true
2. `fetches.Errors()` contains at least one error

This is what I've written so far, but I'm still not sure it is correct:

```go
fetches := c.client.PollFetches(ctx)
// If the client is closed, or the context was canceled, return the error
// as no fetches were polled. We use this instead of [kgo.IsClientClosed]
// so we can also check if the context was canceled.
if err := fetches.Err0(); err != nil {
	if errors.Is(err, kgo.ErrClientClosed) || errors.Is(err, context.Canceled) {
		return err
	}
}
// We also need to check if there are errors from the fetches. However,
// just because a fetch failed, it does not mean we fetched no records.
// It is possible to have errors for some fetches and records for others
// at the same time.
var numFetchErrs int
fetches.EachError(func(topic string, partition int32, err error) {
	level.Error(c.logger).Log("msg", "failed to fetch records", "topic", topic, "partition", partition, "err", err.Error())
	numFetchErrs++
})
// If we fetched no records, we can assume this is because all fetches
// failed. We return an error so the caller can decide whether to backoff
// between attempts.
if fetches.Empty() {
	if numFetchErrs > 0 {
		return fmt.Errorf("no records fetched, %d fetches failed", numFetchErrs)
	}
	return errors.New("no records fetched")
}
fetches.EachPartition(c.processFetchTopicPartition(ctx))
return nil
```

In addition, is it possible for `PollFetches` to return an empty fetch where:

1. `fetches.Empty()` is true
2. `fetches.Errors()` contains no errors

If so, when does this happen? And is this something that is at risk of becoming a tight (spin) loop?

grobinson-grafana:
I also got confused here:

> // PollFetches waits for fetches to be available, returning as soon as any
> // broker returns a fetch. If the context is nil, this function will return
> // immediately with any currently buffered records.

and:

> // PollRecords waits for records to be available, returning as soon as any
> // broker returns records in a fetch. If the context is nil, this function will
> // return immediately with any currently buffered records.

[...]

PR #1097, opened 2025-09-11
Add clarity on the difference between PollFetches and PollRecords

grobinson-grafana:
This pull request adds some additional clarity in the docs on the differences (and similarities) between `PollFetches` and `PollRecords`.

https://github.com/twmb/franz-go/issues/1089

### DIFF
--- a/pkg/kgo/consumer.go
+++ b/pkg/kgo/consumer.go
@@ -382,7 +382,8 @@ func NewErrFetch(err error) Fetches {
 
 // PollFetches waits for fetches to be available, returning as soon as any
 // broker returns a fetch. If the context is nil, this function will return
-// immediately with any currently buffered records.
+// immediately with any currently buffered records. It is functionally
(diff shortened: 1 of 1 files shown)


pkg/kgo/consumer.go
```diff
@@ -402,9 +403,9 @@ func (cl *Client) PollFetches(ctx context.Context) Fetches {
 	return cl.PollRecords(ctx, 0)
 }
 
-// PollRecords waits for records to be available, returning as soon as any
```
grobinson-grafana, line 405: As PollRecords actually waits for fetches, not records, I thought it would help avoid confusion to be explicit here. I got confused about the difference between polling fetches and polling records from the perspective of writing code that used kgo, and I had to go read the source code to understand they are equivalent.

twmb: eh sure, works

Issue #1095, opened 2025-09-10
Is there a reason for "request timeout min overhead"/RequestTimeoutOverhead being 1 second?


deep-creek:
In our product, we need to determine very fast, that the connection to Kafka is broken. 
E.g. if there's a network outage leading to hanging TCP connections (switch uplink broken), we need to determine that within way less then a second. 
What i try to do is, that i configure a consumer with FetchMaxWait=250ms and RequestTimeoutOverhead=100ms to detect that. Additionally i use the OnBrokerRead hook and based on fetch results, i internally flag the kafka connection to be ok or broken.

What blocks me right now,  is, that the config validation (cfg::validate) declines any RequestTimeoutOverhead values < 1 second. As connTimeouter::timeouts always applies that overhead to the configured values, it renders FetchMaxWait setting to 250ms useless. 

When using reflection to alter Client::connTimeouter::def after creating a new client (kgo.NewClient), i'll be able to get the timing as expected. 

Is there a good reason for the restriction of RequestTimeoutOverhead to be > 1 second?

twmb:
My assumption was that 1s is still already too low of a lower bound given networks often are laggy. But, this doesn't consider networking setups that are super duper good and don't traverse thousands of miles of land.

The other consideration is that the timeout overhead encompasses the full time for a broker to handle the request and respond. Notably, metadata requests aren't always the _fastest_ because the broker may have to piece together a large response.

100ms is pretty aggressive and can cause your cluster problems if you're inadvertently repeatedly killing connections just because the broker took a tiny bit longer. You'll trigger the client to reconnect and reissue requests, which will increase load on the cluster and potentially make whatever blip the broker is experiencing worse.

That said, if you're fine with these potential problems and want to PR changing the lower bound, I'm fine merging that PR.

twmb:
Open to creating a PR?

deep-creek:
sorry for the late response. Yes, i'll open up a PR.

twmb:
Closing now, your PR will be in the next release, thank you again

PR #1104, opened 2025-09-23
#1095: alter 'request timeout overhead'; set minimum to 100ms

deep-creek:
pull request for #1095 
Set lower limit for request timeout overhead to 100ms


twmb:
Actually since this can squeeze in a patch release, I have no concerns about merging now. But, this will wait a little bit to be available in a full release. Thanks again!

PR #1109, opened 2025-09-26
Adopt OffsetCommit value v4

### LINKED ISSUE #1105: Kafka 4.1.0 offset commit value decoding issue
There's an issue with decoding consumer offset messages directly from the topic on Kafka 4.1.0. I found out indirectly, running Kminion on some Kafka clusters which have recently been upgraded to 4.1.0 in testing environment.

While Kminion (v2.2.14) as built in the repo does not use the latest franz-go packages, testing locally with latest franz-go did not change anything in the outcome.

Kminion fails on decoding the offset commit value by calling franz-go's `OffsetCommitValue.ReadFrom` with error

```
response did not contain enough data to be valid
```

Digging deeper shows data like this:
CommitValue (hex) = `0004000000000021d4a8ffffffff010000019964f2d83600`

```
0004             = Version     (4)
000000000021d4a8 = Offset      (2217128)
ffffffff         = LeaderEpoch (-1)
```

Next up should be the MetaData string, whose length (`0100`) is 256 - which is the point at which decoding fails.
When just taking the remaining data after `01` (which is `0000019964f2d836`/decimal 1758335260726) this would result in a correct timestamp value.

So IMO, something in the encoding/decoding the metadata string in the offset commit record value has changed. I haven't been able to dig deeper into recent changes in Kafka in this regard though.

schustersv:
In Kafka 4.1.0, the OffsetCommit value record has been bumped to v4, This introduced an incompatibility in kmsg when decoding OffsetCommit values.

This patch brings decoding OffsetCommit values up to the new v4 spec.

Running fine on both Kafka 4.0.0 and 4.1.0.

Closes #1105

### DIFF
--- a/generate/definitions/misc
+++ b/generate/definitions/misc
@@ -233,7 +233,9 @@ OffsetCommitKey => not top level, with version field
 //
 // KAFKA-7437 commit 9f7267dd2f, proposed in KIP-320 and included in 2.1.0
 // released version 3.
-OffsetCommitValue => not top level, with version field
+//
+// KAFKA-19141 commit 199772a included in 4.1.0 released version 4.
+OffsetCommitValue => not top level, with version field, flexible v4+
   // Version is which encoding version this value is using.
   Version: int16
   // Offset is the committed offset.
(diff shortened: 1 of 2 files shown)


twmb:
+1 thank you!

Timeline for me for the next release is ideally within the month; I've working on the hardest bit first.

generate/definitions/misc
```diff
@@ -247,6 +249,8 @@ OffsetCommitValue => not top level, with version field
   // ExpireTimestamp, introduced in v1 and dropped in v2 with KIP-111,
   // is when this commit expires.
   ExpireTimestamp: int64 // v1-v1
+  // TopicId is the topic id of the committed offset
+  TopicId: uuid // tag 0
```
twmb: Can you make this `TopicID`?

schustersv: Thanks for spotting this, fixed

PR #1122, opened 2025-10-10
Support get/is_registered methods for serde

siller174:
Dear maintainer.

As a Developer, I want to check if the schema is present or not. I added methods to do it. 

I will be happy to hear your feedback and fix comments if you have any.

### DIFF
--- a/pkg/sr/serde_test.go
+++ b/pkg/sr/serde_test.go
@@ -280,7 +280,7 @@ func TestIsRegistered(t *testing.T) {
 
 	// Test unregistered ID
 	if serde.IsRegistered(1) {
-		t.Error("expected ID 1 to not be registered")
+		t.Error("expected ID 1 is not registered")
 	}
 
 	// Register ID 1 with all functions
@@ -298,8 +298,8 @@ func TestIsRegistered(t *testing.T) {
 	)
 
 	// Test registered ID without options (should return true if exists)
-	if !serde.IsRegistered(1) {
-		t.Error("expected ID 1 to be registered")
+	if serde.IsRegistered(1) {
+		t.Error("expected ID 1 is not registered")
 	}
 
 	// Test registered ID with EncodeFn option
@@ -388,8 +388,8 @@ func TestIsRegistered(t *testing.T) {
 
 	// Test registered ID without any functions (should still return true if exists)
 	serde.Register(4, testStruct1{})
-	if !serde.IsRegistered(4) {
-		t.Error("expected ID 4 to be registered (exists but no functions)")
+	if serde.IsRegistered(4) {
+		t.Error("expected ID 4 is not registered (exists but no functions)")
 	}
 
 	// Test registered ID 4 without functions, checking for EncodeFn


pkg/sr/serde.go
```diff
@@ -518,6 +518,70 @@ func (*ConfluentHeader) DecodeIndex(b []byte, maxLength int) ([]int, []byte, err
 
 type bReader struct{ b []byte }
 
+// IsIDRegistered returns true if the given schema ID is registered and has
+// a decode function available.
+func (s *Serde) IsIDRegistered(id int) bool {
```
twmb: What's the need for this, when you have `ForEncoding` and `ForDecoding` below?

pkg/sr/serde.go
```diff
@@ -518,6 +518,70 @@ func (*ConfluentHeader) DecodeIndex(b []byte, maxLength int) ([]int, []byte, err
...
+func (s *Serde) IsIDRegisteredForDecoding(id int) bool {
+	return s.IsIDRegistered(id)
+}
+
+// IsIDRegisteredForEncoding returns true if the given schema ID is registered
+// (regardless of whether it has a decode function).
+func (s *Serde) IsIDRegisteredForEncoding(id int) bool {
+	ids := s.loadIDs()
+	t := ids[id]
+	return t.exists
```
twmb: This doesn't check if encode or appendEncode are non-nil?

pkg/sr/serde.go
```diff
@@ -518,6 +518,70 @@ func (*ConfluentHeader) DecodeIndex(b []byte, maxLength int) ([]int, []byte, err
...
+}
+
+// IsTypeRegistered returns true if the given type is registered and has
+// an encode function available.
+func (s *Serde) IsTypeRegistered(v any) bool {
+	t, ok := s.loadTypes()[reflect.TypeOf(v)]
```
twmb: Since types are only used for encoding, is there any reason not to move this to checking just `IsIDRegisteredForEncoding`?

pkg/sr/serde.go
```diff
@@ -518,6 +518,70 @@ func (*ConfluentHeader) DecodeIndex(b []byte, maxLength int) ([]int, []byte, err
...
+	return ok && (t.encode != nil || t.appendEncode != nil)
+}
+
+// GetRegisteredIDs returns a slice of all registered schema IDs (for encoding).
```
twmb: Why is this specific to just encoding? Naming is inconsistent with the other getter.

pkg/sr/serde.go
```diff
@@ -518,6 +518,70 @@ func (*ConfluentHeader) DecodeIndex(b []byte, maxLength int) ([]int, []byte, err
...
+	result := make([]int, 0, len(ids))
+	for id, t := range ids {
+		if t.exists && t.decode != nil {
+			result = append(result, id)
+		}
+	}
+	return result
+}
+
+// GetRegisteredTypes returns a slice of all registered types.
```
twmb: What's the need for this? This seems like some internal niche guts

pkg/sr/serde.go
siller174: I want to check if we support schema or not before decoding, and if we do not support it, implement logic to get a new schema or something

pkg/sr/serde.go
siller174: Mainly, I need only `IsIDRegistered` and `GetRegisteredIDs`; this I implemented additionally

pkg/sr/serde.go
twmb: Given that, are the getters needed?

siller174:
Hello @twmb, thank you for your comments. I reimplemented the methods. Could you please check again?

siller174:
Also, what do you think about adding?

```
// Confluent Wire Format:
// [0]         : magic byte
// [1..4]      : schema ID (big-endian, uint32)
// [5..end]    : payload (Avro/Protobuf/JSON Schema encoded data)
const confluentKey byte = 0x00

var (
	ErrTooShort     = errors.New("message too short: need at least 5 bytes")
	ErrBadMagicByte = errors.New("bad magic byte: expected 0x00")
)

// GetSchemaID returns the schema ID from a message encoded in Confluent Wire Format.
func GetSchemaID(msg []byte) (uint32, error) {
	if len(msg) < 5 {
		return 0, ErrTooShort
	}
	if msg[0] != confluentKey {
		return 0, ErrBadMagicByte
	}
	schemaID := binary.BigEndian.Uint32(msg[1:5])

	return schemaID, nil
}
```

twmb:
> Also, what do you think about adding?

That already exists, via `(&sr.ConfluentHeader{}).DecodeID(...)`

twmb:
Would this API suffice for you? It's a bit slower since it uses does some string lookups and comparisons, but it's one API and covers all possible encoding opts:

```go
// IsRegistered returns whether the id was registered with _all_ the requested
// EncodingOpts. As an example, "IsRegistered(4, EncodeFn, DecodeFn)" returns
// true if the id 4 was registered with both an encode fn and decode fn.
func (s *Serde) IsRegistered(id int, encodingOptFns ...any) bool {
        ids := s.loadIDs()
        t, ok := ids[id]
        if !ok || !t.exists {
                return false
        }

        is := true
        for _, optFn := range encodingOptFns {
                name := namefn(optFn)
                if s, ok := optFn.(string); ok {
                        name = s
                }
                switch name {
                case namefn(EncodeFn):
                        is = is && t.encode != nil
                case namefn(AppendEncodeFn):
                        is = is && t.appendEncode != nil
                case namefn(DecodeFn):
                        is = is && t.decode != nil
                case namefn(GenerateFn):
                        is = is && t.gen != nil
                }
        }
        return is
}
```

Also, based on your use case, it sounds like `GetRegisteredIDForEncoding` and `GetRegisteredIDForDecoding` are unnecessary -- do you need these two functions?

siller174:
```
func (s *Serde) IsRegistered(id int, encodingOptFns ...any) bool {
        ids := s.loadIDs()
        t, ok := ids[id]
        if !ok || !t.exists {
                return false
        }

        is := true
        for _, optFn := range encodingOptFns {
                name := namefn(optFn)
                if s, ok := optFn.(string); ok {
                        name = s
                }
                switch name {
                case namefn(EncodeFn):
                        is = is && t.encode != nil
                case namefn(AppendEncodeFn):
                        is = is && t.appendEncode != nil
                case namefn(DecodeFn):
                        is = is && t.decode != nil
                case namefn(GenerateFn):
                        is = is && t.gen != nil
                }
        }
        return is
}
```

Oh, looks hard, but you are a maintainer, feel free to use your methods.
I prefer to have two different methods to check schema is registered for encoding or decoding.

In your case I have to implement, right? 

```
schemaID := 5 

	c.serde.Register(
		int(schemaID),
		key,
		sr.EncodeFn(func(v any) ([]byte, error) {
			return avro.Marshal(avro.MustParse(schema.Schema), v)
		}),
		sr.DecodeFn(func(b []byte, v any) error {
			return avro.Unmarshal(avro.MustParse(schema.Schema), b, v)
		}))



isRegistered := c.serde.IsRegistered(5, "DecodeFn") 
```

twmb:
`isRegistered := c.serde.IsRegistered(5, sr.DecodeFn)`
`isRegistered := c.serde.IsRegistered(5, sr.EncodeFn, sr.AppendEncodeFn)`
You don't need to pass a raw string, you pass the function itself.

siller174:
```
isRegistered := c.serde.IsRegistered(5, DecodeFn)
isRegistered := c.serde.IsRegistered(5, EncodeFn, AppendEncodeFn)
You don't need to pass a raw string; you pass the function itself.
```

I don't think it's a good approach, because I'm working on a common library for several services, and I don't want to know/care about functions inside services. In my common library, I just want to be sure about the schema (registered for encoding/decoding, depending on the case) before message processing. Also, message processing inside the service, which I don't care about at all

twmb:
I'm confused -- my API proposal answers your same question. There's nothing more to it.

> I just want to be sure about the schema (registered for encoding/decoding, depending on the case) before message processing

The API proposal answers that question:
`IsRegistered(5, sr.EncodeFn, sr.AppendEncodeFn)` tells you whether ID 5 has been registered with AppendEncode has been registered. This gives you the same information as `IsIDRegisteredForEncoding`.

I'll think about the API a bit, though. I'm not a fan of my own proposal's `...any`.

One reason I like the API is symmetry with `Register`:

```go
sr.Register(5, {type}, sr.EncodeFn(...))
sr.IsRegistered(5, sr.EncodeFn)
```

siller174:
>One reason I like the API is symmetry with Register:

For me, it's ok as well. 
But please leave this as example in documentation, just for future users 

```
sr.Register(5, {type}, sr.EncodeFn(...))
sr.IsRegistered(5, sr.EncodeFn)
```

twmb:
Sorry I forgot about this, I got wrapped up in finishing the 1.20 release and the following bug patches...
How does the new implementation feel to you?

twmb:
Let me know if these "feel" like a good user experience; I recognize it was a bit off of an API proposal at first but I'm hoping the UX is better with symmetry to Register.

siller174:
@twmb hey, sorry for late response. 
I added examples and docs for the function, and now it's ok for me.

twmb:
Last thing before I merge and release -- how performant does this code need to be for you? Is it in a tight bit of logic, are you checking on every message?

siller174:
>Last thing before I merge and release -- how performant does this code need to be for you? Is it in a tight bit of logic, are you checking on every message?

Yes, I check it for every message, but for me it's fine, speed it's not essential for me

twmb:
Thanks, I'll merge. If performance proves to be an issue, I'll likely introduce some other API that is based purely around bools (i.e. closer to your original PR, but probably under one method with a struct with fields checking all functions).

siller174:
@twmb thank you!

PR #1133, opened 2025-10-16
Cl

twmb:


CHANGELOG.md
```diff
@@ -1,3 +1,93 @@
...
+This complements the existing `kfake` package that allows you to run a fake
+in-memory Kafka "cluster" for unit testing. If you did not know of either of these,
+check them out! `kfake` supports many Kafka features, but transactions are still WIP.
+All franz-go tests except transaction based tests pass against a kfake "cluster",
+so odds are, it'll work for you.
+
+## Behavior changes
+
+* This library now lingers by default for 10ms. You can switch back to 0ms
+  lingering by adding `kgo.ProducerLinger(10*time.Millisecond)` to your options
```
twmb, line 20: @hhromic yes this should be 0*time.Millisecond (or just 0). Thanks for catching, not sure where your comment is though since I'm not seeing it here.

hhromic: Apologies, I made the comment directly into the commit here: <https://github.com/twmb/franz-go/commit/d7b6fa2b75a1c72f294740845696954a081db2ab#comments>.
Should have done it in this PR indeed, GitHub's UI regarding comments in commits is not very good.

And yes, I was also thinking on "just 0" (more readable), but couldn't remember from the top of my head if Go is happy with that conversion, given that durations are just aliases for int64. I see you amended the Changelog for just 0 so I guess is fine! Cheers!

Issue #1142, opened 2025-10-20
Empty metdata response when upgrading to v1.20.0 kadm/v1.17.0


countableSet:
After upgrading to `v1.20.0` and `kadm/v1.17.0` metadata requests now return empty results. I think this might have something to do with the metadata cache changes.

I tracked it to this commit https://github.com/twmb/franz-go/pull/1121/commits/30ed0f36b7afaa9255b451af5257bb94619cd5b9 that now uses method `RequestCachedMetadata`
```go
fn := func() (*kmsg.MetadataResponse, error) {
	return cl.cl.RequestCachedMetadata(ctx, req, 0)
}
```

In the case of no topics provided to metadata request
```go
if noTopics {
	req.Topics = []kmsg.MetadataRequestTopic{}
}
```

`fetchMappedMetadata` within the client the request metadata cache will have an empty slice of topics

<details>
<img width="868" height="518" alt="Image" src="https://github.com/user-attachments/assets/291fe7c6-1c1f-489c-93f4-5507a60875b5" />
</details>

https://github.com/twmb/franz-go/blob/5e6bc01e3872785d4506c1843c66e8816893e7c8/pkg/kgo/client.go#L2680-L2688

and `fetchCachedMappedMetadata` will return `return nil, ts`

<details>
<img width="861" height="339" alt="Image" src="https://github.com/user-attachments/assets/8e4313a6-7f0c-4fd9-8757-8660184d716d" />
</details>

https://github.com/twmb/franz-go/blob/5e6bc01e3872785d4506c1843c66e8816893e7c8/pkg/kgo/client.go#L2652-L2657

this causes `len(needed) == 0` and `intoMapped == nil`

<details>
<img width="873" height="486" alt="Image" src="https://github.com/user-attachments/assets/b30986e4-9d30-4be1-816b-5734d52f83b3" />
</details>

https://github.com/twmb/franz-go/blob/5e6bc01e3872785d4506c1843c66e8816893e7c8/pkg/kgo/client.go#L2684-L2687

Thus always returning nil metadata and not reaching the actual fetch request here.
https://github.com/twmb/franz-go/blob/5e6bc01e3872785d4506c1843c66e8816893e7c8/pkg/kgo/client.go#L2693

twmb:
I can look later today; it’s probably time for some “integration” tests in the kfake package that exercise some basic kadm functionality. lmk if you’d like to patch.


-------- Original Message --------

countableSet:
[...]

PR #1143, opened 2025-10-20
kgo: fix for empty fetch mapped metadata

countableSet:
- add check for `intoMapped` nil, fetch fetch metadata when there is nothing set in the cache 🤞🏻 
- add a test in the kfake issues test
- add gomod replace to test against main code

Hopefully the test and gomod changes are ok, otherwise I'll revert it 👍🏻 

cc https://github.com/twmb/franz-go/issues/1142


pkg/kfake/issues_test.go
```diff
@@ -549,3 +549,46 @@ func TestIssueTimestampInclusivity(t *testing.T) {
...
+	}
+	if resp.Cluster != "kfake" {
+		t.Fatalf("expected cluster kfake, got %s", resp.Cluster)
+	}
+	if resp.Controller != 0 {
+		t.Fatalf("expected controller 0, got %d", resp.Controller)
+	}
+	if len(resp.Brokers) != 1 {
+		t.Fatalf("expected 1 broker, got %d", len(resp.Brokers))
+	}
```
r-vasquez: Can we add a test for the topics returned? something like

```
	if len(resp.Topics.Names()) != 1 {
		t.Fatalf("expected exactly one topic; got: %d", len(resp.Topics.Names()))
	}
```

Right now we are seeing an empty topic list, mainly because we are passing an empty slice (instead of nil) to the metadata request.

I believe this should fix it unless there is a better way to determine when we need to fetch for all topics when the cache is empty:
```
diff --git a/pkg/kgo/client.go b/pkg/kgo/client.go
index 60e2a907..495a4430 100644
--- a/pkg/kgo/client.go
+++ b/pkg/kgo/client.go
@@ -1296,7 +1296,7 @@ func (cl *Client) Request(ctx context.Context, req kmsg.Request) (kmsg.Response,
 // set to true. This function cannot be used to request topics via TopicID;
 // the direct topic name must be used.
 func (cl *Client) RequestCachedMetadata(ctx context.Context, req *kmsg.MetadataRequest, limit time.Duration) (*kmsg.MetadataResponse, e
rror) {
-       topics := make([]string, 0, len(req.Topics))
+       var topics []string
        for _, t := range req.Topics {
                if t.Topic == nil || *t.Topic == "" {
                        return nil, errors.New("unable to request cached metadata with a missing topic name (topic IDs are not supported
)")
@@ -2690,7 +2690,7 @@ func (cl *Client) fetchMappedMetadata(ctx context.Context, topics []string, useC
                intoMapped = make(map[string]mappedMetadataTopic)
        }

-       _, _, err := cl.fetchMetadataForTopics(ctx, false, needed, intoMapped)
+       _, _, err := cl.fetchMetadataForTopics(ctx, topics == nil, needed, intoMapped)
        return intoMapped, err
 }
```

twmb: Just to be a little bit more paranoid about the logic here, wdyt of this diff?
```diff
 func (cl *Client) RequestCachedMetadata(ctx context.Context, req *kmsg.MetadataRequest, limit time.Duration) (*kmsg.MetadataResponse, error) {
-       topics := make([]string, 0, len(req.Topics))
+       var topics []string
+       if req.Topics != nil {
+               topics = make([]string, 0, len(req.Topics))
+       }
        for _, t := range req.Topics {
                if t.Topic == nil || *t.Topic == "" {
                        return nil, errors.New("unable to request cached metadata with a missing topic name (topic IDs are not supported)")
@@ -2652,7 +2660,7 @@ func (cl *Client) maybeDeleteMappedMetadata(unknownTopic bool, ts ...string) (sh
 func (cl *Client) fetchCachedMappedMetadata(limit time.Duration, ts ...string) (map[string]mappedMetadataTopic, []string) {
        cl.mappedMetaMu.Lock()
        defer cl.mappedMetaMu.Unlock()
-       if cl.mappedMeta == nil {
+       if len(cl.mappedMeta) == 0 {
                return nil, ts
        }
        cached := make(map[string]mappedMetadataTopic)
@@ -2682,7 +2690,12 @@ func (cl *Client) fetchMappedMetadata(ctx context.Context, topics []string, useC
        needed := topics
        if useCache {
                intoMapped, needed = cl.fetchCachedMappedMetadata(limit, topics...)
-               if len(needed) == 0 {
+               // If intoMapped is nil, we have no cached topics at all and
+               // need to force a metadata load to satisfy broker/controller
+               // aspects of the metadata response. We have either never
+               // issued a metadata request, or the cached data (of no topics)
+               // could be super old. Cache age is only tracked per topic.
+               if intoMapped != nil && len(needed) == 0 {
                        return intoMapped, nil
                }
        }
@@ -2690,7 +2703,7 @@ func (cl *Client) fetchMappedMetadata(ctx context.Context, topics []string, useC
                intoMapped = make(map[string]mappedMetadataTopic)
        }

-       _, _, err := cl.fetchMetadataForTopics(ctx, false, needed, intoMapped)
+       _, _, err := cl.fetchMetadataForTopics(ctx, topics == nil, needed, intoMapped)
        return intoMapped, err
 }
```

The difference is:
* First off, adds a big doc comment on why this logic exists
* Sets topics to non-nil if the request topics is non-nil _even if_ the len is 0
* Returns nil if mappedMeta len is 0 -- which covers the case for if we requested broker-only metadata 1hr ago, and want to re-request it now (i.e.: always bypass the cache if the cache is empty)

The only thing missing is that req.Version == 0 _should_ request metadata when the topics are nil _or_ empty, but I think people aren't going to be setting the request version in the standard use case of RequestCachedMetadata... so I'm fine to leave that out (which might break Kafka 0.8 and 0.9 users, who should upgrade).

twmb:
Left a comment -- good find. Mind splitting the kfake test into a different PR, and dropping the go.mod replace changes?
It took me a lot of staring to realize the issue here, I was hoping based on your initial issue that this was kadm only. I hope one day to have a franz-go release that doesn't require a patch within the first few days.

PR #1144, opened 2025-10-21
Update session timeout comment in config.go

Danoctum:
Updated the default session timeout in comments from 10s to 45s.

The `defaultCfg` in the config.go has a value of `sessionTimeout:    45000 * time.Millisecond,`. Just wanting to reflect that in the documentation.

Am in a situation where I may want to use instanceIds and the timeout seemed a bit too low for the use case.


twmb:
Thanks!

PR #1163, opened 2025-11-12
kgo: support arbitrary Kafka RPCs

rockwotj:
Sometimes it is desireable to have RPCs defined outside the set that
franz-go knows about. This patch enables that by storing API versions
in a map instead of a fixed length array.

Alternative solutions that achieve this same goal I would be open to
implementing instead (since one has to use a lot of low level APIs for
all this anyways):
* Just allow skipping API version checks for keys outside the range kgo
  supports via some configuration option
* Only skip the "brokerMin/brokerMax" checks if the issued request is
  outside the range kgo knows about. At this point it's clear the user
  is issuing a custom command, and it's sort of on them to check the
  versions...
Let me know if either option is preferred, then I'd happily update this
patch.

### DIFF
--- a/pkg/kgo/broker.go
+++ b/pkg/kgo/broker.go
@@ -188,10 +188,10 @@ func (v *brokerVersions) minVersion(key int16) int16 {
 	return -1
 }
 
-func newBrokerVersions() *brokerVersions {
+func newBrokerVersions(capacity int) *brokerVersions {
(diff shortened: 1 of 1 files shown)


rockwotj:
Force push: unexport min/maxVersions method.

twmb:
+1 this is good, I don't think there's any upside to the alternatives

PR #1165, opened 2025-11-13
kgo: Check context error before opening connection to broker

carsonip:
Use a more descriptive error if context gives an error before opening connection to broker, i.e.
`context error before opening connection to broker: context canceled` instead of `unable to dial: dial tcp: lookup kafka.broker.example: operation was canceled` to avoid being mistaken as a genuine DNS error.


carsonip:
https://github.com/twmb/franz-go/pull/1166 seems to fix the root cause better.

twmb:
Closing in favor of #1166

PR #1169, opened 2025-11-14
kgo: source.fetch stop fetching again when context is done

carsonip:
As `select` is pseudo-random when both cases are ready, check for context error in requested branch to avoid retrying immediately in next iteration in fetch loop with a failed context.

This may remove `unable to dial: dial tcp: lookup kafka.broker.example: operation was canceled` noise in the consumer code path OnBrokerConnect.


twmb:
sure, lmk if you find more problems of a similar sort

PR #1174, opened 2025-11-19
kgo: fix remaining usage of kgo.maxVers/kgo.maxVersion

### LINKED ISSUE #1172: After upgrading to 1.20.4, unable to get telemetry subscriptions
Since upgrading to 1.20.4, I get:

```
{"level":"info","err":"broker is too old; the broker has already indicated it will not know how to handle the request","time":"2025-11-17T22:27:29Z","caller":"akvorado/common/kafka/logs.go:42","module":"akvorado/common/kafka","message":"unable to get telemetry subscriptions, retrying in 30s"}
```

I suppose this could be related to 60b8d5b53fb8f40e2b94cc7a8820e6c5d7b56f9c since I don't see why any of the other commit could have introduced that.

This is with Kafka 4.1.0 as a broker.

vincentbernat:
In 60b8d5b53fb8f40e2b94cc7a8820e6c5d7b56f9c, `maxVers` was replaced by a map instead of a slice. Unknown value are now 0 instead of -1. We should use `maxVersion()`, which wraps the expected behavior.

Fix #1172.

### DIFF
--- a/pkg/kgo/broker.go
+++ b/pkg/kgo/broker.go
@@ -328,7 +328,7 @@ start:
 		return
 	}
 
-	// If v.maxVers[0] is non-negative, then we loaded API
+	// If v.maxVersion(0) is non-negative, then we loaded API
 	// versions. If the version for this request is negative, we
 	// know the broker cannot handle this request.
 	if v.maxVersion(0) >= 0 && v.maxVersion(req.Key()) < 0 {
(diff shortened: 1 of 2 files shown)


vincentbernat:
Tests are failing locally, but they were already failing before that.

twmb:
Good find, good fix, I'll release this in a patch soon (likely Friday or this weekend).

PR #1175, opened 2025-11-19
kgo: source fetch loop checks for context before fetch

carsonip:
This compliments the previous PR https://github.com/twmb/franz-go/pull/1169. Remove all pseudo-randomness in session context checks before calling fetch.

### DIFF
--- a/pkg/kgo/source.go
+++ b/pkg/kgo/source.go
@@ -782,23 +782,20 @@ func (s *source) loopFetch() {
 		case session.desireFetch() <- canFetch:
 		}
 
-		// Perform another last minute best-effort check on session context
-		// to avoid calling fetch just because it passed all the
-		// above and following selects by chance due to pseudo-random select behavior.
-		select {
-		case <-session.ctx.Done(): // single case
-			session.cancelFetchCh <- canFetch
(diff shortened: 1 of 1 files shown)


twmb:
Not ready to merge?

carsonip:
I'm not 100% convinced that this is the root cause of the context canceled errors given https://github.com/twmb/franz-go/pull/1169 with maybeFinish should break out of the fetch loop unless source is in stateContinueWorking. So for this PR to be effective, the source needs to be in stateContinueWorking and additionally it will only proceed with 50% * 50% * 50% chance.

Let me gather more data before proceeding with this PR

carsonip:
@twmb I have given it a bit more thought and let's merge this fix. It could be the first run of `s.fetch`, or previous `s.fetch` could have returned `true` in some cases. Either way there should be no harm in checking context for sure before calling `s.fetch`.

I've pushed a commit to push it further down to avoid a race, but it will always be possible for a race to happen further down anyway. Let me know if you prefer the previous select approach better and I can revert.

twmb:
Looks fine here, too. I'll merge.

twmb:
anticipated release is tomorrow

PR #1186, opened 2025-11-20
kgo: mark ErrFirstReadEOF non-retryable after 3 tries

twmb:
On very busy systems, it is possible for connections to be cut immediately after opening even if things are otherwise healthy and tls and sasl are configured properly. We now track EOF errors happening 3x in a row (on a given connection type), and only at that point mark the error as non-retryable.

This looks a bit weird with the deferred track closure, but doing it like this ensures we reset the EOF count on any success state.

Updates b2620e24f7869e8d8d2b85e9044da326cf8f1c03.


twmb:
not good

PR #1187, opened 2025-11-21
kfake: harden fetch and group assignment encoding

cjohnson-confluent:
* Ensure fetch responses always include non-nil `RecordBatches` so empty/error partitions encode as empty-but-present record sets, matching Kafka clients such as `librdkafka`.

* Fix race condition in `client_conn` response writing: the old code reused a single `buf` slice across the response loop and passed `buf[start:]` directly to goroutines for writing. This meant concurrent goroutines could read corrupted data if the main loop modified `buf` for the next response before the previous write completed. The fix copies `buf` into a goroutine-local `payload` slice before launching each write goroutine, ensuring each concurrent writer has an independent snapshot of the response data.

* Simplify `client_conn` request parsing and response framing for flexible vs non-flexible responses, and improve debug logging with request key/version.

* Guarantee group member assignments are always encoded as syntactically valid `ConsumerMemberAssignment` blobs (even when logically empty) via `assignmentOrEmpty`/`ensureMemberAssignment`, so followers and stable members can always decode assignments.

* Add `groups_assignment_test` to exercise `completeLeaderSync` and `handleSync` for leader/follower members, and `seed_topics_test` to verify `SeedTopics`-created topics work with `ProduceSync`/`Ping` against `SeedTopics`-created topics.

* Scope all changes to `pkg/kfake` and tests; no exported API or kgo client semantics are modified.

### DIFF
--- a/pkg/kfake/01_fetch.go
+++ b/pkg/kfake/01_fetch.go
@@ -129,6 +129,11 @@ func (c *Cluster) handleFetch(creq *clientReq, w *watchFetch) (kmsg.Response, er
 		sp := kmsg.NewFetchResponseTopicPartition()
 		sp.Partition = p
 		sp.ErrorCode = errCode
+		// Encode empty record sets as an empty-but-present byte slice rather
+		// than null. A nil slice would be serialized as a length of -1 (or 0
+		// in the compact-nullable form), which some clients (for example,
+		// librdkafka) interpret as an invalid MessageSet size.
+		sp.RecordBatches = make([]byte, 0)
 		st := donet(t, id, 0)
 		st.Partitions = append(st.Partitions, sp)
 		return &st.Partitions[len(st.Partitions)-1]
--- a/pkg/kfake/client_conn.go
+++ b/pkg/kfake/client_conn.go
@@ -90,7 +90,7 @@ func (cc *clientConn) read() {
 			kmsg.SkipTags(&reader)
 		}
 		if err := kreq.ReadFrom(reader.Src); err != nil {
-			cc.c.cfg.logger.Logf(LogLevelDebug, "client %s unable to parse request: %v", who, err)
+			cc.c.cfg.logger.Logf(LogLevelDebug, "client %s unable to parse request (key=%d, version=%d): %v", who, key, version, err)
 			return
 		}
 
@@ -150,21 +150,20 @@ func (cc *clientConn) write() {
 			return
 		}
 
-		// Size, corr, and empty tag section if flexible: 9 bytes max.
-		buf = append(buf[:0], 0, 0, 0, 0, 0, 0, 0, 0, 0)
+		buf = buf[:0]
+		buf = append(buf, 0, 0, 0, 0, 0, 0, 0, 0)
+		hasFlexHeader := resp.kresp.IsFlexible() && resp.kresp.Key() != 18
+		if hasFlexHeader {
+			buf = append(buf, 0) // zero tagged fields
+		}
 		buf = resp.kresp.AppendTo(buf)
 
-		start := 0
-		l := len(buf) - 4
-		if !resp.kresp.IsFlexible() || resp.kresp.Key() == 18 {
-			l--
-			start++
-		}
-		binary.BigEndian.PutUint32(buf[start:], uint32(l))
-		binary.BigEndian.PutUint32(buf[start+4:], uint32(resp.corr))
+		binary.BigEndian.PutUint32(buf[:4], uint32(len(buf)-4))
+		binary.BigEndian.PutUint32(buf[4:8], uint32(resp.corr))
 
+		payload := append([]byte(nil), buf...)
 		go func() {
-			_, err := cc.conn.Write(buf[start:])
+			_, err := cc.conn.Write(payload)
 			writeCh <- err
 		}()
 
--- a/pkg/kfake/groups.go
+++ b/pkg/kfake/groups.go
@@ -101,6 +101,11 @@ func (gs groupState) String() string {
 	}
 }
 
+var emptyMemberAssignment = func() []byte {
+	var assignment kmsg.ConsumerMemberAssignment
+	return assignment.AppendTo(nil)
+}()
+
 func (c *Cluster) coordinator(id string) *broker {
 	gen := c.coordinatorGen.Load()
 	n := hashString(fmt.Sprintf("%d", gen)+"\x00\x00"+id) % uint64(len(c.bs))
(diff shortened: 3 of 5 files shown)


pkg/kfake/client_conn.go
```diff
@@ -150,21 +150,20 @@ func (cc *clientConn) write() {
...
+		hasFlexHeader := resp.kresp.IsFlexible() && resp.kresp.Key() != 18
+		if hasFlexHeader {
+			buf = append(buf, 0) // zero tagged fields
+		}
 		buf = resp.kresp.AppendTo(buf)
 
-		start := 0
-		l := len(buf) - 4
-		if !resp.kresp.IsFlexible() || resp.kresp.Key() == 18 {
-			l--
```
twmb, line 160: I think the reason I had the code ugly in the prior version was so that I could only do two appends, which at most would be two reallocs.

Your version is way better though. The choice for my version makes no sense anyway because the buf slice is reused for future writes, and this is a fake implementation that doesn't need to care about a single alloc anyway.

pkg/kfake/client_conn.go
```diff
@@ -150,21 +150,20 @@ func (cc *clientConn) write() {
...
-		binary.BigEndian.PutUint32(buf[start:], uint32(l))
-		binary.BigEndian.PutUint32(buf[start+4:], uint32(resp.corr))
+		binary.BigEndian.PutUint32(buf[:4], uint32(len(buf)-4))
+		binary.BigEndian.PutUint32(buf[4:8], uint32(resp.corr))
 
+		payload := append([]byte(nil), buf...)
 		go func() {
-			_, err := cc.conn.Write(buf[start:])
+			_, err := cc.conn.Write(payload)
 			writeCh <- err
```
twmb, line 167: This change is actually unnecessary. The select case below waits for either the write to be done, OR for the cluster to be shut down. If the cluster is shut down, this write loop exits. In either case, there is no race on the buf data: it's used for writing, only here. Only if writing completes (causing the select case to exit/continue) do we loop back to the top of the write loop and reuse the buffer.

pkg/kfake/groups.go
```diff
@@ -759,7 +764,7 @@ func (g *group) handleSync(creq *clientReq) kmsg.Response {
 	case groupStable: // member saw join and is now finally calling sync
 		resp.ProtocolType = kmsg.StringPtr(g.protocolType)
 		resp.Protocol = kmsg.StringPtr(g.protocol)
-		resp.MemberAssignment = m.assignment
+		resp.MemberAssignment = ensureMemberAssignment(m)
```
twmb: I think the changes in this file are unnecessary. The annotation of these fields in the definition files is `bytes`, which generates code that does _not_ use `-1` for the size if the input slice is empty or null.

Compare:
```
11_join_group
67:    Metadata: bytes
139:    ProtocolMetadata: bytes

14_sync_group
26:    MemberAssignment: bytes
59:  MemberAssignment: bytes

15_describe_groups
55:      ProtocolMetadata: bytes
59:      MemberAssignment: bytes
```
to
```
00_produce
37:      Records: nullable-bytes

01_fetch
239:      RecordBatches: nullable-bytes
```

In the generated code ([here](https://github.com/twmb/franz-go/blob/b077119a67ec8bac8bdb27624b70526ddc381928/pkg/kmsg/generated.go#L15507-L15512)):
```go
		v := v.MemberAssignment
		if isFlexible {
			dst = kbin.AppendCompactBytes(dst, v)
		} else {
			dst = kbin.AppendBytes(dst, v)
		}
```
where `AppendBytes` looks like:
```go
// AppendBytes appends bytes to dst prefixed with its int32 length.
func AppendBytes(dst, b []byte) []byte {
	dst = AppendInt32(dst, int32(len(b)))
	return append(dst, b...)
}
```

whereas `AppendNullableBytes` actually uses a -1:
```
// AppendNullableBytes appends a potentially nil slice to dst prefixed with its
// int32 length or int32(-1) if nil.
func AppendNullableBytes(dst, b []byte) []byte {
	if b == nil {
		return AppendInt32(dst, -1)
	}
	return AppendBytes(dst, b)
}
```

pkg/kfake/seed_topics_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"context"
+	"testing"
+	"time"
+
+	"github.com/twmb/franz-go/pkg/kgo"
+)
+
+// TestSeedTopicsProduceSync_NoHang verifies that producing to a topic created
+// via SeedTopics succeeds without hanging in the client metadata path.
+func TestSeedTopicsProduceSync_NoHang(t *testing.T) {
```
twmb, line 13: What's the purpose of this test?

pkg/kfake/01_fetch.go
```diff
@@ -129,6 +129,11 @@ func (c *Cluster) handleFetch(creq *clientReq, w *watchFetch) (kmsg.Response, er
 		sp := kmsg.NewFetchResponseTopicPartition()
 		sp.Partition = p
 		sp.ErrorCode = errCode
+		// Encode empty record sets as an empty-but-present byte slice rather
+		// than null. A nil slice would be serialized as a length of -1 (or 0
+		// in the compact-nullable form), which some clients (for example,
+		// librdkafka) interpret as an invalid MessageSet size.
+		sp.RecordBatches = make([]byte, 0)
```
twmb, line 136: I'm okay to add this here, but I'm fairly sure this if running librdkafka is failing on this, it's technically a librdkafka bug:

https://github.com/apache/kafka/blob/d04f171d3123a158a32ea65eb59600bc2f0fd628/generator/src/main/java/org/apache/kafka/message/FieldType.java#L292-L295

I do remember being surprised when implementing the serializing years ago that, technically, `records` could be nullable.

pkg/kfake/groups.go
twmb, line 767: Technically with the changes in `completeLeaderSync,` this line here is unnecessary

twmb:
> Guarantee group member assignments are always encoded as syntactically valid ConsumerMemberAssignment blobs (even when logically empty) via assignmentOrEmpty/ensureMemberAssignment, so followers and stable members can always decode assignments.

Can you explain a bit more about this one? I think the only way clients can experience problems here is if a leader is sending empty assignments for everybody else, and then everybody else is having deserialization problems.

I think that technically the patch is invalid if I use any protocol type besides `consumer`, and technically I can hijack the `consumer` protocol type and do whatever I want with it anyway with a custom implemented group balancer. I know the kafka broker itself has some special code for the `consumer` protocol type (search the Kafka repo under `group-coordinator` for any mention of `.PROTOCOL_TYPE` -- not mentioned in any KIP which is kewl). If we want to mirror some specialization into kfake to work around clients that are sending empty assignments, I think we should only do this if the protocol type == "consumer".

That said, I think this points to a client bug (although I guess the Kafka broker is more tolerant already in this case). How did you notice / where did you run into this?

twmb:
checking in @cjohnson-confluent ^^ on the comments

PR #1203, opened 2025-12-12
kgo: fix mark <=> epoch interaction, make epoch handling more resilient

twmb:
MarkCommit{Records,Offsets} compared the epoch/offset you were marking against the existing head epoch/offset. If the there was no internally existing epoch/offset, or only `dirty` was set (via being consumed) while `head` / `committed` were both the default struct value {0,0}, then trying to mark a record with a negative epoch would be ignored.

Note that returning -1 via the broker epoch requires a broker to both (a) SUPPORT epochs, i.e. implement all the Kafka APIs with leader epoch support, and then (b) NOT SUPPORT epochs, i.e. return / use -1 everywhere. This has only been seen against Azure Event hubs.

Anyway, now, when initializing an EpochOffset internally (for an uncommit {dirty,head,committed}, the epoch is explicitly initialized with -1. Further, for added robustness, MarkCommit{Records,Offsets} only compares against existing values -- if a value does not exist, we auto-accept the mark.

This commit also:
* Improves EpochOffset.Less to assume all negative epochs are -1
* Simplifies the logic in CommitRecords to use EpochOffset.Less
* Adds debug lines in metadata when new partitions are added
* Adds the epoch to an existing debug log line when updating uncommitted
* Sets the LeaderEpoch to -1 for records when producing, and clarifies in docs that -1 should have always been the case.

### DIFF
--- a/pkg/kgo/consumer_group.go
+++ b/pkg/kgo/consumer_group.go
@@ -1878,7 +1878,8 @@ type EpochOffset struct {
 // than the other if this one's epoch is less, or the epoch's are equal and
 // this one's offset is less.
 func (e EpochOffset) Less(o EpochOffset) bool {
-	return e.Epoch < o.Epoch || e.Epoch == o.Epoch && e.Offset < o.Offset
+	ee, oe := max(e.Epoch, -1), max(o.Epoch, -1)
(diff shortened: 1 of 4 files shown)


rockwotj (approved):
Nice. Makes sense

PR #1210, opened 2025-12-24
Include uncompressed/compressed message size in MessageTooLarge error

### LINKED ISSUE #1204: Include the actual record size while returning MessageTooLarge error
When we try to produce a kafka message larger in size than the configured max buffered bytes, we see the following error message being returned - 

```
The request included a message larger than the max message size the server will accept.
```

Would it be possible to also include the size of the actual kafka message that caused this error, i.e. something like - 

```
The request included a message larger than the max message size the server will accept (1468006 > 1048576).
```

Thank you!

anubhav21sharma:
Fixes [1204](https://github.com/twmb/franz-go/issues/1204)

Error message when originated from client-side:

`MESSAGE_TOO_LARGE: The request included a message larger than the max message size the server will accept (uncompressed_bytes=100000500).`

Error message when originated from broker-side:

`MESSAGE_TOO_LARGE: The request included a message larger than the max message size the server will accept (uncompressed_bytes=100000513, compressed_bytes=5000045)`

### DIFF
--- a/pkg/kgo/sink.go
+++ b/pkg/kgo/sink.go
@@ -88,7 +88,6 @@ func (s *sink) createReq(id int64, epoch int16) (*produceRequest, *kmsg.AddParti
 		producerID:    id,
 		producerEpoch: epoch,
 
-		hasHook:    s.cl.producer.hasHookBatchWritten,
 		compressor: s.cl.cfg.compressor,
 
 		wireLength:      s.cl.baseProduceRequestLength(), // start length with no topics
@@ -1788,7 +1787,6 @@ type produceRequest struct {
 	//
 	// We use this in handleReqResp for the OnProduceHook.
 	metrics produceMetrics
-	hasHook bool
 
 	compressor Compressor
 
--- a/pkg/kgo/sink.go
+++ b/pkg/kgo/sink.go
@@ -771,6 +771,7 @@ func (s *sink) handleReqResp(br *broker, req *produceRequest, resp kmsg.Response
 				batch,
 				req.producerID,
 				req.producerEpoch,
+				tmetrics[partition],
 			)
 			if retry {
 				reqRetry.addSeqBatch(topic, tid, partition, batch)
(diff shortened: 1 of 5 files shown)


pkg/kgo/sink.go
```diff
@@ -2200,10 +2205,8 @@ func (p *produceRequest) AppendTo(dst []byte) []byte {
 		}
 
 		var tmetrics map[int32]ProduceBatchMetrics
-		if p.hasHook {
-			tmetrics = make(map[int32]ProduceBatchMetrics)
-			p.metrics[topic] = tmetrics
-		}
+		tmetrics = make(map[int32]ProduceBatchMetrics)
```
twmb, line 2206: I think `hasHook` is unnecessary at this point, worth deleting?

twmb (commented):
one comment - I think we can save a bool in the sink struct - but lgtm otherwise, let me know what you think, think you!

pkg/kgo/sink.go
anubhav21sharma: Yes, makes sense to delete it now.

anubhav21sharma:
> one comment - I think we can save a bool in the sink struct - but lgtm otherwise, let me know what you think, think you!

Fixed in f3a4de112. Thank you!

pkg/kgo/sink.go
```diff
@@ -88,7 +88,6 @@ func (s *sink) createReq(id int64, epoch int16) (*produceRequest, *kmsg.AddParti
 		producerID:    id,
 		producerEpoch: epoch,
```
twmb, line 90: Sorry even further - does `hasHookBatchWritten` even need to exist? It looks like it's only used for setting hasHook, which is now deleted

anubhav21sharma: Yes makes sense to delete this as well. I've made that change in 6c91ee49. Thank you!

twmb (approved):

anubhav21sharma:
@twmb - Can we please get this merged?

twmb:
I'm waiting for a minor for this one, and the one holdup for a minor is KIP-932 support, which is proving difficult enough to Claude&review&guide away that I'm going to have to hand code it.

twmb:
This will finally be released imminently (unless a final round of tests blocks me).

PR #1230, opened 2026-02-03
kfake: remove PartitionLeaderEpoch validation

### LINKED ISSUE #1229: kfake rejects Sarama produce requests due to strict PartitionLeaderEpoch validation
# kfake rejects Sarama produce requests due to strict PartitionLeaderEpoch validation

## Summary

kfake's produce handler rejects record batches where `PartitionLeaderEpoch != -1`, but real Kafka brokers accept any value and overwrite it with the current leader epoch. This causes kfake to be incompatible with Sarama (and potentially other third-party Kafka clients) which send `PartitionLeaderEpoch = 0`.

## Reproduction

```go
package main

import (
    "context"
    "fmt"
    "github.com/Shopify/sarama"  // or github.com/IBM/sarama
    "github.com/twmb/franz-go/pkg/kadm"
)

func main() {
    // Start kfake cluster
    cluster, _ := kfake.NewCluster()
    defer cluster.Close()

    bootstrapServer := cluster.ListenAddrs()[0]

    // Create topic with franz-go
    client, _ := kgo.NewClient(kgo.SeedBrokers(bootstrapServer))
    defer client.Close()
    admin := kadm.NewClient(client)
    admin.CreateTopics(context.Background(), 1, 1, nil, "test-topic")

    // Try to produce with Sarama
    config := sarama.NewConfig()
    config.Producer.Return.Successes = true
    config.Version = sarama.V2_8_0_0

    producer, _ := sarama.NewSyncProducer([]string{bootstrapServer}, config)
    defer producer.Close()

    _, _, err := producer.SendMessage(&sarama.ProducerMessage{
        Topic: "test-topic",
        Value: sarama.StringEncoder("test"),
    })

    fmt.Printf("Result: %v\n", err)
    // Output: Result: kafka server: Message contents does not match its CRC
}
```

**Expected:** Produce succeeds (as it does with real Kafka brokers)
**Actual:** Fails with `kafka server: Message contents does not match its CRC`

Note: The error message is misleading - there is no actual CRC mismatch. This is how Sarama displays Kafka error code 2 (`CorruptMessage`).

### kfake validation ([00_produce.go#L146-L149](https://github.com/twmb/franz-go/blob/main/pkg/kfake/00_produce.go#L146-L149))

```go
if b.PartitionLeaderEpoch != -1 {
    donep(rt, rp, kerr.CorruptMessage.Code, "Partition leader epoch must be -1.")
    continue
}
```

### What Sarama sends ([produce_set.go#L68](https://github.com/IBM/sarama/blob/main/produce_set.go#L68))

```go
batch := &RecordBatch{
    FirstTimestamp:   timestamp,
    Version:          2,
    // ... other fields
    // PartitionLeaderEpoch is NOT set - defaults to 0 (Go zero value)
}
```

### Byte-level evidence

| Client | PartitionLeaderEpoch (bytes 12-15) |
|--------|-----------------------------------|
| franz-go | `0xFFFFFFFF` (-1) |
| Sarama | `0x00000000` (0) |

## Why Real Kafka Brokers Accept This

According to the [Kafka mailing list discussion](https://www.mail-archive.com/users@kafka.apache.org/msg43008.html):

> "Setting the partitionLeaderEpoch happens on every Produce request, it was optimized in the protocol design."
>
> "The partition leader epoch field is not included in the CRC computation to avoid the need to recompute the CRC when this field is assigned for every batch that is received by the broker."

This means **real Kafka brokers overwrite `PartitionLeaderEpoch`** with the current leader epoch when processing produce requests - they don't validate the value sent by the client.

The Java Kafka client defines `NO_PARTITION_LEADER_EPOCH = -1` with documentation stating it's ["used to indicate an unknown leader epoch, which will be the case when the record set is first created by the producer"](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/record/RecordBatch.java), but this appears to be a convention rather than a strict protocol requirement.

## Proposal

Relax the validation in `00_produce.go` to match real Kafka broker behavior:

**Option A:** Remove the validation entirely (broker overwrites anyway)
```go
// Remove lines 146-149
```

**Option B:** Accept common values (0 and -1)
```go
if b.PartitionLeaderEpoch != -1 && b.PartitionLeaderEpoch != 0 {
    donep(rt, rp, kerr.CorruptMessage.Code, "Partition leader epoch must be -1 or 0.")
    continue
}
```

**Option C:** Accept any value (most compatible with real Kafka behavior)
```go
// The broker would overwrite this anyway, so don't validate
// Remove the check entirely
```

## Impact

This issue affects:
- **Sarama** (both IBM/sarama and Shopify/sarama) - confirmed with v1.38.1 through v1.46.3
- Potentially other third-party Kafka clients that don't explicitly set `PartitionLeaderEpoch = -1`

We confirmed the fix works by patching kfake locally to accept `PartitionLeaderEpoch == 0`, after which all Sarama produce tests pass (sync producer, async producer, all compression types).

## Environment

- franz-go/pkg/kfake: v0.0.0-20260202163810-5f30524ad7c5
- Sarama: v1.38.1 (also tested latest v1.46.3 - same issue)
- Go: 1.24

## References

- [Apache Kafka RecordBatch.java - NO_PARTITION_LEADER_EPOCH definition](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/record/RecordBatch.java)
- [Kafka mailing list - PartitionLeaderEpoch and CRC optimization](https://www.mail-archive.com/users@kafka.apache.org/msg43008.html)
- [IBM/sarama produce_set.go - RecordBatch creation](https://github.com/IBM/sarama/blob/main/produce_set.go#L68)
- [kfake 00_produce.go - PartitionLeaderEpoch validation](https://github.com/twmb/franz-go/blob/main/pkg/kfake/00_produce.go#L146)

aasquier:
## Summary

Real Kafka brokers overwrite the `PartitionLeaderEpoch` field with the current leader epoch when processing produce requests - they don't validate the value sent by the client. This field is deliberately excluded from CRC computation to make this overwrite efficient.

This strict validation in kfake caused it to reject produce requests from Sarama, which sends `0` instead of `-1` for this field. By removing the validation, kfake now matches real Kafka behavior and works with all clients.

Fixes #1229

## Changes

- Removed the `PartitionLeaderEpoch != -1` check in `pkg/kfake/00_produce.go`
- Added a comment explaining why the field is not validated

## Testing

All existing kfake tests pass.

### DIFF
--- a/pkg/kfake/00_produce.go
+++ b/pkg/kfake/00_produce.go
@@ -143,10 +143,10 @@ func (c *Cluster) handleProduce(creq *clientReq) (kmsg.Response, error) {
 				donep(rt, rp, kerr.CorruptMessage.Code, "Batch length mismatch.")
 				continue
 			}
-			if b.PartitionLeaderEpoch != -1 {
-				donep(rt, rp, kerr.CorruptMessage.Code, "Partition leader epoch must be -1.")
-				continue
-			}
+			// PartitionLeaderEpoch is not validated: real Kafka brokers
+			// overwrite this field with the current leader epoch. Clients
+			// send different values (franz-go sends -1, Sarama sends 0).
+
 			if b.Magic != 2 {
 				donep(rt, rp, kerr.CorruptMessage.Code, "Unsupported message format version.")
 				continue


twmb:
Thanks!

aasquier:
Thank you sir! Appreciate how responsive you are.

Issue #1231, opened 2026-02-04
Idle connection handling and `connections.max.idle.ms`


carsonip:
I've been looking at the idle connection handling in franz-go, including the code, the [default connIdleTimeout](https://github.com/search?q=repo%3Atwmb%2Ffranz-go%20connidletimeout&type=code), the [reap connection commit](https://github.com/twmb/franz-go/commit/46138f7e27fb1df4192c73d114f9ae8d4f20317d) and [an issue about disabling idle timeout](https://github.com/twmb/franz-go/issues/202).

Observations:
- The default of 20s is too aggressive, in comparison to the `connections.max.idle.ms` default in other clients, e.g. in Java it is 9m. Reaping these connections this frequent will introduce a lot of overhead on connection establishment including TCP, TLS handshakes.
  - For example, if metadata max age is 30s, i.e. metadata request is sent at 0s, 30s, 60s, and connection reaping will be run at 0s, 20s, 40s, 60s, the connection will be reaped at 60s, resulting at the metadata request at 60s to create a new connection. Whereas if connIdleTimeout >= 30s there will never be any new connections created after the initial one.
  - I understand connIdleTimeout is configurable but given the existing implementation it is a bit tricky to configure it nicely (e.g. in [OpenTelemetry configkafka](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45321)), which brings us to the next point:
- The reap connection runs every connIdleTimeout and checks for connection that has been idle for longer than connIdleTimeout. This introduces a weirdness that the connection may be idle for as long as 2x connIdleTimeout.
  - For example, if server idle connection timeout is 10m, to ensure that franz-go client always closes the connection before server, connIdleTimeout must be less than 5m. If it is set to 6m, it risks server closing it first, or server closing at the same time client is sending to the connection, during time between 10m and 12m.

The current connection reaping is designed with simplicity in mind, where it only involves 1 ticker and 1 goroutine, and that is appreciated.

The idea I have in mind:
- Add code to have connection last read / write time checked by the next read / write, and do not reuse it after "connIdleTime".
- Greatly increase connIdleTime to align with the 9m `connections.max.idle.ms` default in other clients. Potentially a risky change for users relying on previous connIdleTimeout=20s who have configured a server idle timeout < 9m.
[...]

Issue #1222, opened 2026-01-30
GroupTransactSession fetched record buffer not flushed after rebalance/fetch


tomplarge:
## Summary

We are using `GroupTransactSession` with `CooperativeStickyBalancer`. After a transaction is aborted due to a rebalance, `PollRecords` sometimes returns records starting from offsets ahead of the last committed offset. We're trying to understand whether this is expected behavior or a bug.

## Environment

- franz-go version: v1.20.5
- Go version: 1.23

## What We Observed

The following sequence of events occurred. I have annotated the full logs (attached below) with `// HERE` comments to highlight the key points:

1. Transaction successfully commits partition 458 up to offset **129752**
3. Subsequent transactions correctly re-fetch partition 458 starting from offset **129753** (correct behavior, noted twice in the logs)
4. Multiple rebalances occur during processing
5. A transaction processes records up to offset **136638** and attempts to commit
6. A rebalance causes the transaction to abort
7. Offset fetch response confirms last committed offset for partition 458 is **129753**
8. kgo logs: `transact session resetting to current committed state` and `aborted buffered records`
9. New transaction begins, `PollRecords` is called
10. **The issue:** `PollRecords` returns records starting at offset **136013** for partition 458, when the committed offset is **129753**
11. Application detects stale state: `EOS processing gap detected`The gap between the committed offset (129753) and the fetched offset (136013) is ~6,260 records.

### Full kgo Configuration

Here is the kgo configuration we use in production:

```go
func produceConsumerConfig(brokers []string, clientID, consumerGroup string, topics []string, transactionID string) []kgo.Opt {
    return []kgo.Opt{
        kgo.SeedBrokers(brokers...),
        kgo.ClientID(clientID),

        kgo.ConsumerGroup(consumerGroup),
        kgo.ConsumeTopics(topics...),

        // EOS configuration
        kgo.TransactionalID(transactionID),
        kgo.FetchIsolationLevel(kgo.ReadCommitted()),
        kgo.DisableAutoCommit(),
        kgo.RequireStableFetchOffsets(),

        // Producer settings
        kgo.StopProducerOnDataLossDetected(),
        kgo.ProducerBatchCompression(kgo.Lz4Compression()),
        kgo.RecordPartitioner(kgo.StickyKeyPartitioner(nil)),
        kgo.ProducerBatchMaxBytes(10 * 1024 * 1024),
        kgo.ProducerLinger(200 * time.Millisecond),

        // Consumer settings
[...]

PR #1245, opened 2026-02-09
kgo: do not reuse idle connection after connIdleTimeout

carsonip:
Check `connIdleTimeout` in `loadConnection` before reusing a connection, instead of solely relying on the background reap loop. If a connection has been idle for longer than `connIdleTimeout`, it is killed and a new connection is created. This gives a hard guarantee that a connection is never reused after the configured idle timeout.

Previously, connections were only cleaned up by the reap loop which runs every `connIdleTimeout`. Because a connection could become idle right after a reap tick, the actual idle time before reuse could be up to 2x `connIdleTimeout` (average 1.5x), making the setting difficult to reason about.

Other changes:
- Default `connIdleTimeout` increased from 20s to 30s, matching the old average idle time (1.5x × 20s = 30s), so existing users see similar behavior.
- Minimum allowed `connIdleTimeout` lowered from 1s to 100ms.
- Extracted idle timeout logic into `brokerCxn.isIdleTimeout`, shared by both `loadConnection` and `reapConnections`.
- Added `TestIssue1245` in `kfake/issues_test.go` to verify the `loadConnection` idle check creates a new connection.

Fixes https://github.com/twmb/franz-go/issues/1231

### DIFF
--- a/pkg/kfake/issues_test.go
+++ b/pkg/kfake/issues_test.go
@@ -2505,3 +2505,102 @@ func TestIssue1217(t *testing.T) {
 		}
 	})
 }
+
+// TestIssue1245 verifies that idle connections are not reused after
+// connIdleTimeout has elapsed. PR #1245 adds a check in loadConnection
+// so that a connection idle for longer than connIdleTimeout is killed
+// and a new connection is established, rather than reusing the stale one.
+//
+// The test is designed to exercise the loadConnection idle check rather
+// than the background reap loop. With ConnIdleTimeout of 200ms, the reap
+// loop ticks at t=200ms, 400ms, ... from client creation. The first
+// produce fires immediately (~t=10ms), so the produce connection's
+// lastRead/lastWrite is ~t=10ms. At the t=200ms reap tick the connection
+// has been idle for only ~190ms which is NOT > 200ms, so it is not
+// reaped. The second produce fires at ~t=310ms (idleTimeout + 100ms
+// after the first produce returns), well before the t=400ms reap tick.
+// At this point the produce connection has been idle for ~300ms > 200ms,
+// so loadConnection's isIdleTimeout check catches it and creates a new
+// connection.
+func TestIssue1245(t *testing.T) {
+	t.Parallel()
+	const testTopic = "foo"
+
+	c, err := NewCluster(
+		NumBrokers(1),
+		SeedTopics(1, testTopic),
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer c.Close()
+
+	// Track the number of new broker connections.
+	connects := new(atomic.Int32)
+
+	idleTimeout := 200 * time.Millisecond
+
+	cl, err := kgo.NewClient(
+		kgo.SeedBrokers(c.ListenAddrs()...),
+		kgo.DefaultProduceTopic(testTopic),
+		kgo.ConnIdleTimeout(idleTimeout),
+		kgo.ProducerLinger(0),
+		// Set metadata min and max age to 1hr so that no background
+		// metadata refreshes create or reuse connections during the test.
+		kgo.MetadataMinAge(time.Hour),
+		kgo.MetadataMaxAge(time.Hour),
+		kgo.WithHooks(&connCountHook{connects}),
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer cl.Close()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	// First produce: fires immediately after client creation (~t=10ms),
+	// establishing a produce connection whose lastRead/lastWrite is set
+	// to approximately now.
+	if err := cl.ProduceSync(ctx, kgo.StringRecord("msg-1")).FirstErr(); err != nil {
+		t.Fatalf("first produce failed: %v", err)
+	}
+
+	connectsAfterFirst := connects.Load()
+	if connectsAfterFirst == 0 {
+		t.Fatal("expected at least 1 connection after first produce")
+	}
+
+	// Sleep idleTimeout + 100ms. This is timed so that:
+	// - The reap loop at t=200ms sees the produce connection idle for
+	//   ~190ms (NOT > 200ms), so it does NOT reap it.
+	// - The second produce fires at ~t=310ms, well before the t=400ms
+	//   reap tick.
+	// - At ~t=310ms the produce connection has been idle for ~300ms >
+	//   200ms, so loadConnection's isIdleTimeout check catches it and
+	//   forces a new connection to be created.
+	time.Sleep(idleTimeout + 100*time.Millisecond)
+
+	// Second produce: loadConnection detects the idle produce connection,
+	// kills it, and creates a new one.
+	if err := cl.ProduceSync(ctx, kgo.StringRecord("msg-2")).FirstErr(); err != nil {
+		t.Fatalf("second produce failed: %v", err)
+	}
+
+	// OnBrokerConnect is only called for new connections, so any increase
+	// in the counter means a new connection was established.
+	if connects.Load() <= connectsAfterFirst {
+		t.Fatal("second produce reused the same connection; expected a new connection to be created")
+	}
+}
+
+// connCountHook implements HookBrokerConnect to count new connections.
+type connCountHook struct {
+	connects *atomic.Int32
+}
+
+func (h *connCountHook) OnBrokerConnect(_ kgo.BrokerMetadata, _ time.Duration, _ net.Conn, err error) {
+	if err == nil {
+		h.connects.Add(1)
+	}
+}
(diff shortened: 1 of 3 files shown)


pkg/kgo/broker.go
```diff
@@ -555,6 +555,12 @@ func (b *broker) loadConnection(ctx context.Context, req kmsg.Request) (*brokerC
 		pcxn = &b.cxnSlow
 	}
 
+	// Do not reuse a connection that has been idle for longer than idle timeout.
+	// Kill it instead.
+	if *pcxn != nil && !(*pcxn).dead.Load() && (*pcxn).isIdleTimeout(b.cl.cfg.connIdleTimeout) {
+		(*pcxn).die()
```
carsonip: [to reviewer] should we `go (*pcxn).die()` instead to avoid blocking `loadConnection` if `.die()` blocks? or it is not expected for `.die()` to block? IMO adding `go` might create more problems than it solves

pkg/kgo/config.go
```diff
@@ -662,12 +662,12 @@ func RequestTimeoutOverhead(overhead time.Duration) Opt {
 	return clientOpt{func(cfg *cfg) { cfg.requestTimeoutOverhead = overhead }}
 }
 
-// ConnIdleTimeout is a rough amount of time to allow connections to idle
-// before they are closed, overriding the default 20s.
+// ConnIdleTimeout sets the amount of time after which an idle connection will
```
carsonip, line 665: [to reviewer] is this also a good time to bump the connIdleTimeout default by 1.5x so that it matches the previous expected duration after which a connection is not reused? No strong opinion here

pkg/kgo/broker.go
twmb: TLS connections do a bit of writing when closing them, and I've experienced problems in the past where that write can hang.

`die` does a connection close, calls a hook, closes a channel, and kills the ring.
* Connection close: can block if TLS, fast otherwise
* Hook: expected to be fast, but entirely in user control. Fine to call it concurrently, though.
* Channel close: fast, also causes any active write to quit (which there should be none)
* Ring: fast to close, causes an active response goroutine to exit (handleResps), but there shouldn't be one

`die` right now is called in any error path, so the connection should (likely) already be broken and we shouldn't expect any nice TLS close notify.

It's _probably_ fine to have this called inline here, but I think it's also fine to call die in a goroutine and looks to have some upside in rare cases.

If you do run this in a goroutine, you'll need to `*pcxn = nil` on the next line in this conditional, though, since it's possible that `dead.Swap(true)` hasn't yet run.

pkg/kgo/config.go
twmb: 30s is fine to me as well

carsonip:
@twmb thanks! I've updated the PR with
- calling die in a goroutine
- new default connIdleTimeout of 30s

I haven't tested the changes though, and copilot wasn't able to generate a test I'm happy with, but TBF testing with time is tricky. ~~Feel free to take over from here if you have the time, otherwise I'll pick it up again tomorrow.~~

edit: I'm working on it now.

pkg/kfake/go.mod
```diff
@@ -10,6 +10,8 @@ require (
 )
 
 require (
-	github.com/klauspost/compress v1.18.1 // indirect
+	github.com/klauspost/compress v1.18.2 // indirect
```
carsonip, line 13: [to reviewer] my IDE won't shut it about the go mod tidy. Let me start a separate PR to fix this.

pkg/kfake/go.mod
```diff
@@ -10,6 +10,8 @@ require (
...
 	github.com/pierrec/lz4/v4 v4.1.22 // indirect
 )
+
+replace github.com/twmb/franz-go => ../../
```
carsonip: [to reviewer] This really is a trap. Spent 1hr only to realize the kfake test is testing a published version of kgo. Is there a reason why there isn't a `replace` here in the first place?

pkg/kfake/issues_test.go
```diff
@@ -2505,3 +2505,100 @@ func TestIssue1217(t *testing.T) {
...
+	}
+
+	// Sleep idleTimeout + 300ms. This is timed so that:
+	// - The reap loop at t=1 sees the produce connection idle for ~0.9s
+	//   (NOT > 1s), so it does NOT reap it.
+	// - The second produce fires at ~t=1.3, well before the t=2 reap tick.
+	// - At t=1.3 the produce connection has been idle for ~1.2s > 1s, so
+	//   loadConnection's isIdleTimeout check catches it and forces a new
+	//   connection to be created.
+	time.Sleep(idleTimeout + 300*time.Millisecond)
```
carsonip: [to reviewer] the art of testing with time in go. I'm not sure if we can make it nicer. Also ideally we want to really test the new check in `loadConnection` instead of the periodic background reap loop. But given there's no good way to disable the reap loop, we're just hoping that when the 2nd produce is called, the idle connection wasn't closed by reap loop.

pkg/kfake/go.mod
carsonip: https://github.com/twmb/franz-go/pull/1246

pkg/kfake/issues_test.go
```diff
@@ -2505,3 +2505,100 @@ func TestIssue1217(t *testing.T) {
...
+// than the background reap loop. With ConnIdleTimeout of 1s (minimum),
+// the reap loop ticks at t=1, 2, 3... from client creation. The first
+// produce fires immediately (~t=0.1), so the produce connection's
+// lastRead/lastWrite is ~t=0.1. At the t=1 reap tick the connection has
+// been idle for only ~0.9s which is NOT > 1s, so it is not reaped. The
+// second produce fires at ~t=1.3 (idleTimeout + 300ms after the first
+// produce returns), well before the t=2 reap tick. At this point the
+// produce connection has been idle for ~1.2s > 1s, so loadConnection's
+// isIdleTimeout check catches it and creates a new connection.
+func TestIssue1245(t *testing.T) {
```
carsonip, line 2525: [to reviewer] since you already brought up Claude, this test is written using claude-4.6-opus-high, hope you don't mind it. It looks fine to me. A bit too much tied to the implementation details but I don't think we can avoid it at this stage.

pkg/kfake/go.mod
twmb: It's to catch FOOLS who don't know the entire codebase in memory!!

From what I know, replace directives in go.mod are poisonous and force importers to have a similar type of replace directive. I don't know replace 100%, but I try to only use it in main go.mod files, not libraries I ship.

Similar-ish thinking about go.work files. I know even less about them, but the main thing I avoid is committing them.

That said, I have a go.work file locally, and the Github action for running these tests initializes a go.work inline


```
go 1.25.0

use (
	.
	../..
	../kadm
	../kmsg
)
```

it'd probably be very useful to have at least a README.md in this directory with a testing section mentioning this. My bad! I've largely been the only developer on this repo

pkg/kfake/issues_test.go
```diff
@@ -2505,3 +2505,100 @@ func TestIssue1217(t *testing.T) {
...
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer c.Close()
+
+	// Track the number of new broker connections.
+	connects := new(atomic.Int32)
+
+	idleTimeout := time.Second // minimum allowed by config validation
```
twmb: Mind dropping the min idle timeout in kgo to 100ms in this PR? I think the more important thing is to reject accidental 0's. I doubt somebody will do 100ms but if they want to have problems, fine. 100ms will allow us to make the test time faster.

pkg/kfake/issues_test.go
twmb: Honestly I think this is fine, the reasoning in the comment makes sense. The main thing I'd change is "second produce" => "second produce (just below)", and then lower the times by changing the lower bound kgo accepts.

pkg/kfake/issues_test.go
twmb: :nice: (pretend this exists)

pkg/kfake/go.mod
carsonip: My understanding of a replace in go mod is that it affects the _direct_ usage of this package, e.g. in CI where we run test on kfake directly. On the other hand, when someone installs kfake package, the replace directive is not applied to them. I imagine the behavior is the same when using go.work, but go.work is explicitly for local workspace.

But given CI is doing the correct thing, let's not concern ourselves with this issue in this PR

pkg/kfake/issues_test.go
carsonip: as we go lower with connIdleTimeout, the more likely that (on slower machines like CI) the reap loop closes the connection instead of the new check we added. Some "flakiness" that won't fail the test, but just not testing what we want to test as the observed behavior should still be the same. Unless we expose a config to disable reap loop we'll have to live with that.

But adding a config will be another can of worms.

twmb: Heh yeah you're right, I wasn't thinking you're drop the wait to 30ms - I think a 200ms idle timeout and then 300ms sleep is slightly more robust, or 300/450, while still keeping the test fast enough. 30ms is _probably_ a bit tight with when the reap goroutine will run.

Overall though, lgtm mind squashing commits into one or two that make sense?

twmb:
Did you mean to push before the latest re-review request?

carsonip:
squashed and force pushed

twmb:
Will be released in next patch eta this week. Looking into #1222 if possible beforehand.

PR #1267, opened 2026-02-23
sr: add schema context support and generic context handling in srfake

c-julin:
## Summary

- Add `WithSchemaContext` / `DefaultSchemaContext` to scope any `sr.Client` request to a schema registry context by prepending `/contexts/{name}` to paths
- Add new API methods: `Contexts`, `DeleteContext`
- Add `ContextPrefix`, `Offset`, and `Limit` query parameters
- Update `srfake` with a generic context-stripping middleware, so all existing endpoints work under `/contexts/{name}/...`

## Test plan

- [x] `go test ./... -count=1 -race` passes in `pkg/sr`
- [x] New tests for previously-unsupported context-prefixed operations (register, list versions, get version, delete subject, get config, get raw schema)
- [x] New end-to-end test for `CreateSchema` with `WithSchemaContext`

### DIFF
--- a/pkg/sr/client.go
+++ b/pkg/sr/client.go
@@ -281,7 +281,7 @@ func (cl *Client) delete(ctx context.Context, path string, into any) error {
 
 func (cl *Client) do(ctx context.Context, method, path string, v, into any) error {
 	ctxName := cl.schemaContext(ctx)
-	if ctxName != "" {
+	if ctxName != "" && !strings.HasPrefix(path, "/contexts/") {
 		path = "/contexts/" + url.PathEscape(ctxName) + path
 	}
 


pkg/sr/api.go
```diff
@@ -166,6 +163,22 @@ func CommSubjectSchemas(l, r []SubjectSchema) (luniq, runiq, common []SubjectSch
...
+// This supports params [ContextPrefix], [Offset], and [Limit].
+func (cl *Client) Contexts(ctx context.Context) ([]string, error) {
+	// GET /contexts
+	var contexts []string
+	err := cl.get(ctx, "/contexts", &contexts)
+	return contexts, err
+}
+
+// DeleteContext deletes a context from the schema registry.
+func (cl *Client) DeleteContext(ctx context.Context, context_ string) error {
```
twmb: fix arg name => context_

pkg/sr/api.go
```diff
@@ -412,10 +425,20 @@ func (cl *Client) CreateSchemaWithIDAndVersion(ctx context.Context, subject stri
 	if err != nil {
 		return SubjectSchema{}, err
 	}
+	// When using schema contexts, the registry stores subjects with a
+	// context prefix (e.g. ":.myctx:subject"). We need to match against
+	// both the plain subject and the context-prefixed form.
```
twmb, line 430: This is universal? Meaning if I opt into using contexts, ALL schemas are now prefixed with the context?

pkg/sr/params.go
```diff
@@ -210,11 +224,17 @@ func Subject(s string) Param { return Param{subject: s} }
 // RawParams returns a Param with raw query parameters.
 func RawParams(v url.Values) Param { return Param{rawParams: v} }
 
+// ContextPrefix returns a Param that filters contexts by prefix.
+func ContextPrefix(pfx string) Param { return Param{contextPrefix: pfx} }
+
+// Offset returns a Param that sets the pagination offset.
+func Offset(o int) Param { return Param{offset: o} }
+
+// Limit returns a Param that sets the maximum number of results to return.
```
twmb, line 233: did I just ... not have this before? even though I had it above in merge / struct def? weird.

pkg/sr/clientopt.go
```diff
@@ -73,6 +73,12 @@ func PreReq(preReq func(req *http.Request) error) ClientOpt {
 	}}
 }
 
+// DefaultSchemaContext sets a default schema context applied to every request.
```
twmb: `sets a default schema registry context (namespacing) applied to every request`

trying to be a bit more clear since SR contexts and Go contexts can be confusing when used in the same space

pkg/sr/client.go
```diff
@@ -230,6 +233,21 @@ func (cl *Client) OptValues(opt any) []any {
 	}
 }
 
+var schemaContextKey = "context_key"
+
+// WithSchemaContext returns a context that scopes schema registry requests
+// to the given context name. For client-wide scoping, use [DefaultSchemaContext].
+func WithSchemaContext(ctx context.Context, name string) context.Context {
```
twmb:
```
// InContext scopes requests to be issued to the given schema registry context (namespacing).
// For client wide scoping, use [DefaultSchemaContext]. The following will return the subjects
// from the "foo" context:
//
//         cl.Subjects(sr.InContext(ctx, "foo"))
```

also wdyt of `InContext, or `WithContext` as the name? Maybe With is better, it's more standard. I think dropping "Schema" though helps to keep it shorter. "In" for some reason to my brain separates it from Go contexts, whereas WithContext kinda just still sounds like with a context.Context (which then means you need the "Schema" => WithSchemaContext).

pkg/sr/api.go
c-julin: have updated to to srContext

pkg/sr/client.go
c-julin: I do like the InContext suggestion, creates a good seperation from Go Contexts.

pkg/sr/api.go
c-julin: there is a default context which is represented by `:.:` but the schema isn't isn't prefixed with  this. However if a subject is created with a context  will always be prefixed.

twmb:
(claude comment)

## Bug: `Contexts()` and `DeleteContext()` break with `DefaultSchemaContext`

`Contexts()` and `DeleteContext()` construct paths starting with `/contexts/`, but `do()` unconditionally prepends `/contexts/{ctx}` when a context is set. This means `DefaultSchemaContext` breaks both methods:

```go
// With DefaultSchemaContext(".myctx"):
cl.Contexts(ctx)
// path starts as "/contexts"
// do() makes it "/contexts/.myctx/contexts"  -- wrong

cl.DeleteContext(ctx, ".other")
// path starts as "/contexts/.other"
// do() makes it "/contexts/.myctx/contexts/.other"  -- wrong
```

These two methods should bypass the context prefix -- either by not going through `do()`, or by having `do()` skip the prefix for paths that already start with `/contexts/`.

## Potential issue: `SchemaUsagesByID` double-qualifies subjects with contexts

When `CreateSchemaWithIDAndVersion` is called within a context, `SchemaUsagesByID` double-qualifies subjects:

1. `SchemaUsagesByID` calls `GET /contexts/.myctx/schemas/ids/{id}/versions`
2. Registry returns context-prefixed subjects: `[{subject: ":.myctx:foo", version: 1}]`
3. `SchemaUsagesByID` then calls `SchemaByVersion(ctx, ":.myctx:foo", 1)` -- passing the already-prefixed subject
4. `pathSubject` escapes it: `/subjects/%3A.myctx%3Afoo/versions/1`
5. `do()` adds context prefix again: `/contexts/.myctx/subjects/%3A.myctx%3Afoo/versions/1`

This is double context qualification -- the subject name includes the context AND the URL path includes the context. The test mocks are specifically coded to accept this double-qualified form, but it's unclear whether Confluent's real registry accepts context-prefixed subject names within context-scoped URL paths. If it doesn't, `SchemaUsagesByID` would 404 on step 3 before the subject matching logic is ever reached, and the `":%s:%s"` matching fix would be a band-aid on a deeper problem.

Worth verifying against a real Confluent registry.

c-julin:
Tested both against against the real schema reg and double context qualification is handled by the api but have updated the do() method to skip prefixing `/contexts/`

pkg/sr/api.go
```diff
@@ -412,10 +425,20 @@ func (cl *Client) CreateSchemaWithIDAndVersion(ctx context.Context, subject stri
...
+	ctxName, _ := ctx.Value(&schemaContextKey).(string)
```
twmb: Can this be `ctxName := cl.schemaContext(ctx)`?

c-julin: Good point, have updated. Thanks

Issue #1250, opened 2026-02-16
Running franzgo code with synctest


cupcicm:
Hello,

We use franz go extensively in our codebase, and in our tests. We are also starting to use [synctest](https://go.dev/blog/synctest) in a bigger and bigger fraction of our tests.
When using synctest and relying on sync.Mutex, the way the go maintainers chose to not make a Mutex part of the synctest bubble can be a problem. A goroutine which is blocked on a Mutex is not "durably blocked" and so it is not enough to make time go forward in a synctest bubble. This problem is explained [here](https://go.dev/blog/synctest#mutexes) and the maintainers go in a little bit more details [here](https://go.dev/blog/testing-time#not-durably-blocking-mutexes)
Most of the time, Mutexes are held for a short amount of time to linearize access to a resource, but sometimes something like this can happen

```
goroutine 1:
  mu.Lock()
  defer mu.Unlock()
  time.Sleep(xxx)
  
goroutine2:
  mu.Lock()
  defer mu.Unlock()
  time.Sleep(yyy)
```
and if this happens the test will be blocked, as time cannot flow forward because there are goroutines that are not durably blocked from the point of view of the synctest bubble. 

In practice, this happens with one of our test where the `noCommitDuringJoinAndSync` mutex is held both by the `loopCommit` goroutine that is waiting for a network call to come back (and this call is batched and so incurs a very small `time.Sleep()` ) and by the `joinAndSync()` goroutine that is blocked for the same reason. 

I cannot guarantee that this is the only such mutex in the franzgo codebase, but in our testing, this is the only case we have found that blocked our synctests. 

I would like to know if you would be open to refactoring the code to either use a channel-based mutex class or refactoring a little bit around this to use a channel rather than a lock. I could provide a pull request. 

Another solution is to create a type alias (`xsync.Mutex`) that is aliased: 
- to `sync.Mutex` if no go tags is provided at compile time
- to a channel-based mutex implementation if you provide the right go tag at compile time.

^ We have done this in our codebase and it works well for our tests. Here again, if you are open to this solution, I could provide you with a PR that implements this. 

I am also available to discuss this, I don't want to impose a solution of course, I would just like to write an even bigger portion of our tests with franzgo and synctest, without having to `kgo.DisableAutoCommit()` or to fork.

twmb:
I'm open to a new type that is internally channel based. I'm not sure how that'd look, implementation wise. syncCommitsMu probably has a similar problem.

twmb:
[...]

PR #1268, opened 2026-02-24
Create an internal "xsync" package that reimplements mutexes.

cupcicm:
When using synctests to test code that uses franzgo, it is possible to get deadlock, as described in #1250.

When build with `GOFLAGS=-tags=synctests`, the tests will use the reimplemented mutexes, which are channel-based so that synctests can track whether they are durably blocked or not.


pkg/kgo/internal/xsync/synctest_mutex.go
```diff
@@ -0,0 +1,182 @@
+//go:build synctests
+
+package xsync
+
+import "sync"
+
+// In this file we reimplement sync.Mutex and sync.RWMutex and we rely exclusively on channels to implement the mutex.
+// The reason we do this is that synctest cannot track the state of mutexes, so it considers that a goroutine is not 'durably blocked' when it is waiting on a mutex.
+// The consequence of this is that it is impossible to test code that uses mutexes with synctest. For example, this test hangs forever with synctest
```
twmb: formatting on comments, I stick to 80 unless I'm writing big comments on multiply-nested guts code

pkg/kgo/internal/xsync/synctest_mutex.go
```diff
@@ -0,0 +1,182 @@
...
+func (m *Mutex) Unlock() {
+	m.init()
+	select {
+	case m.ch <- struct{}{}:
+	default:
+		panic("sync: unlock of unlocked mutex")
+	}
+}
+
+type RWMutex struct {
```
twmb, line 63: I initially thought a rwmutex would be simpler and we'd just swap in the channel approach everywhere. That's not the case, so build tags make sense.

pkg/kgo/internal/xsync/synctest_mutex.go
```diff
@@ -0,0 +1,182 @@
...
+	once         sync.Once
+	w            Mutex
+	mu           Mutex
+	readerCount  int
+	writerWait   bool
+	readerSignal chan struct{}
+	writerSignal chan struct{}
```
twmb, line 68: I pointed claude at this PR and asked if there's a simpler idea, it did come up with something that I think is simpler, but I want to know what you think (it's initial idea was too simple: just make the rwmutex be a mutex, but that doesn't preserve write locks blocking read locks). Anyway:


```go
//go:build synctests

package xsync

import "sync"

// Mutex is a channel-based mutex for use with synctest. A goroutine blocked on
// a sync.Mutex is not "durably blocked" from synctest's perspective, which
// prevents time from advancing. Channel-based blocking is durably blocked.
type Mutex struct {
	once sync.Once
	ch   chan struct{}
}

func (m *Mutex) init() {
	m.once.Do(func() {
		m.ch = make(chan struct{}, 1)
		m.ch <- struct{}{}
	})
}

func (m *Mutex) Lock() {
	m.init()
	<-m.ch
}

func (m *Mutex) TryLock() bool {
	m.init()
	select {
	case <-m.ch:
		return true
	default:
		return false
	}
}

func (m *Mutex) Unlock() {
	m.init()
	select {
	case m.ch <- struct{}{}:
	default:
		panic("sync: unlock of unlocked mutex")
	}
}

// RWMutex is a channel-based readers-writer mutex with writer priority.
//
// The design uses a "gate token" pattern: a buffered-1 channel holds a single
// token. Readers take the token and immediately return it (passing through the
// gate). Writers take the token and hold it, which blocks all new readers until
// the writer unlocks. This provides writer priority naturally - the moment a
// writer calls Lock, new RLock calls block on the gate.
//
// A separate writerSignal channel (buffered 1) is used by the last active
// reader to wake a waiting writer. A stale signal can accumulate if readers
// drain while no writer is waiting; Lock drains any stale signal before
// checking the reader count.
type RWMutex struct {
	once         sync.Once
	mu           Mutex         // protects readerCount
	gate         chan struct{} // buffered 1; readers take+return, writers take+hold
	readerCount  int
	writerSignal chan struct{} // buffered 1; last reader signals writer
}

func (rw *RWMutex) init() {
	rw.once.Do(func() {
		rw.gate = make(chan struct{}, 1)
		rw.gate <- struct{}{}
		rw.writerSignal = make(chan struct{}, 1)
		rw.mu.init()
	})
}

func (rw *RWMutex) RLock() {
	rw.init()
	<-rw.gate // blocks while writer holds gate
	rw.mu.Lock()
	rw.readerCount++
	rw.mu.Unlock()
	rw.gate <- struct{}{} // pass gate along
}

func (rw *RWMutex) TryRLock() bool {
	rw.init()
	select {
	case <-rw.gate:
	default:
		return false
	}
	rw.mu.Lock()
	rw.readerCount++
	rw.mu.Unlock()
	rw.gate <- struct{}{}
	return true
}

func (rw *RWMutex) RUnlock() {
	rw.mu.Lock()
	rw.readerCount--
	if rw.readerCount < 0 {
		rw.mu.Unlock()
		panic("sync: RUnlock of unlocked RWMutex")
	}
	if rw.readerCount == 0 {
		select {
		case rw.writerSignal <- struct{}{}:
		default:
		}
	}
	rw.mu.Unlock()
}

func (rw *RWMutex) Lock() {
	rw.init()
	<-rw.gate // take gate, immediately blocking new readers
	// Drain any stale signal from a previous reader cycle.
	select {
	case <-rw.writerSignal:
	default:
	}
	rw.mu.Lock()
	if rw.readerCount > 0 {
		rw.mu.Unlock()
		<-rw.writerSignal // wait for last active reader
	} else {
		rw.mu.Unlock()
	}
}

func (rw *RWMutex) TryLock() bool {
	rw.init()
	select {
	case <-rw.gate:
	default:
		return false
	}
	select {
	case <-rw.writerSignal:
	default:
	}
	rw.mu.Lock()
	if rw.readerCount > 0 {
		rw.mu.Unlock()
		rw.gate <- struct{}{}
		return false
	}
	rw.mu.Unlock()
	return true
}

func (rw *RWMutex) Unlock() {
	select {
	case rw.gate <- struct{}{}: // return gate, unblocking readers
	default:
		panic("sync: Unlock of unlocked RWMutex")
	}
}

func (rw *RWMutex) RLocker() sync.Locker { return (*rlocker)(rw) }

type rlocker RWMutex

func (r *rlocker) Lock()   { (*RWMutex)(r).RLock() }
func (r *rlocker) Unlock() { (*RWMutex)(r).RUnlock() }

```

point being: mutex implementation is the same, rwmutex implementation uses a nifty two channel "gate" approach. The downside is a spurious drain when grabbing the write lock, but the upside is it's a fair bit conceptually simpler.

twmb:
What's the virtual network stack about, extending the kfake API? Also, mind switching the impl up? (Not to knock on the current impl, I just find the gate channel impl easier to reason about)

pkg/kgo/internal/xsync/synctest_mutex.go
cupcicm: Also, the mutex code is very heavily optimized, it sounds like a pain to lose this for production clients.

cupcicm:
OK I have switched the implementation. 
It's mostly claude written, but it looks legit, and it passes the tests which are the go codebase's tests. So, probably OK.

The virtual networking stack is here because we cannot allow a real TCP connection to happen, these contain IO that are not "durably blocked" (e.g. the synctest code does not know how much to advance the virtual clock when the code is waiting on an IO to complete). So instead we use pipes that are all inside the go process.

twmb:
I only posted the gate rewrite after reviewing both versions :)

Can we strip down the virtual stack to just the minimal amount needed? Similar to the Dialer opt in kgo - we don't need all the interfaces here. If I'm reading it correctly, the sole purpose is to override the listener and the dialer, so the VirtualNetwork can just be:




```go
// some doc here about why a person wants to use this
type VirtualNetwork struct {
    mu        sync.Mutex
    listeners map[int]*virtualListener
}

func (v *VirtualNetwork) Listen(network, address string) (net.Listener, error) { ... }
func (v *VirtualNetwork) DialContext(ctx context.Context, network, address string) (net.Conn, error) { ... }
```

usage per your example would be the same, but API surface would be minimal. If the purpose is to support synctest, only the virtual network is needed, extra delay / chaos wrappers can be in user defined listener wrapping / connection wrapping in their own application (this sounds similar to what I did with the [chaosConn](https://github.com/twmb/franz-go/blob/72fe24297b24b071f1a1e8524d6f6f65af3086bd/pkg/kgo/helpers_test.go#L378-L421)).

pkg/kfake/virtual.go
```diff
@@ -0,0 +1,246 @@
...
+		Conn:  serverConn,
+		delay: delay,
+	}
+	delayedClientConn := &delayedConn{
+		Conn:  clientConn,
+		delay: delay,
+	}
+
+	select {
+	case listener.connections <- delayedServerConn:
```
twmb: this can race with close,
* Close closes channel AFTER the check above
* Close goroutine runs, sees the buffer is empty, returns
* Here, connections is unbuffered, and this case is randomly chosen and the connection goes into the channel forever

Easy fix is to remove the buffer from the channel, which also eliminates the need for the goroutine closing below

cupcicm: Oh you're right ! Well spotted 😅 
Fixing.

twmb:
Ping on the VirtualNetwork comment

cupcicm:
Sorry, had missed it. 
Here is an update that removes all the unused code.

pkg/kfake/virtual.go
```diff
@@ -0,0 +1,141 @@
...
+type VirtualListener struct {
+	port        int
+	stack       *VirtualNetworkingStack
+	connections chan net.Conn
+	closed      chan struct{}
+	closeOnce   sync.Once
+}
+
+// FakeAddr implements net.Addr
+type FakeAddr struct {
```
twmb: This can be private, since it exists to satisfy an interface

pkg/kfake/virtual.go
```diff
@@ -0,0 +1,141 @@
...
+	listeners map[int]*VirtualListener
+}
+
+func NewVirtualNetworkingStack() *VirtualNetworkingStack {
+	return &VirtualNetworkingStack{
+		listeners: make(map[int]*VirtualListener),
+	}
+}
+
+// VirtualListener implements net.Listener using channels
```
twmb: This can be private

pkg/kfake/virtual.go
```diff
@@ -0,0 +1,141 @@
...
+	"net"
+	"strconv"
+	"sync"
+)
+
+// VirtualNetworkingStack represents an in-memory networking stack.
+// You can listen to `localhost:port` and then you can dial to it.
+// When you Dial(), you will be given a pipe and the server will receive
+// the other end of the pipe.
+type VirtualNetworkingStack struct {
```
twmb: please rename to VirtualNetwork

twmb (commented):
Two more important comments (internalize what doesn't need to be public) and one minor comment otherwise lgtm

pkg/kfake/virtual.go
cupcicm: Fixed.

pkg/kfake/virtual.go
cupcicm: Fixed.

pkg/kfake/virtual.go
cupcicm: Done.

twmb (commented):
I don't mean to throw a wrench in this PR at this point, but didn't the original version of this PR only convert the two mutexes that are held for a very long time? Almost all of these mutexes in this PR don't need to be the xsync mutex. You only need it on _I think_ syncCommitsMu and noCommitDuringJoinAndSync. I'm flagging this now because I have a new PR incoming that is also using mutexes (tight mutex) and I realize that editors / muscle memory will mean xsync is missed on new code going forward. I'd rather have targeted edits of mutexes that need this channel-backing, rather than trying to address all mutexes always and then forgetting them on any new mutex.

I'm not 100% tied to my view here though. What do you think?

(also it'd probably be good to have a comment for the reason why a person would use VirtualNetwork, if edits are made)

examples/testing_with_kfake_and_synctest/go.mod
```diff
@@ -0,0 +1,22 @@
+module testing_with_kfake
+
+go 1.26.0
+
+require (
+	github.com/twmb/franz-go v1.20.6
+	github.com/twmb/franz-go/pkg/kfake v0.0.0-20260218055430-fc72d8313608
+)
+
+require (
```
twmb, line 10: worth a go mod tidy

cupcicm:
> I don't mean to throw a wrench in this PR at this point
No problem, would be totally OK, it's your repo :) 

I can for sure change the PR so that only long-held mutexes are swapped.
The way we have solved it in our codebase though is that during CI we have a linter that looks at the code and tells you it's forbidden if you're using sync.Mutext. If you are OK with it, I could also add this linter to your codebase and this way you wouldn't be scared of forgetting. 
You might be annoyed a few times though, because of the muscle memory still writing all those `sync.Mutex` and the CI yelling at you. 

Tell me which one you want and I'll do it.
I'll also rebase / go mod tidy and add the comment

twmb:
I'm open to a linter for now and if I ever get too annoyed I'll change it in the future. I think it could be limited to non-test files in pkg/kgo/**. While you make the change / comment - please squash into one or two or three commits.

twmb:
I'm going to release kgo, so in lieu of a tiny bit of back and forth on nits, I opened #1303 which is this PR, plus:

* Lazy-init VirtualNetwork via sync.Once and removed NewVirtualNetwork, so the only exported kfake addition is the VirtualNetwork type itself. A zero value is usable; example updated to var stack kfake.VirtualNetwork.
* Added a doc comment on VirtualNetwork explaining the synctest rationale (in-process pipes so goroutines are durably blocked) and a no-copy-after-init note.
* Added a forbidigo golangci rule banning sync.Mutex / sync.RWMutex in non-test files directly under pkg/kgo/ — scoped so internal/xsync, tests, and other modules are exempt.
* go mod tidy on the example
* Squashed the main work back to one commit with you as author, plus two small follow-up commits for the lint rule and some unrelated lint cleanups I picked up across the dozen PRs that merged in the meantime.

Closing in favor of #1303. Thanks again!

cupcicm:
Ah thanks.
Sorry I was in holidays, away from my laptop. Thanks for going the extra mile and merging this one!

-------- Original Message --------

PR #1273, opened 2026-03-04
kgo: avoid ProduceSync panic for negative manual partitions

weeco:
## Summary
- avoid a panic in ProduceSync when a record has a negative partition (for example with ManualPartitioner)
- add a regression test that warms metadata then verifies invalid negative partition returns an error instead of panicking

## Root cause
ProduceSync only guarded the upper partition bound before indexing pd.partitions[r.Partition]; negative values could still index and panic.

## Validation
- ran targeted test against a real broker (Redpanda) with one broker and KGO_TEST_RF=1
- command:
  - go test ./pkg/kgo -run TestIssueProduceSyncManualPartitionNegativeDoesNotPanic -count=1
- result: pass after fix

### DIFF
--- a/pkg/kfake/kafka_tests/helpers_test.go
+++ b/pkg/kfake/kafka_tests/helpers_test.go
@@ -157,29 +157,6 @@ func waitForStableGroup(t *testing.T, adm *kadm.Client, group string, nMembers i
 	}
 }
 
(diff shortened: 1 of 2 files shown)


twmb:
Please drop the waitForStableGroupAssigned, I have a different fix locally - then will merge

twmb:
failure likely due to branching off old master, merging

PR #1276, opened 2026-03-05
producer: avoid closure allocation in produce() by inlining buffer check

playmaker:
### Summary

This change removes one allocation on the Produce hot path by inlining the buffer-limit calculation that was previously done inside a closure (`calcNums`). Instead of allocating a new closure on every `Produce()` call, we compute `nextBufRecs`, `nextBufBytes`, `overMaxRecs`, and `overMaxBytes` inline in the locked section and in the blocking goroutine.

### Motivation

On high-throughput produce paths, the closure allocation shows up in allocation profiles and adds avoidable GC pressure. Inlining the logic removes that allocation without changing behavior.

### Changes

- **`pkg/kgo/producer.go`**  
  - Removed the `calcNums` closure.  
  - Inlined the same logic in two places: (1) after acquiring the lock (for the non-blocking path), and (2) inside the goroutine that runs when the caller would block (so the numbers are computed under the same lock before spawning the goroutine).  
  - No change to semantics or to when we block or trigger a drain.

- **`pkg/kgo/produce_alloc_bench_test.go`** (new)  
  - Adds `BenchmarkProduceAllocsMinimalCaller` (and related bench) to measure allocations on the Produce path with minimal caller-side allocation, so that `allocs/op` reflects library behavior.

### How to verify

- Run tests: `go test ./pkg/kgo/ -run TestClient_Produce -count=1` (Kafka required; use `KGO_TEST_RF=1` for single-broker).  
- Run the alloc benchmark: `go test ./pkg/kgo/ -bench=BenchmarkProduceAllocsMinimalCaller -benchmem -benchtime=5000x` (requires Kafka).

### DIFF
--- a/pkg/kgo/produce_alloc_bench_test.go
+++ b/pkg/kgo/produce_alloc_bench_test.go
@@ -1,0 +1,125 @@
+// Package kgo: benchmarks to reproduce Produce path allocations inside the library.
+//
+// These benchmarks minimize caller-side allocations so that allocs/op reflect
+// franz-go internal allocations:
+//   - producer.go: calcNums closure (1 alloc/call), promiseRecord []promisedRec slice
+//   - sink.go: newRecordBatch &recBatch{}, append(records, pr) growth
+//   - AppendVarintBytes / request buffer growth in encode path
+//
+// Baseline (upstream, no patches): BenchmarkProduceAllocsMinimalCaller typically
+// reports ~5 allocs/op. After the inline-only patch (no closure), expect ~4 allocs/op.
+//
+// Requires Kafka (KGO_SEEDS or 127.0.0.1:9092). The package init in helpers_test.go
+// also requires Kafka; if unavailable, the test binary may panic before the benchmark runs.
+//
+// Run:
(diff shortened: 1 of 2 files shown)


twmb:
This addresses one allocation but keeps two, I have a different implementation in #1277:
```
  ┌───────────┬────────┬────────────┬────────────────┐
  │           │ master │  PR #1276  │ perfopt (ours) │
  ├───────────┼────────┼────────────┼────────────────┤
  │ allocs/op │ 3      │ 2 (-33%)   │ 0 (-100%)      │
  ├───────────┼────────┼────────────┼────────────────┤
  │ B/op      │ 242    │ 178 (-26%) │ 160 (-34%)     │
  ├───────────┼────────┼────────────┼────────────────┤
  │ ns/op     │ 488    │ 330 (-32%) │ 319 (-35%)     │
  └───────────┴────────┴────────────┴────────────────┘
```

twmb:
Thank you for the idea, fwiw -  I wouldn't have poked the codepath without the idea

PR #1277, opened 2026-03-05
Perfopt

twmb:
Replaces #1276 

Inspired by @playmaker


playmaker:
Oh yeah , there's  another closure in this path, I forgot it :). Thanks for the improvement. So quick 👍

PR #1287, opened 2026-03-25
kprom, kvictoria: Add `BrokerLabels` Opt for fine-tuning broker metric labels

### LINKED ISSUE #1284: Client metrics `node_id` label can have high cardinality
We use a Kafka-compatible service whose brokers are not long-lived. They are k8s pods that can come and go. I've noticed for some of our services that use franz-go, particularly long-lived clients, [this metric](https://github.com/twmb/franz-go/blob/72fe24297b24b071f1a1e8524d6f6f65af3086bd/plugin/kprom/kprom.go#L10) `_read_errors_total` (and the writes one) sometimes grow to having ~30K different node_id values. When Prometheus comes to scrape metrics, the client application exhibits elevated CPU usage just for sorting/emitting all of these metrics. There are other metrics with `node_id` but for some reason I'm only seeing the error metrics' cardinality climb.

Maybe there could be a setting where `node_id` is suppressed or given a fixed value?

seizethedave:
fixes: https://github.com/twmb/franz-go/issues/1284

* Several labels can be emitted on the broker metrics. This PR adds a config `Opt` to allow control over which labels are emitted.
* Add the same option to `kvictoria` since it has the same label.
* Added some basic tests.

### DIFF
--- a/plugin/kprom/config.go
+++ b/plugin/kprom/config.go
@@ -21,6 +21,7 @@ type cfg struct {
 	histograms       map[Histogram][]float64
 	defBuckets       []float64
 	fetchProduceOpts fetchProduceOpts
+	brokerLabels     []string
 
 	handlerOpts  promhttp.HandlerOpts
 	goCollectors bool
@@ -38,6 +39,7 @@ func newCfg(namespace string, opts ...Opt) cfg {
 			uncompressedBytes: true,
 			labels:            []string{"node_id", "topic"},
 		},
+		brokerLabels: []string{"node_id"},
 	}
 
 	for _, opt := range opts {
@@ -196,7 +198,7 @@ func Histograms(hs ...Histogram) Opt {
 	return HistogramsFromOpts(hos...)
 }
 
-// A Detail is a label that can be set on fetch/produce metrics
+// A Detail is a label that can be set on fetch/produce metrics.
 type Detail uint8
 
 const (
@@ -250,4 +252,20 @@ func FetchAndProduceDetail(details ...Detail) Opt {
 			c.fetchProduceOpts.labels = labels
 		},
 	}
+}
+
+// BrokerNodeLabel controls whether the "node_id" label is included on
+// broker-level connection, read, write, and request metrics. By default
+// node_id is included. Passing false removes the label, aggregating
+// broker-level metrics across all brokers. This is useful for reducing
+// metrics cardinality in environments with many or frequently changing
+// broker IDs.
+func BrokerNodeLabel(include bool) Opt {
+	return opt{func(c *cfg) {
+		if include {
+			c.brokerLabels = []string{"node_id"}
+		} else {
+			c.brokerLabels = nil
+		}
+	}}
 }
--- a/plugin/kprom/config.go
+++ b/plugin/kprom/config.go
@@ -254,18 +254,38 @@ func FetchAndProduceDetail(details ...Detail) Opt {
 	}
 }
 
-// BrokerNodeLabel controls whether the "node_id" label is included on
-// broker-level connection, read, write, and request metrics. By default
-// node_id is included. Passing false removes the label, aggregating
-// broker-level metrics across all brokers. This is useful for reducing
-// metrics cardinality in environments with many or frequently changing
-// broker IDs.
-func BrokerNodeLabel(include bool) Opt {
+// A BrokerLabel is a label that can be set on broker-level metrics.
+type BrokerLabel uint8
+
+const (
+	BrokerNodeID BrokerLabel = iota // Include "node_id" label on broker metrics.
+	BrokerHost                      // Include "host" label on broker metrics.
+	BrokerRack                      // Include "rack" label on broker metrics.
+)
+
+// BrokerLabels configures which labels are included on broker-level
+// connection, read, write, and request metrics, overriding the default
+// of (BrokerNodeID). Calling with no arguments removes all broker-level
+// labels, aggregating metrics across all brokers. This is useful for
+// reducing metrics cardinality in environments with many or frequently
+// changing broker IDs.
+func BrokerLabels(labels ...BrokerLabel) Opt {
 	return opt{func(c *cfg) {
-		if include {
-			c.brokerLabels = []string{"node_id"}
-		} else {
-			c.brokerLabels = nil
+		seen := make(map[BrokerLabel]bool)
+		c.brokerLabels = nil
+		for _, l := range labels {
+			if seen[l] {
+				continue
+			}
+			seen[l] = true
+			switch l {
+			case BrokerNodeID:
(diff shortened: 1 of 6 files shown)


twmb:
Iterating on your idea and having this posted from Claude - what do you think about following a similar pattern to FetchAndProduceDetail and have something like the following:

```go
type BrokerLabel uint8

const (
    BrokerNodeID BrokerLabel = iota // Include "node_id" label on broker metrics.
    BrokerHost                      // Include "host" label on broker metrics.
    BrokerRack                      // Include "rack" label on broker metrics.
)

// BrokerLabels configures which labels are included on broker-level
// connection, read, write, and request metrics, overriding the default
// of (BrokerNodeID). Calling with no arguments removes all broker-level
// labels, aggregating metrics across all brokers.
func BrokerLabels(labels ...BrokerLabel) Opt
```

Usage:

```go
// Remove node_id (your use case):
kprom.BrokerLabels()

// Replace node_id with host (stable addresses instead of rotating IDs):
kprom.BrokerLabels(kprom.BrokerHost)

// Both node_id and rack (per-AZ dashboards):
kprom.BrokerLabels(kprom.BrokerNodeID, kprom.BrokerRack)

// Default (unchanged):
// just don't pass the option
```

The internal wiring stays similar to what you have. For kprom, `cfg.brokerLabels` remains `[]string` (passed directly to `NewCounterVec`), and the helper becomes:

```go
func (m *Metrics) brokerLabelValues(meta kgo.BrokerMetadata) []string {
    if len(m.cfg.brokerLabels) == 0 {
        return nil
    }
    values := make([]string, len(m.cfg.brokerLabels))
    for i, l := range m.cfg.brokerLabels {
        switch l {
        case "node_id":
            values[i] = kgo.NodeName(meta.NodeID)
        case "host":
            values[i] = meta.Host
        case "rack":
            if meta.Rack != nil {
                values[i] = *meta.Rack
            }
        }
    }
    return values
}
```

For kvictoria, store `map[BrokerLabel]bool` internally and check membership in the `brokerLabels(meta)` helper.

seizethedave:
That's nice as it's more general for the other labels. OTOH clients who pass this opt this will future-clobber any broker labels you may add in the future. Is that bad? Or, maybe it's reasonable behavior as ostensibly people passing the opt are picky about those labels anyway.

twmb:
I figure users that pass the option are deliberately opting into control, so missing out on anything I add in the future is fine for them. Also, I don't think I'll be adding even more broker labels.

seizethedave:
@twmb updated.

plugin/kvictoria/kvictoria.go
```diff
@@ -250,19 +234,30 @@ func (m *Metrics) OnProduceBatchWritten(meta kgo.BrokerMetadata, topic string, p
...
 
+func (m *Metrics) brokerLabels(meta kgo.BrokerMetadata) map[string]string {
+	labels := map[string]string{"client_id": m.clientID}
+	if m.cfg.brokerLabelSet[BrokerNodeID] {
+		labels["node_id"] = kgo.NodeName(meta.NodeID)
+	}
+	if m.cfg.brokerLabelSet[BrokerHost] {
+		labels["host"] = meta.Host
+	}
+	if m.cfg.brokerLabelSet[BrokerRack] && meta.Rack != nil {
```
twmb: I think we should include an empty rack name if the person wants the label, otherwise empty vs. non-empty results in a different time series (say a person sets a rack later). wdyt?

plugin/kvictoria/kvictoria.go
```diff
@@ -250,19 +234,30 @@ func (m *Metrics) OnProduceBatchWritten(meta kgo.BrokerMetadata, topic string, p
...
 
 	m.set.GetOrCreateCounter(m.buildName("fetch_uncompressed_bytes_total", labels)).Add(metrics.UncompressedBytes)
 	m.set.GetOrCreateCounter(m.buildName("fetch_compressed_bytes_total", labels)).Add(metrics.CompressedBytes)
 	m.set.GetOrCreateCounter(m.buildName("fetch_batches_total", labels)).Inc()
 	m.set.GetOrCreateCounter(m.buildName("fetch_records_total", labels)).Add(metrics.NumRecords)
 }
 
+func (m *Metrics) brokerLabels(meta kgo.BrokerMetadata) map[string]string {
+	labels := map[string]string{"client_id": m.clientID}
+	if m.cfg.brokerLabelSet[BrokerNodeID] {
```
twmb, line 249: Any reason for the different implementation that does the same thing between kprom and kvictoria? kprom uses a slice and does string matching when deciding what labels to add, this uses a map

plugin/kprom/config.go
```diff
@@ -251,3 +253,39 @@ func FetchAndProduceDetail(details ...Detail) Opt {
...
+const (
+	BrokerNodeID BrokerLabel = iota // Include "node_id" label on broker metrics.
+	BrokerHost                      // Include "host" label on broker metrics.
+	BrokerRack                      // Include "rack" label on broker metrics.
+)
+
+// BrokerLabels configures which labels are included on broker-level
+// connection, read, write, and request metrics, overriding the default
+// of (BrokerNodeID). Calling with no arguments removes all broker-level
+// labels, aggregating metrics across all brokers. This is useful for
```
twmb: after "across all brokers", please make the comment something like:
```
// ... across all brokers:
//
//    kprom.BrokerLabels( /* none */ )
//
// This is useful for...
```
(same comment in kvictoria).

plugin/kvictoria/kvictoria.go
seizethedave: Ah yes, just like kprom.

plugin/kvictoria/kvictoria.go
seizethedave: Yeah, the gist is the Prometheus client_golang library gets all the label names bound at metric instantiation time, and at metric tracking time it just takes a slice of the corresponding values.

New to VM but it looks like they do the `statsd`-like thing where they accept all name/value pairs at tracking time, hence the map.

I'll see if there's a reasonable opportunity to make them more similar.

seizethedave: Now having looked, I don't think it makes sense to do anything about that.

plugin/kvictoria/kvictoria.go
seizethedave: 9868fc38464dbba8d3e55c0580784d29abfd342a

plugin/kprom/config.go
seizethedave: 0724aafa

plugin/kvictoria/kvictoria.go
twmb: Sounds good, thanks for looking

twmb (approved):

PR #1289, opened 2026-03-31
kgo: add HookPollStart hook

rarguelloF:
As part of the franz-go [dd-trace-go integration](https://github.com/DataDog/dd-trace-go/pull/4623), we need a way to finish consumer spans. In other Kafka integrations we follow the convention of closing the previous poll's spans when the next fetch begins, but franz-go had no hook for that boundary.

This PR adds `HookPollStart `, which fires at the top of every `PollFetches`/`PollRecords` call before any records are drained, giving integrations a clean place to finish spans opened during the previous poll.

### DIFF
--- a/pkg/kgo/hooks.go
+++ b/pkg/kgo/hooks.go
@@ -1,6 +1,7 @@
 package kgo
 
 import (
+	"context"
 	"net"
 	"time"
 )
@@ -394,7 +395,7 @@ type HookFetchRecordUnbuffered interface {
 	OnFetchRecordUnbuffered(r *Record, polled bool)
 }
 
-// HookPollRecordsStart is called at the beginning of every PollFetches or
+// HookPollStart is called at the beginning of every PollFetches or
 // PollRecords call, before any records are drained from internal buffers and
 // before any HookFetchRecordUnbuffered hooks fire for the same poll.
 //
@@ -402,10 +403,10 @@ type HookFetchRecordUnbuffered interface {
 // tracing integration that opens a span per record via HookFetchRecordUnbuffered
 // can implement this hook to finish the previous poll's spans before new ones
 // are created.
-type HookPollRecordsStart interface {
-	// OnPollRecordsStart is called at the start of every PollFetches or
-	// PollRecords call.
-	OnPollRecordsStart()
+type HookPollStart interface {
+	// OnPollStart is called at the start of every PollFetches or
+	// PollRecords call with the context passed by the caller.
+	OnPollStart(ctx context.Context)
 }
 
 /////////////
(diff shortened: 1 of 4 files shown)


rarguelloF:
Hey @twmb, when you get a chance, could you please take a look at this PR? I’d really appreciate it, thanks! 🙏

twmb:
Not entirely opposed, but this seems very niche - what blocks you from finalizing prior spans right before calling poll, application side?

genesor:
Hey @twmb, we do not control the code on the application side that's why we'd like to have a hook allowing us to terminate the spans started in the previous poll. 

We're aiming to have the tracing as seamless as possible, starting spans/traces and asking the customer to end it is not ideal. Some of our integrations work like this and it tends to create issues (un-finished spans and traces causing the trace to never be sent)

We've explored multiple ways of implementing this, one being wrapping the client and terminating spans before passing the call to the original client (https://github.com/DataDog/dd-trace-go/pull/4250/changes#diff-c86fb5badb2ad9affd4117cc7b2b0e52ef54e9252a13495f3c4ac60d129977dcR85) but it's not ideal and can cause changes on application code for users

pkg/kgo/hooks.go
```diff
@@ -394,6 +394,20 @@ type HookFetchRecordUnbuffered interface {
...
+// before any HookFetchRecordUnbuffered hooks fire for the same poll.
+//
+// This hook is useful for instrumenting the poll boundary: for example, a
+// tracing integration that opens a span per record via HookFetchRecordUnbuffered
+// can implement this hook to finish the previous poll's spans before new ones
+// are created.
+type HookPollRecordsStart interface {
+	// OnPollRecordsStart is called at the start of every PollFetches or
+	// PollRecords call.
+	OnPollRecordsStart()
```
twmb: Do you foresee any use of having this take the context that the user passes to `PollRecords`?
Should this be named `OnPollStart` / `HookPollStart` since it's used for both PollFetches and PollRecords?

rarguelloF: > Do you foresee any use of having this take the context that the user passes to PollRecords?

Not for our use case, but I don't think there's any harm on adding it 👍 

>Should this be named OnPollStart / HookPollStart since it's used for both PollFetches and PollRecords?

Makes sense! changing it to `OnPollStart` / `HookPollStart`

twmb (approved):

genesor:
Hi @twmb, thanks for merging the PR 🚀 

Do you have an idea of when you're going to release the next version so we can start using it in dd-trace-go ?

twmb:
follow #1302

PR #1295, opened 2026-04-12
Share me

twmb:


adamdecaf:
@twmb was it intentional to change the behavior of `MaxConcurrentFetches(0)` to a single fetch allowed? This was noticed in our performance testing that we need to use `-1` now. 

https://github.com/twmb/franz-go/commit/0f0bca224fb176899299b1ce7a5b3b76e63e85cb#diff-6ce7762f7c471d61c8c76b07e817224c5734aff491eeb382ceb04ad981588728R1421

Issue #239, opened 2022-10-27
Producer failover


drTr0jan:
I've a Kafka cluster with 3 brokers. But I've faced a problem while one broker was falled down (ex., upgrading or failure).
The producer continues trying to `ProduceSync()` to fallen broker infinitely. Sometimes producer switches to other workable broker.

I've investigated how the `ProduceSync()` works and found out that it uses `Sinc.produce()` and as a consequence, it use `Sinc.doSequenced()`. Is it correct?
`Sinc.doSequenced()` uses only one broker, fetched by `Client.brokerOrErr()`. And doesn't changed it after errors.

Is it correct behaviour?

How can I limit the producing loop by time? I want to drop an error if `ProduceSync` cannot to produce after 10s, example.

```
<134>1 14:36:46.810145 beginning transaction transactional_id=http-kafka-28692 
<135>1 14:36:46.810351 opening connection to broker addr=192.168.178.83:9093 broker=3 
<135>1 14:36:46.817657 connection opened to broker broker=3 addr=192.168.178.83:9093 
<135>1 14:36:46.817770 connection initialized successfully addr=192.168.178.83:9093 broker=3 
<135>1 14:36:46.817821 wrote AddPartitionsToTxn v3 time_to_write=29.227µs err=<nil> broker=3 bytes_written=83 write_wait=7.403375ms 
<135>1 14:36:46.823620 read AddPartitionsToTxn v3 err=<nil> broker=3 bytes_read=50 read_wait=78.86µs time_to_read=5.755618ms 
<135>1 14:36:46.823723 retrying request tries=1 backoff=230.298076ms request_error=<nil> response_error=COORDINATOR_NOT_AVAILABLE: The coordinator is not available. 
<135>1 14:36:47.057487 wrote FindCoordinator v4 broker=3 bytes_written=42 write_wait=185.586µs time_to_write=42.689µs err=<nil> 
<135>1 14:36:47.058197 read FindCoordinator v4 broker=3 bytes_read=62 read_wait=145.794µs time_to_read=611.298µs err=<nil> 
<135>1 14:36:47.058400 wrote AddPartitionsToTxn v3 write_wait=51.985µs time_to_write=36.932µs err=<nil> broker=3 bytes_written=83 
<135>1 14:36:47.059458 read AddPartitionsToTxn v3 broker=3 bytes_read=50 read_wait=143.05µs time_to_read=910.882µs err=<nil> 
[...]

PR #1297, opened 2026-04-15
kgo: retry transient dial errors for the configured retry budget

twmb:
When a broker is mid-rolling-restart, dial attempts get ECONNREFUSED.
The failDial mechanism converted these into retryable errors, but capped
at 3 attempts (~1.5s with default backoff). Typical broker restarts last
5-30s, so the cap was an order of magnitude too short and silently
overrode the user-configured RetryTimeout.

Drop the cap. The retry loop is already bounded by RetryTimeout (default
30s), RequestRetries (default 20), and the caller's context. These are
the knobs users configure and expect to be respected.

Additionally:

- Distinguish permanent dial errors (NXDOMAIN, EACCES, EPERM) and
  surface them immediately rather than retrying. This preserves the
  fast-fail behavior for misconfigured seed addresses.

- On any dial failure, clear all cached roles pointing at the dead
  broker (controller ID, all group/txn coordinator entries), not just
  the narrow entry for the current request. A dead broker is a dead
  broker regardless of which code path discovered it; clearing
  everything lets concurrent and future requests discover role
  reassignments via fresh FindCoordinator / metadata lookups.

- Add failDial to Broker.RetriableRequest, which previously had no
  dial-error retry handling at all. A single ECONNREFUSED was surfaced
  directly to the caller despite the function promising to retry
  connection errors.

The 3-cap was introduced in e2e80bf (Dec 2022, #239) as a heuristic for
"transient blips," paired with coordinator cache invalidation. It was
effectively dead code until d922b88 (March 2026) fixed failDial to
actually retry. With the fix in place, 3 attempts is still far too few
for real-world restart windows.

### DIFF
--- a/pkg/kgo/client.go
+++ b/pkg/kgo/client.go
@@ -1486,28 +1486,30 @@ type retryable struct {
 }
 
 type failDial struct {
-	fails   int8
 	clearFn func()
 }
 
-// The controller and group/txn coordinators are cached. If dialing the
-// cached broker fails, we clear the cache immediately so that the next
-// retry does a fresh lookup, and we return errChosenBrokerDead to make
-// the retryable loop actually retry with backoff (dial errors are
-// otherwise not retryable by shouldRetry, and shouldRetryNext cannot
-// help because loadCoordinator returns the same stale cached broker).
+// handleDialErr converts transient dial errors into errChosenBrokerDead so
+// that the retryable loop retries with backoff. isRetryableBrokerErr excludes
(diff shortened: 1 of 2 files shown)


dcrodman (approved):
Confirmed that while running a typical consumer using a `PollRecords` / `CommitUncommittedOffsets` loop, restarting a broker no longer causes either of those client methods to return errors within the 1.5s window described. Instead, they appear to respect RetryTimeout.

PR #1301, opened 2026-04-16
sr: document retry configuration on HTTPClient

### LINKED ISSUE #1288: http client: add retries (schema registry)
Hello,

Our schema registry can sometimes not be available for a short time.
It would be nice to see retries on http request failures when the schema registry can not be reached.
Ideally this would be configurable for the developer using the library (retry count + technique: exponential backoff settings or linear).

best

twmb:
The client performs no retries beyond one pass over configured URLs on
transport errors, and does not retry on 429/5xx. Document this on the
HTTPClient option and show how to layer retries via a wrapping
RoundTripper (e.g. hashicorp/go-retryablehttp).

Closes #1288.


twmb:
Merging now; not pushing a tag for sr because so far the package only has a doc change.

PR #1311, opened 2026-04-23
EnsureProduceTopicsAreReady proposal

deep-creek:
we have an issue in our services when testing regarding the readiness. 
Consider the following case:
- Our test framework instantiates a service and waits until the service reports ready
  - Internally it uses franz-go, instantiates a client, calls _EnsureProduceConnectionIsOpen_ and _ProducerID_ on it and assumes it is ready 
- the test framework waits until the service reports ready and sends input data
- the service tries to send a Kafka message using _Produce_

Now, if our CI Kafka is a bit slow which it sometimes is (should use Redpanda instead!), we encounter timing issues due to the fact, that the first _Produce_ tries to obtain the Metadata and logs:
`
immediate metadata update triggered; why: forced load because we are producing to a topic for the first time 
producing to a new topic for the first time, fetching metadata to learn its partitions; topic: ...
done waiting for metadata for new topic; topic: ...
`

What we thus need, is a way to ensure that we are really ready. For that purpose, we want to trigger the metadata retrieval process and wait for it's response, before declaring the service ready.
The code is a proposal implementation for that...

Best regards
Ben

### DIFF
--- a/pkg/kgo/producer.go
+++ b/pkg/kgo/producer.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"slices"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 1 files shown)


twmb:
I'm not sure I'm too keen on this one, from what I can see, it's basically useful in only one very niche use case. Most / all things added to the client are at least useful in some form of long-lived use case, but this is useful only in the startup case. It's adding something to the client to try to work around slowness in broker startup.

What's the downside of just relying on produce itself loading metadata? You could set the max metadata age low, or the retry backoff to always return 100ms or something, to basically for spinloop loading the topic info as fast as possible. Producing is meant to load metadata the first time you produce - why are you receiving errors via the default load path?

deep-creek:
The downside of just relying on produce itself to loading metadata at the beginning is the timing. We really try to get into a position with our services, where we know, that they are ready to process incoming requests before bringing the ready probes up. Therefore we are already call _ProducerID_ and _EnsureProduceConnectionIsOpen_ before we declare ready. _Produce_ itself doesn't seem to be a problem regarding latency, but everything regarding metadata requests is, when using Apache Kafka and a high number of topics (>10.000) in our environment. 

As this is mainly an issue when black box testing our services, we might increase the timeout while waiting for a response (currently 100ms). The downside of that, is that we would not only increase the time waiting for message, but also the time when testing that there are no messages being generated. With our completely asynchronous services, that is a problem, if a service e.g. listens to 5 topics and depending on the input either generates or does not generate output.
If the metadata would already be requested by calling _ProducerID_ or _EnsureProduceConnectionIsOpen_ or any other mechanism, that would help as well. Without the proposed function, the first Produce tries to perform the implemented lazy initialization, and that even if the Metadata has already been received as the producer.topics are only populated if the producer itself has issued a metadata request.

twmb:
I'm still not sure about this one. From your writeup, it sounds like this is a test only issue - the test tolerance is too tight such that the current setup - which queries metadata lazily - exceeds 100ms. The easy fix imo is to deflake the test - send a sentinel message through the producer first, and then start a test timeout. The difference between EnsureProduceConnectionIsOpen and this is that EnsureProduceConnectionIsOpen is useful for a generally live client that wants to keep latencies low, throughout the whole lifecycle of a client. This function helps one specific case - startup. A sentinel message can do the same thing, and since this is a test harness, you could easily filter the message as well when being consumed.

I did consider shipping a debug-only escape hatch - a magic context key, GODEBUG-style, with no exported symbol and just a doc reference (similar to the 848 opt in at the moment). I don't think that's worth it though, it starts to complicate the produce path a bit. The PR as is has bugs and the better fix, if ever, would be to go the route I was thinking - add a fake record to the unknownTopics map/slice, and when processing it after metadata load, if the record has the context key (or is an internal sentinel record), just don't partition it.

The readiness probes in production can't really be solved with a fake record because you wouldn't want to inject a fake record through the pipeline, but also I expect production services are both more tolerant of slight latency spikes (I mean, producing itself to kafka can have variance of >100ms) and don't worry about the startup time as much. For testing, well,... you can control the expectations of your testing harnesses. The other aspect of this is your PR was opened days after I shipped 1.21.0, and I'm likely not going to ship any API changes for months now at this point while I wait for either Kafka 4.3 or enough feature requests to build up.

deep-creek:
I’m going to close this PR for now.

After thinking more about this in the context of real‑world production usage, I don’t think the assumption this change relies on is strong enough. Even if a producer successfully retrieves metadata during startup, there are several realistic scenarios where it still might not be ready later on — especially if the application doesn’t produce for an extended period of time. In those cases, the initially fetched metadata can easily be stale by the time the first record is actually sent, which undermines the guarantee this approach is trying to provide.

Because of that, I’m hesitant to bake this behavior into the core library right now. The problem space is broader than startup readiness alone, and handling it robustly likely requires a different shape or approach. I’m not ruling this out entirely, but given these concerns, I don’t think this PR is the right solution as‑is.

Lastly, a sincere thank you, Travis for the outstanding maintenance of franz‑go. The level of care and consistency you bring to the project makes it much easier to reason about subtle trade‑offs like this.

PR #1313, opened 2026-04-23
kgo: add DisableRequireStableFetchOffsets option

### LINKED ISSUE #1312: Consumer group stops joining against Azure Event Hubs after upgrade to v1.21.0 — UNKNOWN_TOPIC_OR_PARTITION on offset fetch
## Summary

After upgrading [`github.com/twmb/franz-go`](https://github.com/twmb/franz-go) from `v1.20.6` to `v1.21.0`, a consumer group that talks to Azure Event Hubs (Kafka-compatible endpoint) can no longer join its group. Every join attempt fails during the offset-fetch step with a group-level `UNKNOWN_TOPIC_OR_PARTITION` error. Reverting the dependency to `v1.20.6` — with no application code changes and no broker / Event Hubs configuration changes — immediately resolves the problem and the group rejoins cleanly. The regression is purely client-side.

## Environment

- `franz-go`: `v1.21.0` (broken), `v1.20.6` (working)
- Broker: Azure Event Hubs, Kafka-compatible endpoint
- Go: 1.26
- Client type: plain consumer group (no transactional producer, `CommitOffsets` used for offset management)

## Client configuration (relevant bits)

```go
kgo.NewClient(
    kgo.ConsumerGroup(group),
    kgo.ConsumeTopics(topic),
    kgo.MinVersions(kversion.V1_0_0()),
    kgo.MaxVersions(kversion.V3_8_0()), // Event Hubs doesn't support newer
    // no kgo.RequireStableFetchOffsets() — never opted in
)
```

## Observed behavior on v1.21.0

Every rejoin attempt the library logs:

```
LogLevelError  fetch offsets failed with group-level error  group=<group>  err=UNKNOWN_TOPIC_OR_PARTITION: This server does not host this topic-partition.
LogLevelError  group manage loop errored                    group=<group>  err=UNKNOWN_TOPIC_OR_PARTITION: This server does not host this topic-partition.
```

These are emitted from [`consumer_group.go#L1797`](https://github.com/twmb/franz-go/blob/v1.21.0/pkg/kgo/consumer_group.go#L1797) and [`consumer_group.go#L479`](https://github.com/twmb/franz-go/blob/v1.21.0/pkg/kgo/consumer_group.go#L479) respectively — and `fetch offsets failed with group-level error` is the exact new branch introduced in v1.21.0 that surfaces the response-level `resp.ErrorCode` on `OffsetFetch` (see [`consumer_group.go#L1773`](https://github.com/twmb/franz-go/blob/v1.21.0/pkg/kgo/consumer_group.go#L1773)). The topic exists and is reachable — a standalone test producer against the same endpoint writes to it without issue. Downgrading to `v1.20.6` makes both messages disappear instantly.

## Behavior on v1.20.6

No errors. Group joins, offsets fetch successfully, consumer polls records normally. Same brokers, same topic, same client code, same broker/Event Hubs configuration — only `franz-go` version differs.

## Suspected cause

Looking at the [`v1.20.6...v1.21.0`](https://github.com/twmb/franz-go/compare/v1.20.6...v1.21.0) diff for `pkg/kgo/consumer_group.go`, two changes land in `fetchOffsets` that together match the symptom:

1. [`req.RequireStable = true`](https://github.com/twmb/franz-go/blob/v1.21.0/pkg/kgo/consumer_group.go#L1721) is now hardcoded instead of reading from `g.cfg.requireStable`. The previous opt-in option (`RequireStableFetchOffsets`) is now a deprecated no-op with no way to turn the behavior off.
2. The response's top-level `resp.ErrorCode` is now [checked and surfaced](https://github.com/twmb/franz-go/blob/v1.21.0/pkg/kgo/consumer_group.go#L1773) as a group-level error (`fetch offsets failed with group-level error`), where previously only per-partition error codes were inspected.

Per [KIP-447](https://cwiki.apache.org/confluence/display/KAFKA/KIP-447%3A+Producer+scalability+for+exactly+once+semantics), when `RequireStable=true` is set and the group has pending transactional offset commits, the broker is expected to return `PENDING_TRANSACTION` (error code 85) and the client retries. Event Hubs is instead returning `UNKNOWN_TOPIC_OR_PARTITION` at the group/response level, which isn't part of KIP-447's vocabulary. Whatever the broker is actually doing internally, the combination that breaks us is:

- pre-v1.21.0: client didn't set `RequireStable=true` by default **and** didn't surface the group-level error code → group joined fine.
- v1.21.0: client forces `RequireStable=true` **and** now checks/surfaces the group-level error code → group join fails.

Either change alone would probably have been survivable; both together aren't.

## Request

Could you restore a client-side opt-out for the `RequireStable` behavior on group `OffsetFetch`? Some form of `kgo.RequireStableFetchOffsets(false)` or a new `kgo.DisableRequireStableFetchOffsets()`, so users on non-transactional brokers can keep the v1.20.x behavior.

A more surgical fix, if preferable: skip setting `RequireStable=true` when the negotiated broker `ApiVersions` don't advertise `OffsetFetch` at a version that supports KIP-447 semantics (v7+), or gate it behind an explicit opt-in again. The current behavior is a silent, non-configurable break for any user running against a Kafka-compatible-but-not-full broker.

---

*Created with Cursor using Claude Opus 4.7*

adamkasztenny:
## Summary

Adds `kgo.DisableRequireStableFetchOffsets()` as a group-consumer option that sets `RequireStable=false` on `OffsetFetch` requests, restoring the pre-v1.21.0 wire behavior for users whose brokers do not implement KIP-447.

Fixes #1312.

## Context

In v1.21.0, [`consumer_group.go`](https://github.com/twmb/franz-go/blob/v1.21.0/pkg/kgo/consumer_group.go#L1721) started hardcoding `req.RequireStable = true` on group `OffsetFetch` requests, and the previous opt-in `RequireStableFetchOffsets()` became a deprecated no-op. This is correct against any Kafka broker that implements [KIP-447](https://cwiki.apache.org/confluence/display/KAFKA/KIP-447%3A+Producer+scalability+for+exactly+once+semantics) (Kafka 2.5+): if there are pending transactional offset commits, the broker returns `PENDING_TRANSACTION` and the client retries.

Kafka-API-compatible brokers that don't fully implement KIP-447 can instead respond with an unrelated error code (`UNKNOWN_TOPIC_OR_PARTITION` at the group/response level in the reported case), which — combined with the new group-level `resp.ErrorCode` check also added in v1.21.0 — prevents the consumer group from joining. There's no way to opt out with the current API.

## Change

- New option: `kgo.DisableRequireStableFetchOffsets() GroupOpt`. When applied, the group consumer sends `RequireStable=false` on `OffsetFetch`. Default behavior is unchanged (`RequireStable=true`, as shipped in v1.21.0).
- The one-line wire change lives in `fetchOffsets`: `req.RequireStable = !g.cfg.disableRequireStable`.
- `OptValue`/`OptValues` updated: `DisableRequireStableFetchOffsets` returns the configured bool
- The deprecated `RequireStableFetchOffsets()` godoc now points users at the new function for the opt-out path.

---

*Created with Cursor using Claude Opus 4.7*

### DIFF
--- a/pkg/kgo/client.go
+++ b/pkg/kgo/client.go
@@ -435,6 +435,8 @@ func (cl *Client) OptValues(opt any) []any {
 		return []any{cfg.group}
 	case namefn(DisableAutoCommit):
 		return []any{cfg.autocommitDisable}
+	case namefn(DisableRequireStableFetchOffsets):
+		return []any{cfg.disableRequireStable}
 	case namefn(GreedyAutoCommit):
 		return []any{cfg.autocommitGreedy}
 	case namefn(GroupProtocol):
@@ -459,7 +461,7 @@ func (cl *Client) OptValues(opt any) []any {
 	case namefn(RebalanceTimeout):
 		return []any{cfg.rebalanceTimeout}
 	case namefn(RequireStableFetchOffsets):
-		return []any{true}
+		return []any{!cfg.disableRequireStable}
 	case namefn(SessionTimeout):
 		return []any{cfg.sessionTimeout}
 
--- a/pkg/kgo/client.go
+++ b/pkg/kgo/client.go
@@ -461,7 +461,7 @@ func (cl *Client) OptValues(opt any) []any {
 	case namefn(RebalanceTimeout):
 		return []any{cfg.rebalanceTimeout}
 	case namefn(RequireStableFetchOffsets):
-		return []any{!cfg.disableRequireStable}
+		return []any{true}
 	case namefn(SessionTimeout):
 		return []any{cfg.sessionTimeout}
 
--- a/pkg/kgo/client_test.go
+++ b/pkg/kgo/client_test.go
@@ -442,9 +442,6 @@ func TestDisableRequireStableFetchOffsetsOptValue(t *testing.T) {
 	if got := cl.OptValue(DisableRequireStableFetchOffsets); got != true {
 		t.Errorf("OptValue(DisableRequireStableFetchOffsets) = %v, want true", got)
 	}
-	if got := cl.OptValue(RequireStableFetchOffsets); got != false {
-		t.Errorf("OptValue(RequireStableFetchOffsets) = %v, want false", got)
-	}
 }
 
 func TestDefaultRequireStableFetchOffsetsOptValue(t *testing.T) {
@@ -459,8 +456,5 @@ func TestDefaultRequireStableFetchOffsetsOptValue(t *testing.T) {
 
 	if got := cl.OptValue(DisableRequireStableFetchOffsets); got != false {
 		t.Errorf("OptValue(DisableRequireStableFetchOffsets) = %v, want false", got)
-	}
-	if got := cl.OptValue(RequireStableFetchOffsets); got != true {
-		t.Errorf("OptValue(RequireStableFetchOffsets) = %v, want true", got)
 	}
 }
--- a/pkg/kgo/client_test.go
+++ b/pkg/kgo/client_test.go
@@ -415,3 +415,52 @@ func TestShareGroup(t *testing.T) {
 		t.Fatalf("expected %d records, got %d", totalRecords, n)
 	}
 }
+
+func TestDisableRequireStableFetchOffsetsOpt(t *testing.T) {
+	cfg := defaultCfg()
+	if cfg.disableRequireStable {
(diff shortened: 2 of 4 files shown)


adamkasztenny:
Closing because [this](https://github.com/twmb/franz-go/issues/1312#issuecomment-4346078614) is a better solution

PR #1321, opened 2026-05-12
add support for custom AWS MSK dns names

janmoritzmeyer0210:
when working with custom DNS Names for AWS MSK, the identifyRegion function fails as the URL format is different. Therefore this library is currently not usable when working with custom DNS names, as described for example here (https://aws.amazon.com/de/blogs/big-data/configure-a-custom-domain-name-for-your-amazon-msk-cluster/). This approach uses the env var to determine the region when the region can't be determined using the url instead of failing. To avoid confusion a log message was added whenever the env variable is used

### DIFF
--- a/pkg/sasl/aws/aws.go
+++ b/pkg/sasl/aws/aws.go
@@ -125,7 +125,6 @@ func challenge(auth Auth, host string) ([]byte, error) {
 		if region == "" {
 			return nil, err
 		}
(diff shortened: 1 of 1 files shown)


twmb:
I think the log can be dropped, otherwise lgtm

janmoritzmeyer0210:
@twmb thanks, I removed the log

PR #1322, opened 2026-05-12
kgo: fix delTopics not writing back modified pausedPartitions struct

gorakdev:
## Bug

`delTopics` reads the `pausedPartitions` value from the map into a local copy, sets `all=false` on the copy, but never writes it back to the map. When partition-level pauses exist for the topic (`len(pps.m) > 0`), the `delete` branch is skipped and the modified copy is discarded — leaving `all=true` in the map. The topic remains permanently paused after `ResumeFetchTopics`.

```go
// before — pps is a copy; m[topic] is never updated
pps.all = false
if !pps.all && len(pps.m) == 0 {
    delete(m, topic)
}
```

## Fix

Write `pps` back to the map after clearing `all`, consistent with `addTopics`, `addPartitions`, and every other mutating function in this file.

```go
pps.all = false
m[topic] = pps  // write back
if !pps.all && len(pps.m) == 0 {
    delete(m, topic)
}
```

## Test

Added `TestDelTopicsWithCoexistingPartitionPauses` in `topics_and_partitions_test.go` covering the coexistence case.

### DIFF
--- a/pkg/kgo/topics_and_partitions.go
+++ b/pkg/kgo/topics_and_partitions.go
@@ -216,6 +216,7 @@ func (m pausedTopics) delTopics(topics ...string) {
 			continue
 		}
 		pps.all = false
+		m[topic] = pps
 		if !pps.all && len(pps.m) == 0 {
 			delete(m, topic)
 		}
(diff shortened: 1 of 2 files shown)


twmb:
Good fix thanks!

twmb (approved):

PR #1326, opened 2026-05-15
Fix commit-triggered group rejoin

v14dis14v:
## Problem
After a short network/coordinator disruption, the broker may lose the current consumer group member.

In this state, `OffsetCommit` can return fatal member errors (for example `UNKNOWN_MEMBER_ID` or `ILLEGAL_GENERATION`).

Before this change, franz-go mostly logged commit errors and continued until heartbeat later noticed the broken session. That left a window where workers could keep running in an invalid group session and repeatedly fail commits.

## Fix
When `OffsetCommitResponse` contains fatal member-identity errors, trigger an immediate group rejoin from the commit path.

This removes the delay and recovers the consumer session right away.

## Tests
Added behavior tests in `pkg/kfake/behavior_test.go`:
- classic group: commit `UNKNOWN_MEMBER_ID` triggers rejoin and recovery
- KIP-848 group: commit `UNKNOWN_MEMBER_ID` triggers rejoin/re-assignment and recovery
- classic group: commit `ILLEGAL_GENERATION` triggers rejoin and recovery

Also verified with `go test ./... -count=1` in `pkg/kfake`.

@twmb

### DIFF
--- a/pkg/kfake/behavior_test.go
+++ b/pkg/kfake/behavior_test.go
@@ -6064,3 +6064,182 @@ func TestShareGroupRecyclePoolAliasing(t *testing.T) {
 		}
 	}
 }
+
+// injectOffsetCommitError returns a ControlKey callback that replies to
+// every OffsetCommit with the given per-partition error code. The control
+// is one-shot: returning true pops it after first use.
+func injectOffsetCommitError(code int16) func(kmsg.Request) (kmsg.Response, error, bool) {
+	return func(kreq kmsg.Request) (kmsg.Response, error, bool) {
+		req := kreq.(*kmsg.OffsetCommitRequest)
+		resp := req.ResponseKind().(*kmsg.OffsetCommitResponse)
+		for _, t := range req.Topics {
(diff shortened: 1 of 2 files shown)


twmb:
Literally like 2 minutes after I just released 1.21.2 heh..

Did you discover this while testing the 848 path?

v14dis14v:
Yes — this came up while testing the KIP-848 path, but the issue is not 848-specific. The same commit-path gap exists for classic groups too: fatal OffsetCommit member errors were logged but did not trigger immediate rejoin.

twmb:
Fix lgtm, but I'll be waiting on this for a bit (up to a month maybe) to see if any further bugs come in. The default worst case of this is  up to 5s of consuming partitions that have been revoked, with the average case being 2.5s (and the most average case is this not being hit).

twmb:
I'll merge, in a week or two

twmb:
Actually the 848 logic has some problems, I went ahead and scoped this to only classic consumers groups in #1340 - will merge soon, closing this one.

PR #1327, opened 2026-05-18
kgo: fall back to all partitions when no writable partitions exist

ericsg666:
## Summary

When every partition of a topic temporarily has a retriable leader error (LeaderNotAvailable, NotLeaderForPartition, leader election during a rolling restart or controller failover), Produce previously failed records up front with the synthetic `unable to partition record due to no usable partitions` error, even though the sink retry and metadata refresh paths are already engineered to recover.

This mirrors the Apache Kafka Java client's `BuiltInPartitioner`: when the `writable` partition list is empty, fall back to the full partition list. The record buffers into a `recBuf` whose sink is already populated (via the existing `unknownSeedID` rewrite in `metadata.go`) and drains once a leader recovers, the same way ordinary leader rebalances are handled. If the outage outlasts `RecordDeliveryTimeout`, the record fails with the real Kafka error code so `errors.Is(err, kerr.LeaderNotAvailable)` works, instead of the synthetic string.

Paths that require partition consistency (`ManualPartitioner`, `BasicConsistentPartitioner`, keyed `StickyKeyPartitioner`) and topic-level fatal errors (deleted, unauthorized) are unchanged.

## Use Case
Identified an edge case failure occurring during a rolling restart when a topic is configured with a Replication Factor of 1 (RF=1) and all of its partitions reside on the restarting broker.

## Change

- `pkg/kgo/producer.go`: `doPartition` falls back to `partsData.partitions` when `writablePartitions` is empty and the partitioner does not require consistency.
- `pkg/kfake/producer_empty_writable_partitions_test.go`: regression tests covering the recovery path, the RF=1 single-broker restart timeline, and the bounded-fail behavior when the outage exceeds `RecordDeliveryTimeout`.

## Impact on different types of partitioners
| Partitioner | RequiresConsistency | Behavior for Keyed Records |
| :--- | :--- | :--- |
| **StickyKeyPartitioner** *(default)* | `true` when `r.Key != nil` *(partitioner.go:564)* | Unchanged — still routes through `partsData.partitions` |
| **BasicConsistentPartitioner** / **ManualPartitioner** | always `true` *(partitioner.go:109, :118)* | Unchanged |
| **UniformBytesPartitioner** *(with keys=true)* | `true` when `r.Key != nil` *(partitioner.go:317)* | Unchanged |
| **RoundRobin** / **Sticky** / **LeastBackup** | always `false` | Affected by fallback *(no key, no determinism contract)* |

## Test plan

- [x] `go vet ./pkg/kfake/... ./pkg/kgo/...`
- [x] `go build ./...`
- [x] `go test -race -run 'TestDoPartition_' ./pkg/kfake` (7 tests pass, including the no-regression check on the healthy-topic and ManualPartitioner paths)

### DIFF
--- a/pkg/kfake/producer_empty_writable_partitions_test.go
+++ b/pkg/kfake/producer_empty_writable_partitions_test.go
@@ -1,0 +1,645 @@
+package kfake
+
+// These tests cover the doPartition fallback fix in
+// pkg/kgo/producer.go. When every partition of a topic has a retriable
+// leader load error and the partitioner does not require consistency,
+// the producer must buffer records into the partition set instead of
+// failing with the synthetic error "unable to partition record due to
+// no usable partitions". Recovery is delegated to the existing sink
+// retry / migrateProductionTo / recBuf retry machinery -- the same
+// path that handles ordinary leader migrations.
+//
+// The branches of doPartition exercised here:
+//
+//   1. requiresConsistency=true, len(partitions)>0
+//      -- TestDoPartition_ManualPartitioner_AllLeadersUnavailable_StillBuffers
+//   2. requiresConsistency=false, len(writablePartitions)>0
+//      -- TestDoPartition_HealthyTopic_NoRegression
+//   3. requiresConsistency=false, len(writablePartitions)==0,
+//      len(partitions)>0           [the fix's new fallback branch]
+//      -- TestDoPartition_AllLeadersUnavailable_BuffersThenProduces
+//      -- TestDoPartition_AllLeadersUnavailable_ConcurrentProducesAllSucceed
+//      -- TestDoPartition_RF1_BrokerOutage_RecordsSurviveAndProduceAfterRecovery
+//
+// The full RF=1 single-broker rolling-restart timeline from the
+// patch doc is exercised by injecting both Metadata AND Produce
+// errors for a configurable outage window:
+//   -- TestDoPartition_RF1_BrokerOutage_RecordsSurviveAndProduceAfterRecovery
+//      (outage < RecordDeliveryTimeout -- records survive and produce)
+//   -- TestDoPartition_OutageExceedsRecordTimeout_FailsWithRealKafkaError
+//      (outage > RecordDeliveryTimeout -- records fail with the real
+//      Kafka error, NOT the synthetic "no usable partitions" string)
+//
+// The final outer guard at producer.go's
+// `if len(mapping) == 0` -> synthetic error is unreachable from real
+// callers post-fix: partitionsForTopicProduce filters out the
+// len(partitions)==0 case before calling doPartition, and
+// storePartitionsUpdate has the same guard at topics_and_partitions.go.
+// There is no realistic kfake setup that hits the synthetic error
+// after the fix.
+
+import (
+	"context"
+	"errors"
+	"net"
+	"strconv"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/twmb/franz-go/pkg/kerr"
+	"github.com/twmb/franz-go/pkg/kgo"
+	"github.com/twmb/franz-go/pkg/kmsg"
+)
+
+// metadataNoLeaderResp builds a MetadataResponse for testTopic that
+// reports LeaderNotAvailable on every partition. The broker remains
+// reachable so subsequent (non-injected) metadata calls return real
+// data, simulating recovery from a brief leader unavailability.
+func metadataNoLeaderResp(
+	t *testing.T,
+	c *Cluster,
+	testTopic string,
+	nPartitions int,
+	req *kmsg.MetadataRequest,
+) *kmsg.MetadataResponse {
+	t.Helper()
+	resp := req.ResponseKind().(*kmsg.MetadataResponse)
+
+	host, portStr, _ := net.SplitHostPort(c.ListenAddrs()[0])
+	port, _ := strconv.Atoi(portStr)
+	sb := kmsg.NewMetadataResponseBroker()
+	sb.NodeID = 0
+	sb.Host = host
+	sb.Port = int32(port)
+	resp.Brokers = append(resp.Brokers, sb)
+	resp.ControllerID = 0
+
+	ti := c.TopicInfo(testTopic)
+	st := kmsg.NewMetadataResponseTopic()
+	st.Topic = kmsg.StringPtr(testTopic)
+	st.TopicID = ti.TopicID
+	for p := 0; p < nPartitions; p++ {
+		sp := kmsg.NewMetadataResponseTopicPartition()
+		sp.Partition = int32(p)
+		sp.ErrorCode = kerr.LeaderNotAvailable.Code
+		sp.Leader = -1
+		sp.LeaderEpoch = 0
+		sp.Replicas = []int32{0}
+		sp.ISR = []int32{0}
+		st.Partitions = append(st.Partitions, sp)
+	}
+	resp.Topics = append(resp.Topics, st)
+	return resp
+}
+
+// injectAllLeadersUnavailableForNTopicRequests injects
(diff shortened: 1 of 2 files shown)


twmb:
I guess I don't really see much of a downside to this change, I'm not entirely convinced on the need, but the PR kinda unifies the behavior for keyed and unkeyed records, so that's good.

I admire the kfake test but I try to keep every individual kfake unit test to <2s. I'll see if claude can post a comment on a line / some lines to suggest speeding it up - but I'd probably nix the test since the change is very small and straightforward.

Can you also simplify the conditional?
```go
if parts.partitioner.RequiresConsistency(pr.Record) || len(mapping) == 0 {
```

twmb:
From claude:

The producer.go change traces out correctly: both `doPartition` call sites guarantee `len(partitions) > 0`, so the synthetic error fires exactly when every partition has a retriable load error and the partitioner does not require consistency. A keyed record in the identical situation already takes `mapping = partsData.partitions` and buffers fine, so this also removes an inconsistency between keyed and unkeyed records, and the recovery machinery (failing recBufs skipped by the sink, `bumpRepeatedLoadErr` on refresh, migrate + clearFailing on recovery) genuinely handles the rest. Two suggestions:

## 1. The fix can be one condition

```go
mapping := partsData.writablePartitions
if parts.partitioner.RequiresConsistency(pr.Record) || len(mapping) == 0 {
	mapping = partsData.partitions
}
```

The `len(partsData.partitions) > 0` check is redundant: if partitions is also empty, falling back to it hits the same `len(mapping) == 0` guard below with identical behavior.

## 2. The test file can shrink to ~110 lines / 2 tests, and run in ~1.2s instead of ~9s

Per test:

- **`HealthyTopic_NoRegression`**: every successful produce in the package already covers this. Drop.
- **`ManualPartitioner_...`**: the `RequiresConsistency` arm is untouched by this diff (it already used `partsData.partitions`). Drop.
- **`ConcurrentProducesAllSucceed`**: the new branch runs under the same per-topic `partsMu` as every pre-existing `doPartition` path; there is no new synchronization to stress, so this tests the lock, not the fix. Drop.
- **`SyntheticErrorNotEmittedDuringRecovery`**: its assertion (`r.Err != nil && strings.Contains(...)`) only fails on the synthetic string, so any *other* per-record error passes silently. Folding it into the recovery test with a plain `r.Err != nil` check is strictly stronger. Merge.
- **`RF1_BrokerOutage_...`**: the produce-rejection leg (produce goes out, gets NotLeaderForPartition, `handleRetryBatches` sets the recBuf failing, recovery) exercises pre-existing sink-retry machinery - it is how every ordinary leader migration already works. The changed code runs at buffer time and cannot tell whether the recBuf pre-existed. The 3s wall-clock window also has a quiet failure mode: on a slow CI machine the records can buffer after the outage already lapsed, and the test passes while testing nothing. Drop (or event-gate the window if you want to keep it).
- **`OutageExceedsRecordTimeout_...`**: keep, but it can be tightened. Because metadata is leaderless from the very first load, the recBufs are failing from birth and the sink never drains them - the Produce control in this test never fires, so `produceErrResp` can be deleted. That also makes the failure path deterministic: `bumpRepeatedLoadErr` delivers the metadata load error, always `kerr.LeaderNotAvailable`, so the `errors.Is` check can be a hard assert instead of a `t.Logf`. `RecordDeliveryTimeout` can drop to 1s (the validated minimum).

Other speed/style notes: `MetadataMinAge` can be 10ms (the validated floor; the metadata retry loop waits `min(metadataMinAge, 250ms)`, so this directly drives recovery speed), `t.Parallel()` on everything (each test has its own cluster), and the cluster/client boilerplate can use the existing `newCluster`/`newPlainClient` helpers in `helpers_test.go`. The injection helper's returned `*atomic.Int32` is never read by any test and can go; controls run serially on the Cluster.run goroutine, so a plain int counter suffices.

Proposed replacement file:

```go
// injectLeaderless makes the first n topic-specific metadata responses
// (all of them if n < 0) report LeaderNotAvailable on every partition.
// The real broker list is included so the client keeps its connections.
func injectLeaderless(c *Cluster, topic string, parts int32, n int) {
	var injected int
	c.ControlKey(int16(kmsg.Metadata), func(kreq kmsg.Request) (kmsg.Response, error, bool) {
		c.KeepControl()
		req := kreq.(*kmsg.MetadataRequest)
		if len(req.Topics) == 0 || n >= 0 && injected >= n {
			return nil, nil, false
		}
		injected++
		resp := req.ResponseKind().(*kmsg.MetadataResponse)
		host, portStr, _ := net.SplitHostPort(c.ListenAddrs()[0])
		port, _ := strconv.Atoi(portStr)
		b := kmsg.NewMetadataResponseBroker()
		b.NodeID, b.Host, b.Port = 0, host, int32(port)
		resp.Brokers = append(resp.Brokers, b)
		rt := kmsg.NewMetadataResponseTopic()
		rt.Topic = kmsg.StringPtr(topic)
		rt.TopicID = c.TopicInfo(topic).TopicID
		for p := int32(0); p < parts; p++ {
			rp := kmsg.NewMetadataResponseTopicPartition()
			rp.Partition = p
			rp.ErrorCode = kerr.LeaderNotAvailable.Code
			rp.Leader = -1
			rt.Partitions = append(rt.Partitions, rp)
		}
		resp.Topics = append(resp.Topics, rt)
		return resp, nil, true
	})
}

// Records produced while every partition is leaderless must buffer and
// then produce once metadata recovers, rather than failing with the
// synthetic "no usable partitions" error.
func TestProduceAllPartitionsLeaderless(t *testing.T) {
	t.Parallel()
	const topic = "t"
	c := newCluster(t, NumBrokers(1), SeedTopics(3, topic))
	injectLeaderless(c, topic, 3, 2)
	cl := newPlainClient(t, c,
		kgo.DefaultProduceTopic(topic),
		kgo.MetadataMinAge(10*time.Millisecond),
	)
	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
	defer cancel()
	recs := make([]*kgo.Record, 16)
	for i := range recs {
		recs[i] = kgo.StringRecord(strconv.Itoa(i))
	}
	for i, r := range cl.ProduceSync(ctx, recs...) {
		if r.Err != nil {
			t.Errorf("record %d should buffer through the leaderless window and produce on recovery: %v", i, r.Err)
		}
	}
}

// If the outage outlasts RecordDeliveryTimeout, the record must fail
// with the real Kafka error so errors.Is works, not the synthetic
// string.
func TestProduceAllPartitionsLeaderlessTimeout(t *testing.T) {
	t.Parallel()
	const topic = "t"
	c := newCluster(t, NumBrokers(1), SeedTopics(3, topic))
	injectLeaderless(c, topic, 3, -1) // never recovers
	cl := newPlainClient(t, c,
		kgo.DefaultProduceTopic(topic),
		kgo.MetadataMinAge(10*time.Millisecond),
		kgo.RecordDeliveryTimeout(time.Second), // validated minimum
	)
	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
	defer cancel()
	err := cl.ProduceSync(ctx, kgo.StringRecord("doomed")).FirstErr()
	if !errors.Is(err, kerr.LeaderNotAvailable) {
		t.Errorf("want the real load error after RecordDeliveryTimeout, got: %v", err)
	}
}
```

The first record of the first test necessarily transits `doPartition` while `writablePartitions` is empty, which fails immediately with the synthetic error pre-fix - so the test fails before the change and passes after it, which is the whole job. Nothing else in the current file pins behavior the diff introduces.

twmb:
I went ahead and scoped it to just the line in #1340

ericsg666:
Hi @twmb 

Thanks for the review comments and feedback
This fix can help resolve one of our producer errors (corner case, when all partitions are in the same broker and RF=1) on production 

glad to see you take the proposal, look forward to the new release 

thanks

PR #1329, opened 2026-05-25
kgo: isolate metadata cache from broker response

### LINKED ISSUE #1328: Race condition in metadata cache shared Partitions array
We are trying to update `kgo` and `kadm` to latest releases in [grafana/mimir][1] but our tests are failing with a race condition ([example][2]). After spending some time investigating the recent changes, in kgo, I think, the race stems from how metadata cache shares the `Partitions` array.

That is, the `kgo` client runs an internal metadata refresh. This races with user-level concurrent calls from `kadm`, that go through the metadata cache. This boils down to this (from current HEAD):

```
Write at metadata.go:710  — sort.Slice(topicMeta.Partitions, ...)
                             in fetchTopicMetadata (metadata loop goroutine)

Read  at client.go:1493   — for _, p := range t.Partitions
                             in RequestCachedMetadata.dupt (caller goroutine)
```

Fixing it naively doesn't seem complicated, e.g. by copying the mentioned `Partitions` slice. But this will obviously come with a trade-off, and I don't have a full mental model for which code paths are hotter than the other. Thus, I went with opening an issue, rather a PR.

### How to reproduce

_AI disclosure: the test was built with the help from Claude (Opus 4.6, if that's matter)_.

```go
// Run with -race to detect.
func TestStoreCachedMetaRace(t *testing.T) {
	t.Parallel()

	const topic = "test-topic"
	const numPartitions = 10

	c, err := NewCluster(
		NumBrokers(1),
		SeedTopics(numPartitions, topic),
	)
	if err != nil {
		t.Fatal(err)
	}
	defer c.Close()

	ti := c.TopicInfo(topic)
	host, portStr, _ := net.SplitHostPort(c.ListenAddrs()[0])
	port, _ := strconv.Atoi(portStr)

	c.ControlKey(int16(kmsg.Metadata), func(kreq kmsg.Request) (kmsg.Response, error, bool) {
		c.KeepControl()

		resp := kreq.(*kmsg.MetadataRequest).ResponseKind().(*kmsg.MetadataResponse)
		sb := kmsg.NewMetadataResponseBroker()
		sb.NodeID = 0
		sb.Host = host
		sb.Port = int32(port)
		resp.Brokers = append(resp.Brokers, sb)
		resp.ControllerID = 0

		st := kmsg.NewMetadataResponseTopic()
		st.Topic = kmsg.StringPtr(topic)
		st.TopicID = ti.TopicID

		// Build partitions in reverse order so sort.Slice in fetchTopicMetadata actually swaps elements.
		for p := int32(numPartitions - 1); p >= 0; p-- {
			sp := kmsg.NewMetadataResponseTopicPartition()
			sp.Partition = p
			sp.Leader = -1
			sp.Replicas = []int32{0}
			sp.ISR = []int32{0}
			st.Partitions = append(st.Partitions, sp)
		}
		resp.Topics = append(resp.Topics, st)
		return resp, nil, true
	})

	cl, err := kgo.NewClient(
		kgo.SeedBrokers(c.ListenAddrs()...),
		kgo.MetadataMinAge(10*time.Millisecond),
		kgo.MetadataMaxAge(50*time.Millisecond),
	)
	if err != nil {
		t.Fatal(err)
	}
	defer cl.Close()

	adm := kadm.NewClient(cl)

	ctx, cancel := context.WithTimeout(t.Context(), time.Second)
	defer cancel()

	var wg sync.WaitGroup
	for range 3 {
		wg.Add(1)
		go func() {
			defer wg.Done()
			for ctx.Err() == nil {
				adm.Metadata(ctx, topic)
			}
		}()
	}

	wg.Add(1)
	go func() {
		defer wg.Done()
		for ctx.Err() == nil {
			cl.ForceMetadataRefresh()
			time.Sleep(10 * time.Millisecond)
		}
	}()

	wg.Wait()
}
```

[1]: https://github.com/grafana/mimir/pull/15428
[2]: https://github.com/grafana/mimir/actions/runs/26399648603/job/77708651442?pr=15428

twmb:
## Summary

- Fixes the data race reported in #1328: `storeCachedMeta` cached slice and `*string` references from the broker response, so the in-place `sort.Slice(topicMeta.Partitions, ...)` in `fetchTopicMetadata` (metadata.go:710) raced concurrent readers iterating the cached entry via `kadm.Metadata` => `RequestCachedMetadata` (`dupt` at client.go:1493).
- Deep-clones on PUT so the cache owns fully independent state: `storeCachedMeta` clones `Topic *string`, `Partitions`, and each partition's `Replicas`/`ISR`/`OfflineReplicas`; `newBroker` clones `Rack *string`; `updateMetadataBrokers` clones `ClusterID *string`.
- Extends `dups(b.meta.Rack)` so the GET-side broker loop in `RequestCachedMetadata` is symmetric with the rest of `dupp`/`dupt`. `UnknownTags` is intentionally left shared - no idiomatic write path exists for it.

## Test plan

- [x] `TestIssue1328` in `pkg/kfake` reproduces the exact race the issue reports: without the fix, `-race` fires on `RequestCachedMetadata.func3` reading `t.Partitions` vs `sort.Slice` writing it.
- [x] With the fix applied, `TestIssue1328` passes 10x under `-race`.
- [x] `TestIssue1328`, `TestRequestCachedMetadata`, `TestKadmCachedMetadata`, `TestFirstMetadataPartitionErrors`, `TestIssue1142`, `TestIssue1167`: 3x under `-race`, all pass.
- [x] `go build`, `go vet`, `gofmt` clean across `pkg/kgo` and `pkg/kfake`.

Closes #1328

### DIFF
--- a/pkg/kfake/issues_test.go
+++ b/pkg/kfake/issues_test.go
@@ -3207,3 +3207,104 @@ func TestIssue1330(t *testing.T) {
 		}
 	}
 }
+
+// TestIssue1328 reproduces a data race in the metadata cache.
+//
+// Sequence of events:
+//   - The metadata loop calls fetchMetadata, which caches the broker
+//     response by storing a reference to topic.Partitions in metaCache.
+//   - fetchMetadata returns; fetchTopicMetadata then sort.Slice's the
+//     response's Partitions in place to validate strictly-increasing
+//     partition IDs.
+//   - Concurrently, a caller (kadm.Metadata => RequestCachedMetadata)
+//     reads the cached topic. resolveTopicMeta copies the cached entry
+//     under metaCache.mu, but the copy still shares the underlying
+//     Partitions slice. The caller then iterates that slice without the
+//     lock, racing the in-flight sort.
+//
+// The fix is to clone Partitions when storing into the cache so that
+// later mutations of the broker response do not touch the cached slice.
+func TestIssue1328(t *testing.T) {
+	t.Parallel()
+	const (
+		topic         = "test-topic"
+		numPartitions = 10
+	)
+
+	c, err := NewCluster(
+		NumBrokers(1),
+		SeedTopics(numPartitions, topic),
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer c.Close()
+
+	ti := c.TopicInfo(topic)
+	host, portStr, _ := net.SplitHostPort(c.ListenAddrs()[0])
+	port, _ := strconv.Atoi(portStr)
+
+	// Build partitions in reverse order so sort.Slice in fetchTopicMetadata
+	// actually swaps elements -- already-sorted input sorts without writes,
+	// and without a write there is no race to detect.
+	c.ControlKey(int16(kmsg.Metadata), func(kreq kmsg.Request) (kmsg.Response, error, bool) {
+		c.KeepControl()
+
+		resp := kreq.(*kmsg.MetadataRequest).ResponseKind().(*kmsg.MetadataResponse)
+		sb := kmsg.NewMetadataResponseBroker()
+		sb.NodeID = 0
+		sb.Host = host
+		sb.Port = int32(port)
+		resp.Brokers = append(resp.Brokers, sb)
+		resp.ControllerID = 0
+
+		st := kmsg.NewMetadataResponseTopic()
+		st.Topic = kmsg.StringPtr(topic)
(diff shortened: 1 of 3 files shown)


narqo:
> Closes https://github.com/twmb/franz-go/issues/1328

Just to let you know, I've tested the changes from this branch in https://github.com/grafana/mimir/pull/15428, and, at least from Mimir's own tests, the original race has been fixed.

PR #1335, opened 2026-06-02
kgo: add RebootstrapTrigger for timeout-based rebootstrap (KIP-1102)

david-yu:
## Summary

Adds `RebootstrapTrigger(d time.Duration)`, the **timeout-based** rebootstrap trigger described in [KIP-1102](https://cwiki.apache.org/confluence/display/KAFKA/KIP-1102%3A+Enable+clients+to+rebootstrap+based+on+timeout+or+error+code).

franz-go already implements the KIP-1102 **error-code** trigger via `OnRebootstrapRequired` (acting on `REBOOTSTRAP_REQUIRED`). KIP-1102 defines a second trigger — a timeout — which is the Java client's `metadata.recovery.rebootstrap.trigger.ms`. This PR adds that counterpart.

## Behavior

`RebootstrapTrigger(d)` sets a duration after which, if the client has been unable to obtain a **successful metadata response** for that long, it rebootstraps from its seed brokers (re-resolving their addresses) and reconnects.

- `0` (default) disables the behavior — existing clients are unaffected.
- If `OnRebootstrapRequired` is set, it is consulted for the seeds to use; otherwise the originally configured `SeedBrokers` are re-resolved (useful for DNS-based failover, where the seed addresses repoint to a different cluster).

I kept the option opt-in (default off) rather than mirroring the Java default of 5m, since franz-go already periodically queries a seed broker to recover from the "all known brokers down" case — so this is additive and non-behavior-changing unless explicitly enabled. Happy to change the default if you'd prefer KIP parity.

## Implementation

- `config.go`: new `rebootstrapTrigger` cfg field + `RebootstrapTrigger` option.
- `client.go`: `OptValues` case for the new option.
- `metadata.go`: `updateMetadataLoop` tracks the time of the last successful update; on continued failure past the trigger it calls a new `rebootstrap()` helper, then resets the window. The helper selects seeds (callback vs. configured), calls `UpdateSeedBrokers`, and drops current connections via `updateBrokers(nil)` — the same primitives the error-code path already uses.

## Tests

`rebootstrap_test.go` (no cluster required):
- option wiring (`RebootstrapTrigger` → cfg)
- `rebootstrap()` re-seeds from `OnRebootstrapRequired` when set
- `rebootstrap()` re-resolves the configured `SeedBrokers` when no callback is set

`go build ./...`, `go vet ./pkg/kgo/`, and `go test -run TestRebootstrap ./pkg/kgo/` all pass.

## Notes / open questions

- I left `CHANGELOG.md` untouched — happy to add an entry under your preferred section.
- The timeout trigger reuses `OnRebootstrapRequired` for seed selection when present, which means that callback can now also fire on the timeout path (not just `REBOOTSTRAP_REQUIRED`). I think that's the desirable single-source behavior, but I can split them if you'd rather keep the callback strictly tied to the error code.
- Motivation: enabling seamless failover between Redpanda shadow/primary clusters in Redpanda Connect, where the broker may not always emit `REBOOTSTRAP_REQUIRED`.

### DIFF
--- a/pkg/kgo/client.go
+++ b/pkg/kgo/client.go
@@ -319,6 +319,8 @@ func (cl *Client) OptValues(opt any) []any {
 		return []any{cfg.missingTopicDelete}
 	case namefn(OnRebootstrapRequired):
 		return []any{cfg.onRebootstrapRequired}
+	case namefn(RebootstrapTrigger):
+		return []any{cfg.rebootstrapTrigger}
 	case namefn(WithContext):
 		return []any{cfg.ctx}
 	case namefn(DisableClientMetrics):
--- a/pkg/kgo/config.go
+++ b/pkg/kgo/config.go
@@ -93,6 +93,7 @@ type cfg struct {
 	minVersions *kversion.Versions
 
 	onRebootstrapRequired func() ([]string, error)
+	rebootstrapTrigger    time.Duration
 
 	retryBackoff func(int) time.Duration
 	retries      int64
@@ -1008,6 +1009,29 @@ func ConsiderMissingTopicDeletedAfter(t time.Duration) Opt {
 // You can read KIP-1102 for more info about this option.
 func OnRebootstrapRequired(fn func() ([]string, error)) Opt {
 	return clientOpt{func(cfg *cfg) { cfg.onRebootstrapRequired = fn }}
+}
+
+// RebootstrapTrigger sets a duration after which, if the client has been unable
+// to obtain a successful metadata response for that long, the client
+// rebootstraps using its seed brokers (re-resolving them) and reconnects. A
+// value of 0 (the default) disables timeout based rebootstrapping.
+//
+// This is the timeout based rebootstrap trigger described in KIP-1102, the
+// counterpart to the error code trigger handled by [OnRebootstrapRequired].
+// Whereas the error code trigger relies on the broker returning
+// REBOOTSTRAP_REQUIRED, this trigger lets the client rebootstrap on its own
+// when its currently known brokers all become unreachable, mirroring the Java
+// client's metadata.recovery.rebootstrap.trigger.ms.
+//
+// Note that franz-go already periodically queries a seed broker to recover from
+// the scenario where all previously discovered brokers are unavailable, so this
+// option is primarily useful when you want a bounded, configurable window after
+// which a full rebootstrap (including re-resolving seed broker addresses, e.g.
+// for DNS based failover) is forced. If [OnRebootstrapRequired] is set, it is
+// consulted for the seeds to use; otherwise the originally configured
+// [SeedBrokers] are re-resolved.
+func RebootstrapTrigger(d time.Duration) Opt {
+	return clientOpt{func(cfg *cfg) { cfg.rebootstrapTrigger = d }}
 }
 
 // DisableClientMetrics opts out of collecting and sending client metrics to
--- a/pkg/kgo/metadata.go
+++ b/pkg/kgo/metadata.go
@@ -187,6 +187,7 @@ func (cl *Client) updateMetadataLoop() {
 	defer close(cl.metadone)
 	var consecutiveErrors int
 	var lastAt time.Time
+	lastSuccess := time.Now()
 
 	ticker := time.NewTicker(cl.cfg.metadataMaxAge)
 	defer ticker.Stop()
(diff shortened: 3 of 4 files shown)


david-yu:
### Runtime verification ✅

Tested this branch end-to-end against a real broker (single-node Redpanda in Docker) to confirm the **timeout** trigger fires on its own — no `REBOOTSTRAP_REQUIRED` involved.

**Setup:** a consumer client seeded with a **dead** broker (`127.0.0.1:9099`), `MetadataMaxAge(1s)`, `RebootstrapTrigger(3s)`, and `OnRebootstrapRequired` returning the **live** broker (`127.0.0.1:9092`) — i.e. a primary→shadow failover where the broker never signals a rebootstrap.

**Result:** the client failed against the dead seed, then after the 3s trigger window it rebootstrapped itself onto the live broker and recovered (Ping OK) in ~4.6s total:

```
Client started with dead seed 127.0.0.1:9099; rebootstrap trigger = 3s
Initial Ping (expected to FAIL against dead seed): err=unable to dial: dial tcp 127.0.0.1:9099: connect: connection refused

Waiting for the timeout-based rebootstrap to kick in...
    [WARN] unable to open connection to broker [addr 127.0.0.1:9099 broker seed_0 err dial tcp ...: connection refused]
    ... (repeated metadata failures) ...
>>> OnRebootstrapRequired called (#1) -> returning live broker 127.0.0.1:9092
    [INFO] rebootstrapped from seed brokers after metadata was unavailable for longer than the rebootstrap trigger [trigger 3s]

SUCCESS after 4.6s:
  - rebootstrap log emitted:           true
  - OnRebootstrapRequired call count:  1
  - Ping to live broker after rebootstrap: OK
```

```
  cl, err := kgo.NewClient(
      kgo.SeedBrokers("127.0.0.1:9099"),            // a DEAD broker (nothing listening) — the "primary"
      kgo.MetadataMinAge(500*time.Millisecond),     // lowered so the metadata loop retries quickly
      kgo.MetadataMaxAge(time.Second),              // drives frequent metadata attempts (so failures accrue fast)
      kgo.RebootstrapTrigger(3*time.Second),        // ← the new option under test (KIP-1102 timeout trigger)
      kgo.OnRebootstrapRequired(func() ([]string, error) {
          // Called by the timeout trigger; returns the LIVE broker — the "shadow"
          return []string{"127.0.0.1:9092"}, nil
      }),
      kgo.ConsumeTopics("rebootstrap-demo"),        // gives the metadata loop a reason to actively fetch
      kgo.WithLogger(logger),                       // captured logs to detect the rebootstrap line
  )
``` 

So the timeout trigger correctly: (1) waits the configured window, (2) consults `OnRebootstrapRequired` for seeds, (3) re-seeds via `UpdateSeedBrokers`, and (4) reconnects to the new cluster. The callback fired exactly once.

(Demo harness was throwaway and isn't part of the PR; the committed unit tests cover the option wiring and seed-selection logic without a cluster.)

david-yu:
### Follow-up: where the timeout trigger does / doesn't fire

To sanity-check the practical value, I ran failover tests against two live single-node Redpanda clusters (primary + shadow), measuring recovery via a real **produce** to the surviving cluster.

**Topology 1 — both clusters statically listed in `SeedBrokers`:**

| Config | Failed over? | `OnRebootstrapRequired` calls |
|---|---|---|
| error-code only (`OnRebootstrapRequired`) | ✅ ~300ms | 0 |
| timeout trigger (`RebootstrapTrigger`, this PR) | ✅ ~0s | 0 |
| baseline (no rebootstrap opts) | ✅ ~0s | 0 |

**Topology 2 — single stable DNS endpoint (hostname seed repointed to shadow on failover):**

| Config | Failed over? | callback calls | seed re-dialed → shadow |
|---|---|---|---|
| error-code only | ✅ ~200ms | 0 | 1 |
| timeout trigger (this PR) | ✅ ~800ms | 0 | 1 |
| baseline | ✅ ~800ms | 0 | 1 |

In both topologies, recovery happened via franz-go's **built-in periodic seed re-query** — the seed was re-dialed / the DNS hostname re-resolved to the surviving cluster automatically, in well under a second. `RebootstrapTrigger` never fired in those runs because built-in recovery completed long before the trigger window.

The timeout trigger **does** fire and recover correctly in the scenario where built-in recovery can't help on its own — i.e. when `OnRebootstrapRequired` returns a **different** seed set than originally configured (the earlier dead-seed→new-address demo above). That's its real niche: forcing a re-seed/re-resolve after a bounded window, especially when the new seeds aren't already in the client's seed list.

Sharing for context — it may be worth documenting in the option's godoc that for seed lists / DNS seeds that already resolve to the failover target, franz-go's existing seed re-query covers failover without this option, so `RebootstrapTrigger` is most useful when (a) you want a bounded forced re-resolve, or (b) the callback supplies new seeds. Happy to adjust the doc comment accordingly if you'd like.

david-yu:
Will go ahead and close.

Issue #908, opened 2025-02-18
Handle topic recreation while producing & consuming


twmb:

twmb:
This is only possible in a topic ID world. If we receive UNKNOWN_TOPIC_ID while consuming, set topic state to "deleted". On metadata update, if topic has state deleted, _allow_ the topic to be replaced.

brianshih1:
@twmb any updates on this issue? 

In this earlier issue (https://github.com/twmb/franz-go/issues/676) you mentioned that you ran into odd internal race conditions. Is that the biggest blocker to allow consumers to consume recreated topics when it sees an `UNKNOWN_TOPIC_ID` error? I also saw this PR from a while ago (https://github.com/twmb/franz-go/pull/377) that seems reverted.

twmb:
I completely forgot I tried implementing this before, and your link to my prior issue reminds me why I backed out...

https://github.com/twmb/franz-go/pull/391/commits/9a8f04a63bb43e8841084599a01d1689f3467be4

If OffsetForLeaderEpoch notices a recreated topic _before_ a metadata update / before a fetch, the code detects this as data loss. I think that properly fixing this would require the client to stall when handling data loss detection, issue metadata request when data loss is detected, then check if the topic ID changed, and only then continue the actual logic...

The real problem is that Confluent never actually implemented all of [KIP-516](https://issues.apache.org/jira/browse/KAFKA-8872). Topic names (as strings) are still the only option in many Kafka protocol APIs. You can see OffsetForLeaders (i.e. OffsetForLeaderEpoch) is still unimplemented in the [tracking epic](https://issues.apache.org/jira/browse/KAFKA-8872).

Until this is implemented everywhere, handling this within the client requires more heuristics than I'd like. It looks like Confluent is going to add TopicIDs to produce requests in 4.1, but it looks like they're still ignoring the OffsetForLeaderEpoch implementation.

Rather than adding heuristics in the client and essentially guessing at times if a topic was recreated / errors should be ignored, I'm going to close this issue and leave the decision to the end user. It's not really _that_ common that a person deletes and recreates a topic _while_ clients are active _and_ that things should continue working. The escape hatch I provided back in the day was `PurgeTopicsFromClient`: if you know you deleted a topic, you can purge it from the client and then resume producing to it (which will discover the topic anew) or, as a consumer, `AddConsumeTopics`.

twmb:
## Design constraints for a zero-loss topic-recreation fix (from an internal audit)

[...]

PR #1336, opened 2026-06-03
kgo: re-resolve topic ID when a known topic's ID changes (recover from UNKNOWN_TOPIC_ID loop)

david-yu:
## Problem

When fresh metadata reports a **different topic ID for a name the client already knows**, the current id2t merge keeps the **stale** ID and ignores the new one until the user calls `PurgeTopicsFromClient`:

```go
// Only update if it's the same ID (normal case), skip if it's a
// different ID (recreated topic).
if _, sameID := merged[mt.id]; sameID {
    merged[mt.id] = mt.topic
}
continue
```

A topic's ID changes in two real scenarios:
1. the topic is **deleted and recreated**, and
2. a client configured with **multiple seed brokers fails over to a different cluster** where the same topic name has a different topic ID (e.g. Redpanda **ShadowLink** cross-region DR / a "super-seed" multi-bootstrap pattern).

In both cases a long-lived **consumer loops indefinitely on `UNKNOWN_TOPIC_ID`** because it keeps fetching with the old ID, and never recovers without an explicit `PurgeTopicsFromClient`.

## Change

On a metadata merge, when a known topic **name** reports a **new ID**, adopt the new ID (dropping the stale `ID -> name` entries so exactly one ID is kept per name) instead of pinning the old one. This lets a long-lived consumer **re-resolve and recover automatically** across a recreation / cross-cluster failover.

## How it was found / tested

Hit while testing Redpanda cross-region DR failover ("super-seed": clients list both clusters' seeds, fail over on outage). The **producer** path fails over fine (franz-go re-queries seeds), but a **long-lived consumer** riding through the promote got stuck looping `UNKNOWN_TOPIC_ID: This server does not host this topic ID` (the promoted topic on the secondary cluster has a different topic ID). A *fresh* consumer worked; the long-lived one did not. With this change the long-lived consumer re-resolves the ID and resumes from the secondary.

End-to-end validated against a custom Redpanda Connect image built with this franz-go (2 EKS clusters, us-west-2 → us-east-1, ShadowLink).

## Notes

- Draft — opening for discussion. This intentionally changes the documented "keep stale ID until purge" behavior; happy to gate it behind an option if preferred.
- `go build ./pkg/kgo/` passes; `go vet` clean.

### DIFF
--- a/pkg/kgo/metadata.go
+++ b/pkg/kgo/metadata.go
@@ -431,12 +431,27 @@ func (cl *Client) updateMetadata() (retryWhy multiUpdateWhy, err error) {
 			}
 			if _, exists := knownNames[mt.topic]; exists {
 				// This name already has an ID in the map.
-				// Only update if it's the same ID (normal
-				// case), skip if it's a different ID
-				// (recreated topic).
 				if _, sameID := merged[mt.id]; sameID {
+					// Same ID (normal case): refresh mapping.
 					merged[mt.id] = mt.topic
+					continue
 				}
+				// A different ID for a name we already know means the
+				// topic's ID changed: the topic was deleted+recreated,
+				// or (super-seed DR) the client failed over to another
+				// cluster where the same topic name has a different ID.
+				// Previously we pinned the stale ID until the user called
+				// PurgeTopicsFromClient, which makes consumers loop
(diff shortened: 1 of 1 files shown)


twmb:
https://github.com/twmb/franz-go/issues/908

PR #1337, opened 2026-06-04
kprom: enable native histograms for all histogram metrics

pracucci:
Native (sparse) Prometheus histograms are more accurate than classic histograms: they capture a value's magnitude with high resolution across a wide dynamic range, instead of bucketing into a fixed, pre-chosen set of boundaries.

This matters in particular for kprom because the classic histograms are configured with `DefBuckets`, whose largest bucket is `2.048s`. Anything slower than ~2s all lands in the `+Inf` bucket, so quantile estimates above that range are effectively meaningless — and in practice Kafka timings (writes, reads, end-to-end request duration, throttling) can and do exceed 2s. Native histograms have no such ceiling: their buckets grow geometrically to cover whatever range the observed values fall into, so high latencies remain measurable without anyone having to tune bucket boundaries up front.

In this PR, I propose to change all histograms to emit both classic and native representations, so existing dashboards and recording rules keep working unchanged.

### DIFF
--- a/plugin/kprom/kprom.go
+++ b/plugin/kprom/kprom.go
@@ -150,13 +150,16 @@ func (m *Metrics) OnNewClient(client *kgo.Client) {
 
 	// newHistogramOpts builds HistogramOpts for histogram h: classic buckets
 	// (custom if configured, else default) plus the shared native-histogram
-	// parameters. Setting NativeHistogramBucketFactor while keeping Buckets makes
-	// each histogram emit both classic and native representations, preserving
-	// backward compatibility.
+	// parameters. By default both classic and native histograms are emitted;
+	// DisableClassicHistograms drops the classic buckets and a native bucket
+	// factor of 1 or less disables native histograms (see NativeHistogramBucketFactor).
 	newHistogramOpts := func(h Histogram, name, help string) prometheus.HistogramOpts {
-		buckets := m.cfg.defBuckets
-		if b, ok := m.cfg.histograms[h]; ok && len(b) != 0 {
-			buckets = b
+		var buckets []float64
+		if !m.cfg.disableClassicHistograms {
+			buckets = m.cfg.defBuckets
+			if b, ok := m.cfg.histograms[h]; ok && len(b) != 0 {
+				buckets = b
+			}
 		}
 		return prometheus.HistogramOpts{
 			Namespace:                       namespace,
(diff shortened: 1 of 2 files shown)


replay (approved):

plugin/kprom/kprom.go
```diff
@@ -148,12 +148,27 @@ func (m *Metrics) OnNewClient(client *kgo.Client) {
...
 
-	// returns Hist buckets if set, otherwise defBucket
-	getHistogramBuckets := func(h Histogram) []float64 {
-		if buckets, ok := m.cfg.histograms[h]; ok && len(buckets) != 0 {
-			return buckets
+	// newHistogramOpts builds HistogramOpts for histogram h: classic buckets
+	// (custom if configured, else default) plus the shared native-histogram
+	// parameters. Setting NativeHistogramBucketFactor while keeping Buckets makes
+	// each histogram emit both classic and native representations, preserving
+	// backward compatibility.
```
twmb: We could probably add an opt out option for users that only want native, but I'm fine to keep both for if you prefer that. As well, we could probably add options for the native hist bucket factor / number / reset duration.

twmb (commented):
Let me know if you'd prefer adding options otherwise I'm +1 on this / good to release it.

twmb (dismissed):

plugin/kprom/kprom.go
pracucci: Thanks for the review. Let me take a look at this feedback, please.

pracucci: Done in 32ab9f00. Both histogram types remain enabled by default, and I added options to configure the native histogram parameters:

- `NativeHistogramBucketFactor(float64)`
- `NativeHistogramMaxBucketNumber(uint32)`
- `NativeHistogramMinResetDuration(time.Duration)`
- `DisableClassicHistograms()` for native-only output

A bucket factor of 1 or less disables native histograms, so `DisableClassicHistograms()` is documented to require native histograms to be enabled (otherwise prometheus silently falls back to its own default classic buckets).

twmb:
I think the options look good, my only thought (that I'm not super tied to) is for some naming consistency. The current classic histograms don't use "histogram" in the name and just use "Buckets" everywhere. What do you think of renaming:

```
NativeHistogramBucketFactor -> NativeBucketFactor
NativeHistogramMaxBucketNumber -> NativeMaxBuckets
NativeHistogramMinResetDuration -> NativeBucketMinReset (kgo options don't use Duration in the name for time.Duration opts)
DefNativeHistogramBucketFactor -> DefNativeBucketFactor
DefNativeHistogramMaxBucketNumber -> DefNativeMaxBuckets
DefNativeHistogramMinResetDuration -> DefNativeBucketMinReset
DisableClassicHistograms -> NativeBucketsOnly
```

The thinking on the last one is: kprom currently has on/off via declarative presence toggles (GoCollectors(), WithClientLabel()) rather than Disable.

pracucci:
@twmb Applied all the renames as suggested (`NativeBucketFactor`, `NativeMaxBuckets`, `NativeBucketMinReset`, the matching `Def*` constants, and `NativeBucketsOnly` in place of `DisableClassicHistograms`). Pushed in 98f63913.

For what it's worth, my personal preference would have been to keep "Histogram" in the names — I find it makes the options easier to understand at a glance (it's clear they relate to the histogram metrics). But these are minor wording details and you're the maintainer, so I'm happy to defer to your call on naming consistency here.

twmb:
I agree fwiw, I'm ~55:45 on dropping histogram vs keeping histogram, with me favoring consistency with historical names at the moment. I'll probably be 45:55 right after I cut the tag. Anyway, thanks! I'll release now.

PR #1339, opened 2026-06-05
kgo: do not exit manageFetchConcurrency while sources are pending in wantFetch

### LINKED ISSUE #1338: Consumer deadlock since v1.21.0: manageFetchConcurrency exits with pending wantFetch; stopSession waits forever, PollRecords never returns
### Summary

Since **v1.21.0**, a consumer can deadlock permanently when a consumer session is stopped (any metadata-driven cursor migration — broker flap, partition leadership movement, …) while fetch sources are registered with the fetch manager and no poll is in progress. `PollRecords` then never returns again, while the process stays up and looks healthy. The client is a zombie: 0 records delivered, no error, forever.

We hit this in production on two different workloads (a gRPC streaming gateway and a fleet of small consumer/transformer services): 6 instances of one service and 3 of 5 instances of the other wedged the same evening after a broker flap, with goroutines stuck from 3 hours up to **7.3 days**. v1.20.7 builds consuming the same clusters through the same flaps never wedged. Affected: v1.21.0, v1.21.1, v1.21.2 (latest); not fixed on master as of this writing.

### Goroutine signature

```
goroutine 13 [sync.Cond.Wait, 198 minutes]:
  sync.(*Cond).Wait
  kgo.(*consumer).stopSession              consumer.go:1819   ← waits for session.workers == 0
  kgo.(*consumerSessionStopper).stop
  kgo.(*topicPartition).migrateCursorTo    topics_and_partitions.go:647
  kgo.(*Client).mergeTopicPartitions       metadata.go:988
  kgo.(*Client).updateMetadata / updateMetadataLoop

goroutine 285319093 [chan send, 198 minutes]:
  kgo.(*source).loopFetch                  source.go:796      ← bare send: session.cancelFetchCh <- canFetch

goroutine 162 [select, 198 minutes]:
  kgo.(*Client).PollRecords                consumer.go:608    ← never returns again
```

### Root cause

v1.21.0 reworked fetch concurrency (`fetchManager` extraction + poll-gated fetch release via `pollActive`). With the default `MaxConcurrentFetches(0)`, `manageFetchConcurrency` now grants a fetch only when `pollAllowed && activeFetches == 0`:

```go
// v1.20.7 — 0 meant unbounded; wantFetch was always drained immediately:
if len(wantFetch) > 0 && (activeFetches < s.allowedFetches || s.allowedFetches == 0) {

// v1.21.0+ — 0 means poll-gated; wantFetch sits ungranted whenever no poll is active:
if len(wantFetch) > 0 && (activeFetches < fm.allowedFetches || pollAllowed && activeFetches == 0 || fm.allowedFetches < 0) {
```

…but the manager's exit condition was **not** updated accordingly:

```go
if wantQuit && activeFetches == 0 {
    return
}
```

Deadlock sequence (default `MaxConcurrentFetches(0)`):

1. Application is between polls (slow processing, or simply not polling at this instant) → `pollAllowed == false`.
2. Several `source.loopFetch` goroutines register on the unbuffered `desireFetchCh` → they accumulate in `wantFetch`, ungranted. Each holds a session worker (`incWorker`/`defer decWorker`).
3. A metadata update migrates a cursor → `migrateCursorTo` → `stopSession` cancels the session ctx and waits on `workersCond` until `session.workers == 0`.
4. `manageFetchConcurrency` sees ctx done → `wantQuit = true`, `activeFetches == 0` → **returns with `len(wantFetch) > 0`**. Nothing will ever drain `desireFetchCh`/`cancelFetchCh` again (the manager is started at most once per session via `fetchManagerStarted.Swap`).
5. The registered sources observe ctx done and take the cancel path: the bare `session.cancelFetchCh <- canFetch` send (source.go). `cancelFetchCh` is buffered (4): the first four cancels are absorbed, **every additional source blocks forever**, never reaching its `decWorker`.
6. `stopSession` waits forever on `workersCond` → metadata loop dead → `PollRecords` never returns → zombie.

In ≤ v1.20.7 step 2 was practically impossible: with `allowedFetches == 0` meaning *unbounded*, registrations were granted immediately, so `wantFetch` was always empty by the time the manager exited. The exit condition was safe then; v1.21.0's poll-gating made it reachable.

(The pre-existing exit also explains why >4 simultaneously-registered sources are needed to observe the hang — fewer than that and the cancel buffer absorbs them. In practice any consumer fetching from ≥5 brokers/sources while between polls qualifies.)

### Reproducer

Deterministic white-box test (models exactly the `source.loopFetch` handshake; fails in ~5s on v1.21.0–v1.21.2, passes with the fix, `-race` clean):

```go
package kgo

import (
	"context"
	"sync"
	"sync/atomic"
	"testing"
	"time"
)

func TestFetchManagerQuitDrainsPendingFetches(t *testing.T) {
	t.Parallel()

	const sources = 8 // > cancelFetchCh buffer of 4

	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()

	var pollActive atomic.Bool // false: no poll in progress
	pollWake := make(chan struct{}, 1)

	fm := newFetchManager(ctx, cancel, &pollActive, pollWake, 0) // 0 == default MaxConcurrentFetches

	var registered sync.WaitGroup
	var finished sync.WaitGroup
	registered.Add(sources)
	finished.Add(sources)

	for i := 0; i < sources; i++ {
		go func() {
			defer finished.Done()

			// Mirrors source.loopFetch: register the desire to
			// fetch, then either receive the grant or cancel.
			canFetch := make(chan chan bool, 1)
			select {
			case <-ctx.Done():
				registered.Done()
				return
			case fm.desireFetch() <- canFetch:
				registered.Done()
			}

			select {
			case <-ctx.Done():
				fm.cancelFetchCh <- canFetch // bare send, as in source.go
			case doneFetch := <-canFetch:
				select {
				case <-ctx.Done():
				default:
				}
				doneFetch <- true
			}
		}()
	}

	registered.Wait() // all sources are in wantFetch (unbuffered desireFetchCh)
	cancel()          // session stop: stopSession cancels the session ctx

	allDone := make(chan struct{})
	go func() {
		finished.Wait()
		close(allDone)
	}()

	select {
	case <-allDone:
	case <-time.After(5 * time.Second):
		t.Fatal("fetch sources still blocked 5s after session cancel: " +
			"manageFetchConcurrency exited with pending wantFetch, " +
			"orphaning cancelFetchCh; sources hold session workers forever " +
			"and stopSession can never finish")
	}
}
```

### Fix

Keep the manager alive until every registered source has been granted or has canceled:

```diff
--- a/pkg/kgo/consumer.go
+++ b/pkg/kgo/consumer.go
@@ -1679,7 +1679,14 @@ func (fm *fetchManager) manageFetchConcurrency() {
 			continue
 		}
 
-		if wantQuit && activeFetches == 0 {
+		// We cannot return while sources are still registered in
+		// wantFetch: each of them is (or is about to be) blocked
+		// sending on cancelFetchCh, which only we drain. Returning
+		// early orphans those sends once the channel's small buffer
+		// fills; the sources then hold session workers forever and
+		// stopSession never finishes (consumer deadlock). Keep
+		// looping until every registered source has canceled.
+		if wantQuit && activeFetches == 0 && len(wantFetch) == 0 {
 			return
 		}
 	}
```

This is safe on the quit path: once the session ctx is done, every registered source takes its `ctx.Done()` branch and sends on `cancelFetchCh`; the still-running manager drains each one (removing it from `wantFetch`), reaches `len(wantFetch) == 0`, and exits. New registrations cannot arrive after ctx done because the source's register select also has a `ctx.Done()` case.

### Validation

- New test: fails on stock v1.21.2 (5s timeout, blocked sources), passes ×5 with `-race` with the patch.
- `go test ./pkg/kgo` against a live single-node broker (`KGO_TEST_RF=1`): same results with and without the patch (no regressions).

Happy to open a PR with the patch + test if you'd like.

SLoeuillet:
Fixes #1338 — permanent consumer deadlock since v1.21.0 (PollRecords never returns after a session stop while no poll is active).

## What

`manageFetchConcurrency`'s quit condition predates the v1.21.0 poll-gating rework and can return while sources are still registered in `wantFetch`:

- with the default `MaxConcurrentFetches(0)`, registrations are only granted while a poll is in progress (`pollAllowed && activeFetches == 0`), so `wantFetch` lingers non-empty whenever the application is between polls;
- on session stop (`stopSession` via any metadata-driven cursor migration), the manager saw `wantQuit && activeFetches == 0` and returned with those sources still registered;
- nothing drains `cancelFetchCh` after that. The registered sources take the cancel path in `loopFetch` — a bare `session.cancelFetchCh <- canFetch` send. The buffer of 4 absorbs four; every additional source blocks forever, never reaching its deferred `decWorker`;
- `stopSession` waits on `workersCond` for eternity → metadata loop dead → `PollRecords` never returns. The client looks healthy but consumes nothing.

Before v1.21.0 this was unreachable: `allowedFetches == 0` meant *unbounded*, registrations were granted immediately, and `wantFetch` was always empty by quit time.

The fix keeps the manager looping until `wantFetch` is empty. Safe on the quit path: after ctx-done each registered source sends exactly one cancel which the still-running manager drains, and no new registrations can arrive (the register select also has a `ctx.Done` case).

## Testing

- New `TestFetchManagerQuitDrainsPendingFetches` models the exact `loopFetch` handshake (8 sources > cancel buffer of 4, no active poll, then session cancel). It deadlocks (5s timeout) on current master and v1.21.0–v1.21.2, and passes with `-race` on the fix.
- `go test ./pkg/kgo` against a live single-node broker (`KGO_TEST_RF=1`): identical results with and without the patch (no regressions).

Real-world stacks, affected-version analysis and production context are in #1338 (observed on v1.21.1 and v1.21.2 across two different workloads; instances stuck up to 7.3 days).

### DIFF
--- a/pkg/kgo/fetch_manager_quit_test.go
+++ b/pkg/kgo/fetch_manager_quit_test.go
@@ -1,0 +1,94 @@
+package kgo
+
+import (
+	"context"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+)
+
+// TestFetchManagerQuitDrainsPendingFetches models the exact dance every
+// source.loopFetch goroutine performs against the fetchManager, and asserts
+// that all sources can exit once the session context is canceled.
+//
+// Since v1.21.0, manageFetchConcurrency gates fetch grants on pollActive
+// (with the default MaxConcurrentFetches == 0). When no poll is in progress,
+// registered sources accumulate in wantFetch ungranted. If the session is
+// then stopped (metadata update -> migrateCursorTo -> stopSession cancels the
+// session ctx), manageFetchConcurrency observes wantQuit && activeFetches ==
+// 0 and returns immediately -- while wantFetch is still non-empty and the
+// registered sources are about to send on cancelFetchCh. cancelFetchCh is
+// buffered (4): the first four cancels are absorbed, every additional source
+// blocks forever on the bare `session.cancelFetchCh <- canFetch` send
+// (source.go, ctx.Done branch of the fetch handoff). Each such source holds a
+// session worker (incWorker/decWorker), so consumer.stopSession waits on
+// workersCond forever and PollRecords never returns: the client is a zombie.
+//
+// This is the runtime counterpart of the shutdown-path hang: it needs no
+// Close() call, only a session stop while >4 sources are registered and no
+// poll is active.
+func TestFetchManagerQuitDrainsPendingFetches(t *testing.T) {
+	t.Parallel()
+
+	const sources = 8 // > cancelFetchCh buffer of 4
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	var pollActive atomic.Bool // false: no poll in progress
+	pollWake := make(chan struct{}, 1)
+
+	fm := newFetchManager(ctx, cancel, &pollActive, pollWake, 0) // 0 == default MaxConcurrentFetches
+
+	var registered sync.WaitGroup
+	var finished sync.WaitGroup
+	registered.Add(sources)
+	finished.Add(sources)
+
+	for i := 0; i < sources; i++ {
+		go func() {
+			defer finished.Done()
+
+			// Mirrors source.loopFetch: register the desire to
+			// fetch, then either receive the grant or cancel.
+			canFetch := make(chan chan bool, 1)
+			select {
+			case <-ctx.Done():
+				registered.Done()
+				return
+			case fm.desireFetch() <- canFetch:
+				registered.Done()
+			}
+
+			select {
+			case <-ctx.Done():
+				fm.cancelFetchCh <- canFetch // bare send, as in source.go
+			case doneFetch := <-canFetch:
+				select {
+				case <-ctx.Done():
+				default:
+				}
+				doneFetch <- true
+			}
+		}()
+	}
+
+	registered.Wait() // all sources are in wantFetch (unbuffered desireFetchCh)
+	cancel()          // session stop: stopSession cancels the session ctx
+
+	allDone := make(chan struct{})
+	go func() {
+		finished.Wait()
+		close(allDone)
+	}()
+
+	select {
+	case <-allDone:
+	case <-time.After(5 * time.Second):
+		t.Fatal("fetch sources still blocked 5s after session cancel: " +
+			"manageFetchConcurrency exited with pending wantFetch, " +
+			"orphaning cancelFetchCh; sources hold session workers forever " +
+			"and stopSession can never finish")
+	}
+}
--- a/pkg/kgo/fetch_manager_quit_test.go
+++ b/pkg/kgo/fetch_manager_quit_test.go
@@ -13,7 +13,11 @@ import (
 // that all sources can exit once the session context is canceled.
 //
 // Since v1.21.0, manageFetchConcurrency gates fetch grants on pollActive
-// (with the default MaxConcurrentFetches == 0). When no poll is in progress,
+// when MaxConcurrentFetches == 0. Zero is not the default (that is -1,
+// unbounded; before v1.21.0, zero WAS both the default and unbounded): it is
+// poll-gated mode, entered by explicitly opting in with
+// MaxConcurrentFetches(0), or forced for share groups by
+// ShareMaxRecordsStrict. When no poll is in progress,
 // registered sources accumulate in wantFetch ungranted. If the session is
 // then stopped (metadata update -> migrateCursorTo -> stopSession cancels the
 // session ctx), manageFetchConcurrency observes wantQuit && activeFetches ==
@@ -39,7 +43,9 @@ func TestFetchManagerQuitDrainsPendingFetches(t *testing.T) {
 	var pollActive atomic.Bool // false: no poll in progress
 	pollWake := make(chan struct{}, 1)
 
-	fm := newFetchManager(ctx, cancel, &pollActive, pollWake, 0) // 0 == default MaxConcurrentFetches
+	// 0 == poll-gated mode: explicit MaxConcurrentFetches(0), or forced by
+	// ShareMaxRecordsStrict; not the default, which is -1 (unbounded).
+	fm := newFetchManager(ctx, cancel, &pollActive, pollWake, 0)
 
 	var registered sync.WaitGroup
 	var finished sync.WaitGroup
@@ -55,6 +61,13 @@ func TestFetchManagerQuitDrainsPendingFetches(t *testing.T) {
 			canFetch := make(chan chan bool, 1)
 			select {
 			case <-ctx.Done():
+				// Unreachable: cancel() fires only after
+				// registered.Wait() returns. If this ever
+				// fires, the premise below that every source
+				// is sitting in wantFetch is silently broken
+				// (we would count this source as registered
+				// without it registering), so flag it.
+				t.Error("ctx canceled before all sources registered; the all-sources-in-wantFetch premise is broken")
 				registered.Done()
 				return
 			case fm.desireFetch() <- canFetch:
@@ -65,9 +78,15 @@ func TestFetchManagerQuitDrainsPendingFetches(t *testing.T) {
 			case <-ctx.Done():
 				fm.cancelFetchCh <- canFetch // bare send, as in source.go
 			case doneFetch := <-canFetch:
-				select {
-				case <-ctx.Done():
-				default:
+				// Unreachable in this test (no poll is ever
+				// active and allowed fetches is 0, so the
+				// manager never grants); kept to document the
+				// full loopFetch shape, mirroring source.go
+				// exactly: a grant that lost the race with
+				// cancelation returns the token with false.
+				if ctx.Err() != nil {
+					doneFetch <- false
+					return
 				}
 				doneFetch <- true
 			}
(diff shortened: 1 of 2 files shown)


pkg/kgo/fetch_manager_quit_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"testing"
+	"time"
+)
+
+// TestFetchManagerQuitDrainsPendingFetches models the exact dance every
+// source.loopFetch goroutine performs against the fetchManager, and asserts
+// that all sources can exit once the session context is canceled.
+//
+// Since v1.21.0, manageFetchConcurrency gates fetch grants on pollActive
+// (with the default MaxConcurrentFetches == 0). When no poll is in progress,
```
twmb: The default is -1, unless you specify 0 (the old default was 0 which was unlimited, but that was changed in 1.21.0)

Are you manually passing MaxConcurrentFetches(0) for a specific reason? I'm wondering if you're using share groups or if you just passed this old default deliberately.

SLoeuillet: Good question — neither share groups nor a deliberate choice: it's a **carried-over pre-1.21 default**. Our internal wrapper exposes `max-concurrent-fetches` as a config option with a hardcoded default of `0`, set unconditionally on every client since the days when the docs read "A value of 0 implies the allowed concurrency is unbounded" (≤ v1.20.7). The v1.21.0 redefinition of `0` silently switched our whole consumer fleet (~26 services) into poll-gated mode — we bumped through 1.21.0→1.21.2 without noticing the semantic flip, since `0` had been both the default and the unbounded value for years.

We're fixing our side to pass `-1`/unset, which sidesteps the deadlock entirely. That said, I think the fix is still worth taking:

1. `0` is a legal, documented mode ("a single fetch is allowed ONLY when you poll"), and any session stop while >4 sources are registered and no poll is in flight permanently wedges the client in that mode;
2. anyone else who explicitly passed `0` under the old semantics inherits this silently on upgrade (it validates fine — `allowed: -1, badcmp: i64lt`);
3. share groups with `shareMaxRecordsStrict` *force* `maxConcurrentFetches = 0` (config.go:417), so strict share consumers are exposed to the same deadlock without ever passing `0` themselves.

Happy to adjust the test comment to make clear `0` is explicit-or-share-strict rather than "the default" — say the word.

twmb: The PR can move forward, I'll have claude add some comments comments to the test.

pkg/kgo/fetch_manager_quit_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	const sources = 8 // > cancelFetchCh buffer of 4
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	var pollActive atomic.Bool // false: no poll in progress
+	pollWake := make(chan struct{}, 1)
+
+	fm := newFetchManager(ctx, cancel, &pollActive, pollWake, 0) // 0 == default MaxConcurrentFetches
```
twmb: Same comment - not a default anymore

pkg/kgo/fetch_manager_quit_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+			select {
+			case <-ctx.Done():
+				fm.cancelFetchCh <- canFetch // bare send, as in source.go
+			case doneFetch := <-canFetch:
+				select {
+				case <-ctx.Done():
+				default:
+				}
+				doneFetch <- true
```
twmb: This branch is unreachable in this test: pollActive is false and allowedFetches is 0, so the manager can never grant. If you want to keep it as documentation of the loopFetch shape, mirror loopFetch exactly: on a canceled ctx it sends `doneFetch <- false` (source.go), not `true`, and the empty `select { case <-ctx.Done(): default: }` does nothing with its result.

```go
case doneFetch := <-canFetch:
    if ctx.Err() != nil {
        doneFetch <- false
        return
    }
    doneFetch <- true
```

As written, this will mislead anyone who later changes the test config.

pkg/kgo/fetch_manager_quit_test.go
```diff
@@ -0,0 +1,94 @@
...
+	for i := 0; i < sources; i++ {
+		go func() {
+			defer finished.Done()
+
+			// Mirrors source.loopFetch: register the desire to
+			// fetch, then either receive the grant or cancel.
+			canFetch := make(chan chan bool, 1)
+			select {
+			case <-ctx.Done():
+				registered.Done()
```
twmb, line 71: Unreachable in this flow (cancel only happens after registered.Wait()), and if it ever did fire, it would silently break the premise on L77 that all sources are in wantFetch - this path counts a source as registered without registering it. Either drop the case or t.Error here.

twmb (commented):
Please update PR body / commit to reference that `1.21.0 repurposed the meaning of MaxConcurrentFetches(0)`, not that this is still the default (and while at it, add `Closes #1338`)

SLoeuillet:
Thanks for the fast review and for pushing the test cleanup directly — your `8f57ffd` looks right to me and passes locally (`-race`, repeated runs), including the negative check (reverting the consumer.go change makes it fail in 5s again). Nothing further from my side.

One note for the squash: the original commit body still says "with the default MaxConcurrentFetches(0)" — same wording issue you flagged in the test; feel free to fix it up in the merge message.

twmb:
Eh it's fine, second commit addresses

PR #1343, opened 2026-06-12
feat(kotel): support record offset in consume/produce spans

vdorokhin:
Add record offset to Kafka consumer and producer spans. Useful information for debugging and investigation.

While working on this, I explored migrating kotel from semantic conventions v1.18.0 to v1.41.0. However, this migration introduces several naming and convention questions that require team alignment:

In v1.41.0 there is [no longer a publish operation](https://github.com/open-telemetry/opentelemetry-go/blob/main/semconv/v1.41.0/attribute_group.go#L11881) type (unlike in [v1.18.0](https://github.com/open-telemetry/opentelemetry-go/blob/main/semconv/v1.18.0/trace.go#L2779)).
This affects how we should name spans and attributes going forward. Example of the possible new naming style: [fork](https://github.com/vdorokhin/franz-go/pull/1/changes)

I’d like feedback on whether we should proceed with the semconv upgrade as part of (or right after) this change. Happy to include it in this MR or handle it separately

### DIFF
--- a/plugin/kotel/tracer.go
+++ b/plugin/kotel/tracer.go
@@ -136,7 +136,7 @@ func (t *Tracer) WithProcessSpan(r *kgo.Record) (context.Context, trace.Span) {
 		semconv.MessagingSourceName(r.Topic),
 		semconv.MessagingOperationProcess,
 		semconv.MessagingKafkaSourcePartition(int(r.Partition)),
-		semconv.MessagingKafkaMessageOffset(int(r.Offset)),
+		semconv.MessagingKafkaMessageOffsetKey.Int64(r.Offset),
 	}
 	attrs = t.maybeKeyAttr(attrs, r)
 	if t.clientID != "" {
@@ -213,6 +213,7 @@ func (t *Tracer) OnProduceRecordUnbuffered(r *kgo.Record, err error) {
 	defer span.End()
 	span.SetAttributes(
 		semconv.MessagingKafkaDestinationPartition(int(r.Partition)),
+		semconv.MessagingKafkaMessageOffsetKey.Int64(r.Offset),
 	)
 	if err != nil {
 		span.SetStatus(codes.Error, err.Error())
@@ -235,6 +236,7 @@ func (t *Tracer) OnFetchRecordBuffered(r *kgo.Record) {
 		semconv.MessagingSourceName(r.Topic),
 		semconv.MessagingOperationReceive,
 		semconv.MessagingKafkaSourcePartition(int(r.Partition)),
+		semconv.MessagingKafkaMessageOffsetKey.Int64(r.Offset),
 	}
 	attrs = t.maybeKeyAttr(attrs, r)
 	if t.clientID != "" {
--- a/plugin/kotel/tracer_test.go
+++ b/plugin/kotel/tracer_test.go
@@ -6,7 +6,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/propagation"
-	semconv "go.opentelemetry.io/otel/semconv/v1.12.0"
+	semconv "go.opentelemetry.io/otel/semconv/v1.18.0"
 	"go.opentelemetry.io/otel/trace"
 )
 


twmb:
This PR lgtm - I had Claude dig into your 1.41 problem and came up with this proposal, wdyt?

The problem, as you pointed out, is breaking name changes. Otel apparently has an answer for this in the spec ([messaging semconv](https://opentelemetry.io/docs/specs/semconv/messaging/)):

> - `messaging` - emit the new, stable messaging conventions, and stop emitting the old experimental messaging conventions
> - `messaging/dup` - emit both the old and the stable messaging conventions, allowing for a seamless transition
>
> The default behavior (in the absence of one of these values) is to continue emitting whatever version of the old experimental messaging conventions the instrumentation was emitting previously.

We can implement `OTEL_SEMCONV_STABILITY_OPT_IN` - [otelhttp](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/instrumentation/net/http/otelhttp/v0.58.0/instrumentation/net/http/otelhttp/internal/semconv/env.go#L132) did it.

Semantics:

- comma-separated list of category tokens; ours is `messaging`
- `messaging`: emit stable (new) only
- `messaging/dup`: emit both old + new (note: span name is single-valued, so dup keeps the old name; only stable flips it)
- unset: keep old (the safe default)
- `/dup` wins if both appear

Proposal (let me know if you'd want to handle this in this PR or in a new PR): read `OTEL_SEMCONV_STABILITY_OPT_IN` once in `NewTracer`, resolve to a mode, and branch on it in the hooks.

vdorokhin:
> This PR lgtm - I had Claude dig into your 1.41 problem and came up with this proposal, wdyt?
> 
> The problem, as you pointed out, is breaking name changes. Otel apparently has an answer for this in the spec ([messaging semconv](https://opentelemetry.io/docs/specs/semconv/messaging/)):
> 
> > * `messaging` - emit the new, stable messaging conventions, and stop emitting the old experimental messaging conventions
> > * `messaging/dup` - emit both the old and the stable messaging conventions, allowing for a seamless transition
> > 
> > The default behavior (in the absence of one of these values) is to continue emitting whatever version of the old experimental messaging conventions the instrumentation was emitting previously.
> 
> We can implement `OTEL_SEMCONV_STABILITY_OPT_IN` - [otelhttp](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/instrumentation/net/http/otelhttp/v0.58.0/instrumentation/net/http/otelhttp/internal/semconv/env.go#L132) did it.
> 
> Semantics:
> 
> * comma-separated list of category tokens; ours is `messaging`
> * `messaging`: emit stable (new) only
> * `messaging/dup`: emit both old + new (note: span name is single-valued, so dup keeps the old name; only stable flips it)
> * unset: keep old (the safe default)
> * `/dup` wins if both appear
> 
> Proposal (let me know if you'd want to handle this in this PR or in a new PR): read `OTEL_SEMCONV_STABILITY_OPT_IN` once in `NewTracer`, resolve to a mode, and branch on it in the hooks.

Sounds pretty good. I suggest merging this MR so as not to hold adding context to the spans. And I'll bring a new MR this week that will implement the migration to the current version of otel semconv with OTEL_SEMCONV_STABILITY_OPT_IN support in it

Do you think TODO comments should be left on the use of OTEL_SEMCONV_STABILITY_OPT_IN envs? I understand that they are supposed to be removed in the future along with the transition to the new major version.

vdorokhin:
I investigated what a migration to a more recent version might look like, and it does not appear to be worthwhile at this stage. The main issue is that most of the attributes introduced after 1.18.0 are not yet stable and are still marked as Development. As a result, it is difficult to justify referring to one version as “old” and the other as “stable.” At the same time, such a migration would introduce a considerable amount of additional code without delivering meaningful benefits.

At this point, the most reasonable approach seems to be to add the helpful attributes that already included in this MR, update the dependencies (optional), and remain on version 1.18.0 until the messaging convention reaches stable status. Once that happens, we can introduce the necessary opt-in mechanisms and proceed with the migration

I have [an example MR](https://github.com/vdorokhin/franz-go/pull/2/changes) where I implemented the solution described above. Take a look please. You may have some additional thoughts or suggestions on this approach.
Otherwise, I suggest we simply merge the current changes and postpone the migration idea until it becomes worthwhile

vdorokhin:
@twmb hi, what do you think?

twmb:
Was delayed by a massive review of #1348... I'm happy with what you're suggesting and waiting for the attributes to be stable.

twmb:
Test failure is a bug in share fetch sessions (yayyyyy the #1348 audit missed things), I can merge this now.

vdorokhin:
> Was delayed by a massive review of #1348... I'm happy with what you're suggesting and waiting for the attributes to be stable.

Yes, I saw that "monster". It is surprising that such a significant amount of changes only resulted in a patch version bump by 1 😅 

> Test failure is a bug in share fetch sessions (yayyyyy the #1348 audit missed things), I can merge this now.

Thank you

twmb:
Released as plugin/kotel/v1.7.0, thanks!

PR #1347, opened 2026-06-15
kgo: add rack-aware producer partitioning (KIP-1123)

### LINKED ISSUE #1341: Add support for KIP-1123
I just found about this project and let me tell you that is looks very well maintained, with clear information about the supported features. We are looking into this project to migrate away from `segmentio/kafka-go`.

Are there any plans to add support for [KIP-1123](https://cwiki.apache.org/confluence/display/KAFKA/KIP-1123%3A+Rack-aware+partitioning+for+Kafka+Producer)?

twmb:
Adds `RackAwarePartitioning()`, an opt-in producer option implementing [KIP-1123](https://cwiki.apache.org/confluence/display/KAFKA/KIP-1123%3A+Rack-aware+partitioning+for+Kafka+Producer). When `Rack` is also set, records that do not require a consistent partition are produced to partitions whose leader is in the client's rack, falling back to all writable partitions when no leader is. Keyed records are never affected.

### Changelog

* Added `RackAwarePartitioning`, an opt-in producer option implementing
  KIP-1123. When `Rack` is also set, unkeyed records are produced to
  partitions whose leader is in the client's rack, falling back to all
  partitions when no leader is. Keyed records are never affected. Unlike the
  Java client, this works with any partitioner. See the option's documentation
  for the partition skew this can cause.

### Design

- The same-rack subset is filtered lazily in `doPartition` and cached on `topicPartitions` under `partsMu`, keyed by the `topicPartitionsData` pointer. Every metadata update and every KIP-951 move stores a new pointer, so the subset is recomputed once per update and is never stale. This replaces the earlier `rackEligible` slice precomputed in `mergeTopicPartitions`, which threaded `brokerRacks` through `updateMetadata` and would now also need maintaining in the KIP-951 move path.
- With the option off, `doPartition` does one extra bool check per record. With it on: one pointer compare per unkeyed record, plus an O(partitions + brokers) filter once per metadata update per topic.
- Works with any partitioner: the filter runs before the partitioner is consulted. The Java client only makes its built-in partitioner rack aware.
- Records the partitioner requires consistency for (keyed, by default) are never filtered, so key placement is unchanged.
- `Rack` stays a `ConsumerOpt` (usable anywhere an `Opt` is); its doc now points at this option.
- Preferring same-rack leaders skews partition distribution when producers are not spread across racks in proportion to leaders. The option's doc says so; the KIP leaves this to operators.

### Tests

One test, `TestRackAwarePartitioning`, drives `Produce` through `doPartition` with a partitioner that always picks the last candidate, so the partition a record lands on shows which candidate set it was given: the same-rack subset for unkeyed records, all partitions for keyed, all writable when no leader is in the rack, a recomputed subset after a metadata update, and no filtering with an empty rack. Mutation checked: disabling the option, dropping the empty-rack guard, or never recomputing each fails it. The kgo suite passes against `kcl fake`.

Closes #1341

### DIFF
--- a/pkg/kgo/config.go
+++ b/pkg/kgo/config.go
@@ -126,6 +126,7 @@ type cfg struct {
 	maxProduceInflight                 int                // if idempotency is disabled, we allow a configurable max inflight
 	compression                        []CompressionCodec // order of preference
 
+	rackAwarePartitioning     bool
 	defaultProduceTopic       string
 	defaultProduceTopicAlways bool
 	maxRecordBatchBytes       func(string) int32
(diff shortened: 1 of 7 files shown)


pkg/kgo/config.go
```diff
@@ -1322,6 +1323,36 @@ func StopProducerOnDataLossDetected() ProducerOpt {
...
+// which only makes its built-in default partitioner rack-aware.
+//
+// Preferring same-rack leaders reduces cross-rack (often cross datacenter)
+// produce traffic and cost. However, it can SKEW partition distribution: if
+// your producers are not spread across racks in roughly the same proportion as
+// your partition leaders, the racks with more producers will receive
+// disproportionately more records, leaving some partitions comparatively
+// under-filled. Ensuring a sensible producer-to-leader rack balance is an
+// operator responsibility; monitor your partition distribution when using this.
+// This is why the behavior is opt-in.
```
twmb: Remove the final line, and add the KIP in the README section.

Issue #1194, opened 2025-11-26
Add an internal limit to the batch-finishing queue so it does not grow without bound


twmb:
It is possible for records to fail faster than they are promised, if you are (contrived) spin loop producing failing records. This should push memory issues back to your application, rather than OOMing within the client.

PR #1348, opened 2026-06-15
Audit fixes

twmb:
The most substantive runs were during the Fable week, with follow up rounds running through areas / scenarios of less importance. Anyway, time to review commit by commit.

### DIFF
--- a/pkg/kfake/issues_test.go
+++ b/pkg/kfake/issues_test.go
@@ -2,6 +2,7 @@ package kfake
 
 import (
 	"context"
+	"errors"
 	"hash/crc32"
 	"net"
 	"strconv"
@@ -3307,4 +3308,153 @@ func TestIssue1328(t *testing.T) {
 		}
 	}()
 	wg.Wait()
+}
+
+// TestProduceUnknownFailLimitRecreatedTopic verifies that a producer whose
+// topic is deleted and recreated fails its buffered records with
+// UNKNOWN_TOPIC_ID once the unknown-topic fail limit (UnknownTopicRetries)
+// is reached, rather than retrying forever.
+//
+// Produce v13+ addresses topics by ID (KIP-516). The client deliberately
+// never adopts a recreated topic's new ID, so every produce after recreation
+// is answered with UNKNOWN_TOPIC_ID. That error is retriable and the default
+// produce limits are unbounded (RecordRetries is effectively infinite and
+// RecordDeliveryTimeout is disabled), so the unknown-topic fail limit is the
+// only bound: UNKNOWN_TOPIC_ID must count toward it. It used to reset the
+// count instead, leaving records buffered and retrying silently forever.
+func TestProduceUnknownFailLimitRecreatedTopic(t *testing.T) {
+	t.Parallel()
+	const testTopic = "foo"
+
+	c, err := NewCluster(
+		NumBrokers(1),
+		SeedTopics(1, testTopic),
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer c.Close()
+
+	// MetadataMaxAge is long so that no periodic refresh can observe the
+	// sub-millisecond delete/create window below and fail the topic's
+	// load with UNKNOWN_TOPIC_OR_PARTITION; we want the produce failures
+	// alone (always UNKNOWN_TOPIC_ID) to drive the test.
+	cl, err := kgo.NewClient(
+		kgo.SeedBrokers(c.ListenAddrs()...),
+		kgo.DefaultProduceTopic(testTopic),
+		kgo.UnknownTopicRetries(2),
+		kgo.RetryBackoffFn(func(int) time.Duration { return time.Millisecond }),
+		kgo.MetadataMinAge(10*time.Millisecond),
+		kgo.MetadataMaxAge(time.Minute),
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer cl.Close()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	// Produce once successfully so the client locks in the original
+	// incarnation's topic ID on its per-partition produce buffer.
+	if err := cl.ProduceSync(ctx, kgo.StringRecord("before")).FirstErr(); err != nil {
+		t.Fatal(err)
+	}
+
+	// Recreate the topic: the new incarnation gets a new topic ID, and
+	// the client keeps producing with the old one.
+	adm := kadm.NewClient(cl)
+	if _, err := adm.DeleteTopics(ctx, testTopic); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := adm.CreateTopics(ctx, 1, 1, nil, testTopic); err != nil {
+		t.Fatal(err)
+	}
+
+	done := make(chan error, 1)
+	go func() {
+		done <- cl.ProduceSync(context.Background(), kgo.StringRecord("after")).FirstErr()
+	}()
+	select {
+	case err := <-done:
+		if !errors.Is(err, kerr.UnknownTopicID) {
+			t.Fatalf("got %v, want UnknownTopicID", err)
+		}
+	case <-time.After(10 * time.Second):
+		t.Fatal("produce to a recreated topic did not fail within 10s; UNKNOWN_TOPIC_ID is defeating the unknown-topic fail limit")
+	}
+}
+
+// TestProduceUnknownFailLimitNotResetByOtherErrors verifies that the
+// unknown-topic fail limit cannot be defeated by other errors interleaving
+// with UNKNOWN_TOPIC_OR_PARTITION. The limit used to reset on any
+// non-unknown error, so a broker alternating UNKNOWN_TOPIC_OR_PARTITION
+// with any other retriable error kept the count below the limit forever and
+// the records never failed. Now only a successful produce resets the count;
+// other errors leave it unchanged.
+func TestProduceUnknownFailLimitNotResetByOtherErrors(t *testing.T) {
+	t.Parallel()
+	const testTopic = "foo"
+
+	c, err := NewCluster(
+		NumBrokers(1),
+		SeedTopics(1, testTopic),
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer c.Close()
+
+	// Fail every produce request, alternating UNKNOWN_TOPIC_OR_PARTITION
+	// with the retriable NOT_ENOUGH_REPLICAS. Only the unknown errors may
+	// count toward the fail limit; the others must not reset the count.
+	var produceCalls atomic.Int32
+	c.ControlKey(int16(kmsg.Produce), func(kreq kmsg.Request) (kmsg.Response, error, bool) {
+		c.KeepControl()
+		req := kreq.(*kmsg.ProduceRequest)
+		errCode := kerr.UnknownTopicOrPartition.Code
+		if produceCalls.Add(1)%2 == 0 {
+			errCode = kerr.NotEnoughReplicas.Code
+		}
+		resp := req.ResponseKind().(*kmsg.ProduceResponse)
+		for _, rt := range req.Topics {
+			st := kmsg.NewProduceResponseTopic()
+			st.Topic = rt.Topic
+			st.TopicID = rt.TopicID
+			for _, rp := range rt.Partitions {
+				sp := kmsg.NewProduceResponseTopicPartition()
+				sp.Partition = rp.Partition
+				sp.ErrorCode = errCode
+				st.Partitions = append(st.Partitions, sp)
+			}
+			resp.Topics = append(resp.Topics, st)
+		}
+		return resp, nil, true
+	})
+
+	cl, err := kgo.NewClient(
+		kgo.SeedBrokers(c.ListenAddrs()...),
+		kgo.DefaultProduceTopic(testTopic),
+		kgo.UnknownTopicRetries(2),
+		kgo.RetryBackoffFn(func(int) time.Duration { return time.Millisecond }),
+		kgo.MetadataMinAge(10*time.Millisecond),
+		kgo.MetadataMaxAge(time.Minute),
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer cl.Close()
+
+	done := make(chan error, 1)
+	go func() {
+		done <- cl.ProduceSync(context.Background(), kgo.StringRecord("x")).FirstErr()
+	}()
+	select {
+	case err := <-done:
+		if !errors.Is(err, kerr.UnknownTopicOrPartition) {
+			t.Fatalf("got %v, want UnknownTopicOrPartition", err)
+		}
+	case <-time.After(10 * time.Second):
+		t.Fatal("produce did not fail within 10s; interleaved errors are resetting the unknown-topic fail limit")
+	}
 }
--- a/pkg/kfake/txn_churn_test.go
+++ b/pkg/kfake/txn_churn_test.go
@@ -1,0 +1,264 @@
+package kfake
+
+// Round 4 of the FRANZ_AUDIT.md program: transaction coordinator churn and
+// fencing during EndTransaction (txn-churn.md). Every test here asserts the
+// CORRECT behavior; tests whose fatal message starts with BUG REPRODUCED
+// fail against the pre-fix client and flip to passing with the txn-churn
+// fixes. The control tests document neighboring behavior that already holds
+// and must keep holding.
+
+import (
+	"context"
+	"sync/atomic"
+	"testing"
+
+	"github.com/twmb/franz-go/pkg/kerr"
+	"github.com/twmb/franz-go/pkg/kgo"
+	"github.com/twmb/franz-go/pkg/kmsg"
+	"github.com/twmb/franz-go/pkg/kversion"
+)
+
+// A transaction coordinator answers InitProducerID with
+// CONCURRENT_TRANSACTIONS whenever it is still completing a transaction for
+// the transactional ID: most notably, taking over a crashed incarnation's
+// ongoing transaction ALWAYS returns it at least once, because the broker
+// fences and aborts the old transaction and tells the new producer to retry
+// (TransactionCoordinator sendRetriableErrorCallback). The client must retry
+// in place; surfacing the error makes BeginTransaction report a routine
+// recovery as "producer ID has a fatal, unrecoverable error".
+func TestAuditTxnInitPidConcurrentTransactions(t *testing.T) {
+	t.Parallel()
+	c := newCluster(t, NumBrokers(1), SeedTopics(1, "audit-init-ct"))
+
+	var injected atomic.Int32
+	c.ControlKey(int16(kmsg.InitProducerID), func(req kmsg.Request) (kmsg.Response, error, bool) {
+		injected.Add(1)
+		resp := req.ResponseKind().(*kmsg.InitProducerIDResponse)
+		resp.ErrorCode = kerr.ConcurrentTransactions.Code
+		return resp, nil, true
+	})
+
+	cl := newPlainClient(t, c, kgo.TransactionalID("audit-init-ct"))
+	if err := cl.BeginTransaction(); err != nil {
+		t.Fatalf("BUG REPRODUCED: BeginTransaction failed after one transient CONCURRENT_TRANSACTIONS (injected %d): %v", injected.Load(), err)
+	}
+	if injected.Load() == 0 {
+		t.Fatal("control was not consumed; the test did not exercise the injection")
+	}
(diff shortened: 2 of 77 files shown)


pkg/kgo/config.go
```diff
@@ -1296,16 +1296,21 @@ func RecordRetries(n int) ProducerOpt {
...
 // multiple metadata queries (which are going to different brokers), then we
-// may as well stop trying and fail the records.
-//
-// If this is -1, the client never fails records with this error.
+// may as well stop trying and fail the records. The count is reset whenever
+// a produce to the partition succeeds; errors other than the two unknown
+// topic errors leave it unchanged. UNKNOWN_TOPIC_ID in particular is what
+// produce requests receive after a topic is deleted and recreated: produce
+// v13+ addresses topics by ID, and the client keeps the old ID until the
+// topic is purged and re-added (see PurgeTopicsFromClient).
```
twmb: Delete everything including and after `UNKNOWN_TOPIC_ID in particular` in this paragraph. Overly explanatory and no useful for users.

pkg/kgo/sink.go
```diff
@@ -1650,13 +1650,28 @@ func (recBuf *recBuf) bumpRepeatedLoadErr(err error) {
...
+//
+// UNKNOWN_TOPIC_ID counts the same as UNKNOWN_TOPIC_OR_PARTITION: produce
+// v13+ addresses topics by ID (KIP-516), and if our topic is deleted and
+// recreated, we keep the old ID forever (we never swap to the new ID; the
+// user must purge and re-add the topic). Every produce after recreation
+// returns UNKNOWN_TOPIC_ID, and since the error is retriable and record
+// retries / the record timeout are unbounded by default, this limit is the
+// only thing that fails the records and surfaces the recreation to the user.
+//
+// This returns whether we have exceeded the limit.
```
twmb: Keep prior comment. Unnecessary paragraph. At most, add "resetting the limit only on successful produce (no error)". No need for so much detail here.

pkg/kfake/issues_test.go
```diff
@@ -3308,3 +3309,152 @@ func TestIssue1328(t *testing.T) {
 	}()
 	wg.Wait()
 }
+
+// TestProduceUnknownFailLimitRecreatedTopic verifies that a producer whose
```
twmb, line 3313: These two tests can be merged. And have a tighter comment at the top here.

pkg/kfake/txn_churn_test.go
```diff
@@ -0,0 +1,264 @@
...
+		return nil, nil, false
+	})
+	return &n
+}
+
+// Under KIP-890 part 2 (transaction.version=2), partitions join the
+// transaction implicitly via produce requests, and the client marks a
+// partition as added only on a SUCCESSFUL produce response. If every produce
+// in the transaction failed, the client must still issue EndTxn(abort): the
+// broker registers the partition during produce verification BEFORE the
```
twmb, line 131: Is this mentioned in KIP-890 anywhere? Quote either the KIP or the Java source.

pkg/kgo/producer.go
```diff
@@ -645,6 +654,10 @@ func (cl *Client) produce(
 	p.bufferedBytes += userSize
 	p.mu.Unlock()
 
+	if cl.cfg.txnID != nil && !p.producedInTxn.Load() {
+		p.producedInTxn.Store(true)
```
twmb, line 713: This is set early - before we even issue a produce request. What happens if we actually now issue an EndTxn when one ISNT needed (no produce ever issued)?
Should we tighten this to be set to true only when a produce is issued? That _should_ be safe if a person actually flushes before ending.

Actually though - I see the comment on EndTxn that aborting is always save in 890v2 even if no produce happened; if we mirror that portion of the comment here, I'm fine to keep this code here (it's probably easiest here).

pkg/kgo/txn.go
```diff
@@ -28,9 +28,11 @@ const (
...
 // (EOS).
 //
-// If you are running Kafka 2.5+, it is strongly recommended that you also use
-// RequireStableFetchOffsets. See that config option's documentation for more
-// details.
+// The client always requires stable fetch offsets (KIP-447): an OffsetFetch
+// for partitions with pending transactional offsets returns
+// UNSTABLE_OFFSET_COMMIT and is retried, so a rebalanced-to consumer cannot
+// fetch offsets that a still-open transaction could yet abort. The deprecated
+// RequireStableFetchOffsets option is a no-op.
```
twmb: We can just delete this paragraph in the comment now.

pkg/kgo/consumer_share.go
```diff
@@ -43,6 +43,16 @@ type (
...
 		unresolvedAssigns map[topicID][]int32
 
+		// Whether the last assignPartitions pass skipped activating an
+		// assigned partition because our metadata does not know it yet
+		// (the broker assigned newly added partitions before our
+		// metadata refreshed). We ack the member epoch regardless, so
+		// the broker never re-sends the assignment; while this is true,
+		// handleHeartbeatResp re-returns the current assignment so we
+		// retry activation once metadata catches up. Owned by the
+		// manage goroutine.
```
twmb: Simplify this comment, and remove "activating" - it's using, not activating. The client is now using the partition.

pkg/kgo/consumer_group_848.go
```diff
@@ -91,19 +91,28 @@ outer:
...
-						// and while heartbeating. We have not yet entered heartbeating,
-						// and if the user is concurrently leaving, the lack of a memberID
-						// means both 848 and old group mgmt leaves return early.
+						// and while heartbeating. We have not yet entered heartbeating.
 						g.mu.Lock()
 						g.is848 = false
 						g.mu.Unlock()
+						// We pre-stored a self-generated member id (v1 semantics)
+						// that the server never admitted - the join just failed.
+						// Clear it so a concurrent leave returns early (there is
```
twmb, line 131: Is there a window of concurrent leave AFTER we generate, BEFORE we clear here? What happens then?

pkg/kgo/broker.go
```diff
@@ -411,27 +411,57 @@ start:
...
+					retriedOnNewConnection = true
+					goto start
+				}
+				pr.promise(nil, err)
+				return
 			}
-			pr.promise(nil, err)
-			return
+		} else {
+			cxn.cl.cfg.logger.Log(LogLevelDebug, "sasl expiry limit reached but responses are in flight, postponing reauthentication", "broker", logID(cxn.b.meta.NodeID))
```
twmb: We postpone authentication, but high volume clients can continually flood out the auth, preventing auth from ever running. Can we instead make sleep 1s and goto beginning of the reauth flow? Basically: we forever block requests on this connection until resps is empty, checking every second.

We could PROBABLY hook into cxn.resps.cond somehow to have an exact wait rather than a poll loop.

edit: nvm largely addressed by a3938acb65e93193008566ced1a33964429c7685

pkg/kgo/broker.go
```diff
@@ -777,8 +814,14 @@ type brokerCxn struct {
...
 	dead atomic.Bool
-	// closed in cloneConn; allows throttle waiting to quit
+	// closed in closeConn; allows throttle waiting to quit
 	deadCh chan struct{}
+	// hasDiscard is set during init (before the connection is shared) if
+	// this is an acks=0 produce connection running the discard goroutine.
+	// The discard goroutine owns all reads on the connection, so sasl
+	// reauthentication can never happen in place; see handleReq and
+	// loadConnection.
+	hasDiscard bool
```
twmb, line 886: We could probably hook into the discard goroutine via a SetReadDeadline(0), going back to the auth flow, and then re-entering. May not be worth the complication though.

pkg/kgo/producer.go
```diff
@@ -687,8 +698,29 @@ func (p *producer) promiseRecord(pr promisedRec, err error) {
...
 }
 
-func (p *producer) promiseRecordBeforeBuf(pr promisedRec, err error) {
-	p.promiseBatch(batchPromise{recs: []promisedRec{pr}, beforeBuf: true, err: err})
+// promiseRecordBeforeBuf finishes a record that failed before it was ever
+// buffered (and before it counted toward bufferedRecords). Only this entry
+// applies the ring's maxLen backpressure, and only for blocking Produce
+// calls: pre-buffer failures are the one promise source not bounded by the
+// max-buffered admission, so a spin loop of failing produces could otherwise
+// grow the ring without bound (#1194). TryProduce must not park, both per
```
twmb: How do we prevent the #1194 behavior if we allow TryProduce to grow the ring without bound?

edit: backed out of by ddf10e384784a04deadb27820955c8b015410075

pkg/kgo/group_balancer.go
```diff
@@ -205,6 +205,34 @@ func ParseConsumerSyncAssignment(assignment []byte) (map[string][]int32, error)
...
+	// A buggy or hostile broker can list the same member ID twice in one
+	// JoinGroup response. Balancers key plans by member ID, so a duplicate
+	// either merges (range, roundrobin) or, worse, overwrites: the sticky
+	// engine balances the duplicates as two members and then loses one
+	// side's partitions when keying its returned plan -- partitions
+	// assigned to nobody. Keep the first occurrence.
+	for i := 1; i < len(members); i++ {
+		var dup bool
+		for j := range i {
+			if members[j].MemberID == members[i].MemberID {
```
twmb: This loop should be better structured - use a map for dedupe. Range less.

pkg/kgo/source.go
```diff
@@ -1814,9 +1816,28 @@ func (o *ProcessFetchPartitionOpts) processRecordBatch(
...
+		// data batch, a control/marker batch, or a batch wholly below
+		// the requested offset. Nothing will ever call Recycle, so the
+		// recordPools put-back would never run and abort-heavy
+		// read_committed workloads would leak every pool Get. Release
+		// the pooled slices now, zeroing the written-then-discarded
+		// records like Recycle would have. Skip when a share-ack slab
+		// was created: the slab indexes rrecords by pointer, and putting
+		// the slice back would let the pool hand out memory the slab
+		// still references.
+		if slabbed {
```
twmb, line 1888: Does slabbing leak then?

pkg/kgo/sink.go
```diff
@@ -1590,6 +1590,27 @@ func (recBuf *recBuf) bufferRecord(pr promisedRec, abortOnNewBatch bool) bool {
...
+	// sweep): we either observe the cancel and fail here, or we buffer before
+	// the sweep and the sweep fails us. The unknown-topic sibling path already
+	// honors this via waitUnknownTopic's cl.ctx.Done arm; this is the missing
+	// guard on the known-topic sibling. We select on Done rather than calling
+	// Err to keep this per-record hot path free of the context's per-call mutex.
+	select {
+	case <-recBuf.cl.ctx.Done():
+		recBuf.cl.producer.promiseRecord(pr, ErrClientClosed)
+		return true
+	default:
```
twmb, line 1632: Any simpler check we can do? I don't like the perf implications here. I don't really care much about the scenario of producing to a closed client. Alternatively: quantify for me the effect / perf impact. If small, we can keep.

pkg/kgo/consumer_group.go
```diff
@@ -2032,6 +2040,15 @@ start:
 				// Instead we surface the error to the user via a
 				// fake fetch and drop the partition from this
 				// assignment; the rest of the session proceeds.
+				//
+				// Surface each partition's error exactly once across
+				// the goto-start retries: injected persists (declared
+				// before the start label), so a partition already
+				// surfaced on an earlier pass is skipped here rather
+				// than re-injected.
+				if _, ok := injected[topic][rPartition.Partition]; ok {
```
twmb, line 2103: Should we also avoid re-fetching tp's that are in injected when we retry?

pkg/kgo/consumer.go
```diff
@@ -2542,7 +2542,27 @@ func (*Client) loadEpochsForBrokerLoad(ctx context.Context, broker *broker, load
...
+				// "truncated to offset -1", surfacing a spurious ErrDataLoss
+				// and pinning the cursor at -1. Instead carry the sentinel
+				// through so the next fetch hits OFFSET_OUT_OF_RANGE and
+				// resets via the configured ConsumeResetOffset (or, under
+				// NoResetOffset, surfaces OOOR rather than a bogus data-loss
+				// error) - the same reset path the cursor already took,
+				// minus the false alarm. Matches the Java client, which
+				// resets per policy on this sentinel rather than comparing
+				// -1 against the validating position.
+				offset = rPartition.EndOffset
```
twmb, line 2569: Wasn't this commit trying to do the same thing? 7ca4b9712dacfaa0611e26dd09717f7a9a4a3033

pkg/kgo/client.go
```diff
@@ -1172,8 +1172,16 @@ func (cl *Client) updateMetadataBrokers(resp *kmsg.MetadataResponse) {
 }
 
 // updateBrokers is called with the broker portion of every metadata response.
-// All metadata responses contain all known live brokers, so we can always
-// use the response.
+// All metadata responses contain all known live brokers, so the response list
+// is authoritative and we diff-merge it: brokers absent from it are stopped.
+//
+// An EMPTY list therefore wipes all discovered brokers and falls back to
+// seeds, which is INTENTIONAL: an empty Brokers list is the KIP-1102
```
twmb: I mean I've been doing this since before 1102 but whatever

pkg/kgo/config.go
twmb: Trimmed in f447b256 - deleted everything from `UNKNOWN_TOPIC_ID in particular` on; the paragraph now ends at "...leave it unchanged."

pkg/kgo/sink.go
twmb: Trimmed in f447b256 to the terse form: one line on the reset-only-on-success rule, plus a single line naming the three errors that count (UNKNOWN_TOPIC_OR_PARTITION, UNKNOWN_TOPIC_ID, errMissingMetadataPartition). Dropped the long KIP-516 / recreation exposition.

pkg/kfake/issues_test.go
twmb: Leaving the two tests separate. They assert distinct properties - that UNKNOWN_TOPIC_ID counts toward the limit, vs. that non-unknown errors don't reset the count - and the recreated-topic test exercises the real delete/recreate path (the client keeping the old topic ID), which a pure control injection wouldn't. Merging them would couple two failure modes into one body, so I'd rather keep them as-is.

pkg/kfake/txn_churn_test.go
twmb: Verified against the broker. On the produce path verification runs before the local append (ReplicaManager.maybeSendPartitionsToTransactionCoordinator), and it builds the AddPartitionsToTxn with `verifyOnly = !supportsEpochBump`. Under TV2 (transaction.version=2) supportsEpochBump is true, so verifyOnly=false and the coordinator durably adds the partition to __transaction_state before the append - not merely verifies it. A failed append therefore leaves the partition registered server-side, exactly as the comment states. This is KIP-890's server-side-defense mechanism; the verifyOnly=false-on-TV2 path is the concrete proof. Leaving the comment as written.

pkg/kgo/producer.go
twmb: Kept here and mirrored the EndTxn rationale in a comment (f447b256). It's set at buffer time, so it can over-set: a record counted here may still be failed locally (client close, unknown-topic timeout) before any produce reaches the broker. The only consequence is an unnecessary EndTxn abort, which is always legal under 890p2 - an empty abort succeeds, bumps the epoch, and has nothing to commit anyway.

pkg/kgo/txn.go
twmb: Deleted in f447b256.

pkg/kgo/consumer_share.go
twmb: Simplified in f447b256 and dropped "activating"/"activation" - it now reads "skipped an assigned partition ... retry once metadata catches up."

pkg/kgo/consumer_group_848.go
twmb: No harmful window. The guard isn't the clear alone - it's g.managing. g.managing is set true under g.mu in the same critical section that launches manage848 (consumer_group.go:2338-2342), before manage848 stores the self-generated UUID. So any leave() that could observe the transient UUID necessarily observes managing==true, and therefore blocks on `<-g.manageDone` before it ever reads memberID. On the fallback path manage848 hands manageDone ownership to the classic manage(), and leave() has already canceled g.ctx, so the classic manage can't join and exits with memberID still cleared; leave() then reads "" and returns early. The dangerous combo (a captured is848==true reaching leave848 with a real member) is unreachable: capturing is848==true means leave locked before the fallback flipped it to false, i.e. it canceled the ctx before the classic manage was even launched. So the store("", -1) at 136 is defensive plus the clean-classic-join optimization, not the sole guard. Left the code as-is.

pkg/kgo/broker.go
twmb: Confirmed - a3938acb is in the branch. Requests now park per-connection and handleResps pushes a reauthDrainReq sentinel through the request ring when the pipeline drains, so a flood can't starve auth: nothing new is written to the connection while requests are parked, and the worker keeps serving the broker's other connections in the meantime.

pkg/kgo/broker.go
twmb: Agreed, not worth it. The discard goroutine owns all reads precisely because acks=0 has no responses to correlate; in-place reauth would need a three-way ownership handoff (force the blocking read to return, do the handshake reads on the worker, re-enter the loop) just to save a reconnect. loadConnection already recreates the expired discard connection, which authenticates in init. Leaving it documented via hasDiscard.

pkg/kgo/producer.go
twmb: Confirmed - ddf10e38 is in the branch. The exemption is backed out; promiseRecordBeforeBuf parks at maxLen again for blocking Produce and TryProduce alike. The only non-parking caller is the promise worker itself, for which produce-from-promise is documented to spawn a goroutine (whose park is safe).

pkg/kgo/group_balancer.go
twmb: Restructured in f447b256: a single map pass detects duplicates via `len(seen) != len(members)` and rebuilds only when needed. Dropped the O(n^2) scan.

pkg/kgo/source.go
twmb: No leak. release() is a put-back into the user's Pool (PutRecords/PutDecompressBytes), so skipping it just forgoes one pool return - the slices are GC'd normally, not stranded. And in this branch nkept==0, so no kept record carries the slab; it never escapes the function and is GC'd too. The skip is the conservative choice (a slab that did escape would alias pool memory if we returned the slice), and the cost is one un-pooled slice on a batch that kept zero records. Left as-is.

pkg/kgo/sink.go
twmb: Quantified: it's a non-blocking `select { case <-ctx.Done(): default: }` - a selectnbrecv with no lock and no alloc, and ctx.Done() on a cancelCtx is an atomic load on modern Go. Single-digit nanoseconds, inside the recBuf.mu we already hold, dwarfed ~20-50x by the surrounding mutex + timestamp truncation + batch append. Well under 1% of the per-record path. There's no existing closed atomic to read instead, and adding one wouldn't be meaningfully cheaper while needing the same close-ordering argument. Keeping it.

pkg/kgo/consumer_group.go
twmb: Done in f447b256. fetchOffsets now filters injected partitions out of the request on goto-start retries, so we stop re-fetching partitions we've already surfaced a non-retryable error for and dropped. The first pass is unaffected (injected is nil), and the final assignment already excluded them, so there's no behavior change beyond skipping the wasted re-fetch.

pkg/kgo/consumer.go
twmb: Related but complementary, not a duplicate. 7ca4b971 handles leaderEpoch==-1 in the metadata merge (mergeTopicPartitions) - it stops -1 from ever becoming the cursor's epoch. This change handles endOffset==-1 (UNDEFINED_EPOCH_OFFSET) in the OffsetForLeaderEpoch response, which the broker returns even for a valid >=0 requested epoch when its leader-epoch cache is empty/truncated or the epoch is newer than its log. Neither subsumes the other; both match the Java client. Keeping both.

pkg/kgo/client.go
twmb: Reframed in f447b256. The comment now describes the empty-list seed fallback as the long-standing intentional guard it is - fall back to seeds rather than be left with none, and re-cover every seed if the cluster has somehow split - with KIP-1102's rebootstrap just riding the same path rather than being the reason for it.

PR #1350, opened 2026-06-19
kgo: do not requeue offset loads after the session is canceled

### LINKED ISSUE #1349: kgo: Close hangs on a group consumer when an offset load keeps failing during group leave
Since v1.21.0, `Client.Close()` can hang for a consumer group member when a partition offset load is still failing at the time the group is left.

`(*consumerSession).listOrEpoch` reschedules failed loads in a deferred goroutine:

```go
defer s.decWorker()
defer reloads.loadWithSession(s, "reload offsets from load failure")
after := time.NewTimer(time.Second)
defer after.Stop()
select {
case <-after.C:
case <-s.ctx.Done():
    return
}
```

When the session context is canceled (group leave during `Close`), the select returns immediately, but the deferred `reloads.loadWithSession` still runs and requeues the load. The requeued `listOrEpoch` runs with the same already canceled context, fails again, and requeues again with no backoff. This busy loops and keeps a session worker alive, so `stopSession` never observes `workers == 0` and `Close` blocks until the loop happens to break (next forced metadata refresh) or the broker goes away.

This was introduced by 6c7aab5 ("kgo: fix listOrEpoch drain and nil cursor panic in listOffsets"), which removed the session-context early return from the result drain loop. Before that, a canceled session returned before `reloads` was populated, so nothing was requeued.

Reproduction: a group consumer where `ListOffsets` keeps returning a retriable error, then `Close`. With a large `MetadataMinAge` the loop does not break and `Close` hangs. PR with a fix and a regression test to follow.

Segflow:
Fixes #1349.

`(*consumerSession).listOrEpoch` requeued a failed offset load from an unconditional `defer`:

```go
defer s.decWorker()
defer reloads.loadWithSession(s, "reload offsets from load failure")
after := time.NewTimer(time.Second)
defer after.Stop()
select {
case <-after.C:
case <-s.ctx.Done():
    return
}
```

When the session context is canceled (group leave during `Close`), the select returns at once but the deferred `loadWithSession` still requeues the load. The requeued `listOrEpoch` runs with the same canceled context, fails, and requeues again with no backoff. The busy loop keeps a session worker alive, so `stopSession` never sees `workers == 0` and `Close` hangs.

This regressed in 6c7aab5, which removed the session-context early return from the drain loop. Before that, a canceled session returned before `reloads` was populated.

The fix only requeues when the backoff timer fires and returns without requeuing once the session context is done. Added `TestIssue1349` in kfake, which hangs without the fix.

### DIFF
--- a/pkg/kfake/issues_test.go
+++ b/pkg/kfake/issues_test.go
@@ -3308,3 +3308,67 @@ func TestIssue1328(t *testing.T) {
 	}()
 	wg.Wait()
 }
+
+func TestIssue1349(t *testing.T) {
+	t.Parallel()
+	const testTopic = "foo"
+
+	c, err := NewCluster(
+		NumBrokers(1),
+		SeedTopics(1, testTopic),
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer c.Close()
+
+	// Every ListOffsets fails with a reloadable error, so the group member can
+	// never finish loading its start offset and keeps scheduling reloads.
+	c.ControlKey(int16(kmsg.ListOffsets), func(kreq kmsg.Request) (kmsg.Response, error, bool) {
+		c.KeepControl()
+		req := kreq.(*kmsg.ListOffsetsRequest)
+		resp := req.ResponseKind().(*kmsg.ListOffsetsResponse)
+		for _, reqTopic := range req.Topics {
+			rt := kmsg.NewListOffsetsResponseTopic()
+			rt.Topic = reqTopic.Topic
+			for _, reqPart := range reqTopic.Partitions {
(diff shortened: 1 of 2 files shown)


twmb:
Thanks. Coincidentally, I had Claude w/ Fable (during the few days it was available) run audits over the codebase and it caught the same thing — fix here: 34b16df36. It bails out early instead of spin-looping, but importantly it preserves the offset loads still waiting to be issued, whereas this PR drops them — which itself introduces a different bug: on a session stop that *keeps* the partition (cooperative revoke, SetOffsets on other partitions, topic purge), the dropped reload becomes a silently stuck cursor that never resumes.

I'll close this in favor of #1348, which I aim to get out soon.

PR #1355, opened 2026-06-30
kgo: add nil check when loading topics in groupConsumer

iwittkau:
This was removed here: https://github.com/twmb/franz-go/commit/62388b4b84b413e3650974e8cca4d3869167d222#diff-6431ddbe3825dc29fea7d488b4767c32c552edea39a929370d7c303a7b2705d2L1687

This is not easy to reproduce, but it can panic and it does for me:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x58 pc=0x102cbe34c]

goroutine 256 [running]:
github.com/twmb/franz-go/pkg/kgo.(*groupConsumer).fetchOffsets(0x22e96b718000, {0x1043f1430, 0x22e96b3bc3c0}, 0x22e96b57af60)
        go/pkg/mod/github.com/twmb/franz-go@v1.21.4/pkg/kgo/consumer_group.go:1930 +0x36c
github.com/twmb/franz-go/pkg/kgo.(*groupConsumer).setupAssignedAndHeartbeat.func3()
        go/pkg/mod/github.com/twmb/franz-go@v1.21.4/pkg/kgo/consumer_group.go:1082 +0x6c
created by github.com/twmb/franz-go/pkg/kgo.(*groupConsumer).setupAssignedAndHeartbeat in goroutine 40
        go/pkg/mod/github.com/twmb/franz-go@v1.21.4/pkg/kgo/consumer_group.go:1079 +0x3d4
```

Issue #1271, opened 2026-03-03
Data race found in 1.20.7


oliversun9:
```
Write at 0x00c0016e0328 by goroutine 21785:
  github.com/twmb/franz-go/pkg/kgo.(*cursor).setOffset()
      github.com/twmb/franz-go@v1.20.7/pkg/kgo/source.go:223 +0x267
  github.com/twmb/franz-go/pkg/kgo.usedOffsets.finishUsingAllWithSet.func1()
      github.com/twmb/franz-go@v1.20.7/pkg/kgo/source.go:376 +0x146
  github.com/twmb/franz-go/pkg/kgo.usedOffsets.eachOffset()
      github.com/twmb/franz-go@v1.20.7/pkg/kgo/source.go:370 +0x203
  github.com/twmb/franz-go/pkg/kgo.usedOffsets.finishUsingAllWithSet()
      github.com/twmb/franz-go@v1.20.7/pkg/kgo/source.go:376 +0x32
  github.com/twmb/franz-go/pkg/kgo.(*source).takeBufferedFn()
      github.com/twmb/franz-go@v1.20.7/pkg/kgo/source.go:645 +0xb7
  github.com/twmb/franz-go/pkg/kgo.(*source).takeBuffered()
      github.com/twmb/franz-go@v1.20.7/pkg/kgo/source.go:448 +0x127
  github.com/twmb/franz-go/pkg/kgo.(*Client).PollRecords.func1()
      github.com/twmb/franz-go@v1.20.7/pkg/kgo/consumer.go:487 +0x2b4
  github.com/twmb/franz-go/pkg/kgo.(*Client).PollRecords() 
      github.com/twmb/franz-go@v1.20.7/pkg/kgo/consumer.go:529 +0x186
  github.com/twmb/franz-go/pkg/kgo.(*Client).PollFetches()
      github.com/twmb/franz-go@v1.20.7/pkg/kgo/consumer.go:412 +0x459
Previous write at 0x00c0016e0328 by goroutine 29865:
  github.com/twmb/franz-go/pkg/kgo.(*cursor).setOffset()
      github.com/twmb/franz-go@v1.20.7/pkg/kgo/source.go:223 +0x773
  github.com/twmb/franz-go/pkg/kgo.(*consumerSession).handleListOrEpochResults.func2()
      github.com/twmb/franz-go@v1.20.7/pkg/kgo/consumer.go:1936 +0x720
[...]

PR #1357, opened 2026-07-01
kgo: fix 848 stranding a regex topic assigned before metadata catches up

twmb:
## Problem

`Test848TopicCreatedAfterJoin` intermittently timed out in CI (`got 0/10`): a KIP-848 regex consumer never received records from a topic created after it joined. Not a flake -- a real race that permanently strands the topic.

With **server-side regex**, the broker resolves the pattern itself and can assign a newly-created matching topic via heartbeat **before** the client's metadata loop has evaluated the regex and added the topic to the client-side subscription (`g.using` / `g.tps`). Two client paths then dropped the just-assigned topic, and because the server believes the client owns it (its topic id is echoed back), it never re-sends -- the partition sits without a cursor and is never consumed.

A load harness reproduced it deterministically-enough to capture debug logs and pin both drop sites:

- **`revoke(revokeThisSession)`** self-revoked `nowAssigned` topics missing from `g.using`.
- **`fetchOffsets`** dropped `OffsetFetch` response topics missing from the `g.tps` subscription snapshot.

## Fix

Trust the server's assignment for 848 in both paths (gated on `is848`):

- Skip the session-end self-revoke; let the server drive revocation via heartbeats.
- Validate the `OffsetFetch` response against `added` (what the request was built from) instead of the possibly-lagging subscription snapshot. The per-topic/per-partition checks against `added` are unchanged, so the #1271 buggy-broker protection holds.

## Why classic and share groups are unaffected

- **Classic** drives its own subscription (`g.using` feeds JoinGroup and never leads `g.tps`), so `nowAssigned` only diverges from `g.tps` when a topic is *purged* after assignment -- there the topic is genuinely gone and dropping it is correct. The gate is deliberate, not just conservative: trusting `added` for classic would wrongly keep a purged topic.
- **Share groups** send resolved `SubscribedTopicNames`, never server-side regex, so the broker never assigns a topic the client hasn't already subscribed to by name.

## Verification

- Load harness: failed at iteration 302 unfixed; **7597 iterations clean** with the fix.
- `Test848*` pass under `-race`; full kfake package passes `-race`.

Also corrects the v1.21.5 CHANGELOG note for the #1355 `OffsetFetch` nil-panic: that was not from regex consuming but from a topic purged while still assigned (`PurgeTopicsFromConsuming` overlapping `AddConsumeTopics`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,36 @@
+v1.21.6
+===
+
+One bug fix:
+
+* Fixed a KIP-848 consumer stranding a regex-matched topic that was created
+  after the member joined. With server-side regex, the broker resolves the
+  pattern itself and can assign a newly-created topic via heartbeat before the
+  client's metadata loop has evaluated the regex and added the topic to its own
+  subscription. Two client paths then dropped the just-assigned topic -- the
+  end-of-session self-revoke and the `OffsetFetch` response validation. Because
+  the broker believed the client owned the topic, it never re-sent the
+  assignment, so the partition sat without a cursor and was never consumed. The
+  client now trusts the server's assignment for 848 in both paths. Classic and
+  share groups drive their own subscription, so the broker never assigns them a
+  live topic they have not subscribed to; they are unaffected.
+
+## Relevant commits
+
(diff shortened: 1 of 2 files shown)


twmb:
Analysis from a CI failure. Looks good, merging.

PR #1358, opened 2026-07-02
sr: add support for schema guid

kaiwenleee:
Confluent 8.x introduced Schema GUID, which is another schema identifier to be used in message header instead of payload prefix. 

I'd like to add basic support for this new field in schema registry client: 

- Add GUID field in `SubjectSchema` struct
- Add new API `GET /schemas/guids/{guid}`

Reference:
https://www.confluent.io/blog/schema-id-kafka-headers-data-governance/
https://docs.confluent.io/platform/current/schema-registry/develop/api.html#get-schema-by-guid

### DIFF
--- a/pkg/sr/api.go
+++ b/pkg/sr/api.go
@@ -98,6 +98,9 @@ type (
 		// ID is the globally unique ID of the schema.
 		ID int `json:"id,omitempty"`
 
+		// GUID is the globally unique ID of the schema used in Kafka header.
+		GUID string `json:"guid,omitempty"`
+
 		Schema
 	}
 )
@@ -120,6 +123,7 @@ func CommSubjectSchemas(l, r []SubjectSchema) (luniq, runiq, common []SubjectSch
 		s  string
 		v  int
 		id int
+		g  string
 	}
 	m := make(map[svid]SubjectSchema)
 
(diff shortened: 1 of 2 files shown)


pkg/sr/api.go
```diff
@@ -98,6 +98,9 @@ type (
 		// ID is the globally unique ID of the schema.
 		ID int `json:"id,omitempty"`
 
+		// GUID is the globally unique ID of the schema used in Kafka header.
+		GUID string `json:"guid,omitempty"`
```
twmb, line 102: Both this and the ID above are doc'd as "the globally unique ID of the schema", so it's hard to tell them apart. I suggest changing this to:

```
// GUID is the globally unique string identifier of the schema,
// carried in the Kafka record header rather than the payload
// prefix. Introduced in Confluent Platform 8.x.
```

while also changing the doc on ID above to:
```
// ID is the unique numeric identifier of the schema within the registry.
```

twmb:
I've built on top of your PR in #1361, I'll go ahead with that one and close this (your commit is the first in that new branch, I just covered a few other API surfaces that were not implemented and found when auditing this).

kaiwenleee:
Thank you :D

twmb:
Sorry for my delay, releasing now.

PR #1359, opened 2026-07-02
Fable audit rounds 1-5: kgo, kadm, and protocol-definition fixes

twmb:
A second Claude Fable audit wave (the first was #1348): five rounds over kgo plus the first-ever audits of kadm and the generate/definitions DSL, with every fork finding independently re-verified by full-function reads before fixing.

Important fixes:

- **Corrupt produce request on v9+ when a batch fails during serialization**: the null-records placeholder arm skipped the per-partition tagged-field terminator, desyncing the broker's parse of the entire request -- connection killed, every healthy batch failed; silent with acks=0. Round-trip regression test fails pre-fix at exactly v9.
- **EndTransaction with an unconfirmed outcome consumed txn state without failing the producer ID**: the documented retry-with-abort became a wire no-op and the next transaction silently joined the still-ongoing broker transaction under KIP-890 part 2 (EOS duplicates). The producer ID is now failed for reload, fence-aborting anything ongoing.
- **GroupTransactSession.End hung forever (ignoring its context) if the group never joined** and the transaction committed no offsets.
- **io.ErrUnexpectedEOF classified non-retryable**: a graceful disconnect landing mid-frame tore down group sessions after zero retries and could insta-fail buffered produce records.
- **Legacy (pre-0.11) compressed message sets now honor stored inner offsets** (Java/librdkafka parity); contiguous relabeling skipped records over compaction gaps when another client resumed from a franz-go commit.
- **kadm**: broadcast list APIs no longer return silently short results with a nil error; ACL filter builders no longer send the broker-rejected UNKNOWN pattern when unset (unset now means any); ACL list setters are last-call-wins; ListOffsets special listings no longer rewrite the broker's -1 answer; plus a batch of smaller fixes.
- **User callbacks moved off internal locks** (unbuffered fetch hooks, data-loss callback, OnBrokerDisconnect) so re-entrant calls cannot deadlock the client; callbacks that intentionally run under ordering locks now document their constraints.
- **kmsg**: DescribeTopicPartitions ResponsePartitionLimit default corrected to Kafka's 2000; gofumpt dropped from the generation pipeline. (Version bumps for unreleased Kafka 4.4 APIs live on the kafka-4.4 hold branch, per the definitions-must-not-exceed-released-Kafka CI check.)
- **kfake**: validates ACL filters and ApiVersions software name/version like a real broker (both had made tests structurally blind to real-broker rejections).

The full generate/definitions DSL was diffed field-by-field against Kafka's message JSON schemas (all 95 files): zero wire drift found anywhere; the two known remaining validator deltas (TxnOffsetCommit v6, streams-adjacent) are handled on the kafka-4.4 hold branch.

CHANGELOG entry is drafted locally and will land separately.

### DIFF
--- a/.github/workflows/validate-definitions.yml
+++ b/.github/workflows/validate-definitions.yml
@@ -9,6 +9,12 @@ on:
     branches: ["*"]
     paths:
       - 'generate/**'
+  # A new Kafka release moves the latest-tag checkout, so run on a schedule
+  # too: without this, a release that adds request versions produces no
+  # signal until someone happens to touch generate/ (Kafka 4.3 went
+  # unnoticed exactly this way).
+  schedule:
+    - cron: '0 8 * * 1'
 
 concurrency:
   group: validate-definitions-${{ github.ref }}
--- a/pkg/kadm/acls.go
+++ b/pkg/kadm/acls.go
@@ -699,11 +699,15 @@ func (cl *Client) CreateACLs(ctx context.Context, b *ACLBuilder) (CreateACLsResu
 	if pattern == ACLPatternUnknown {
 		pattern = ACLPatternLiteral
 	}
-	if len(b.allow) != 0 && len(b.allowHosts) == 0 {
-		b.allowHosts = []string{"*"}
+	// Default unset hosts to the wildcard in LOCALS: writing the default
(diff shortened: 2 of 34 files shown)


twmb:
Reviewed & worked out problems via current interactive session, +1.

PR #1366, opened 2026-07-07
kgo: add StreamingCompression, compressed-size-bound batch coalescing

twmb:
Adds `kgo.StreamingCompression`, an opt-in producer option that merges a partition's backlog of unsent batches at drain time into batches bounded by their COMPRESSED size, for every codec the default compressor supports. Default off; no behavior change unless enabled.

Batches are bounded by `ProducerBatchMaxBytes` measured on uncompressed records, so with a compression ratio R, a backlog costs about R times more produce round trips than `max.message.bytes` allows. Sizing `ProducerBatchMaxBytes` against an assumed ratio instead risks MESSAGE_TOO_LARGE whenever the ratio dips.

How it works: before the drain goroutine waits on an inflight slot, it freezes each partition's unsent batches under the recBuf lock, streams their records through the codec outside the lock, and splices the merged batch back in. Membership is decided before each write: the compressed size at the last flush plus the codec's worst case for everything written since (raw size plus 1/1024 plus 64 bytes for gzip, lz4, and zstd; the format's per block bound for snappy); the codec is flushed only when that bound would overflow, and the batch is cut only if the record still does not fit. The finished blob is checked against the limit as well: if a codec ever exceeds its slack, the merge is discarded, merging is turned off for the client, and a warning asks for an issue, so the slack decides only how often we flush, never what we send. Snappy merged batches use the xerial framing the Java producer writes (one independent 32KB block at a time), which every reader accepts. A source's lock is held only while its records are copied out and is released while each 32KB chunk compresses, so a failure sweep (purge, close, abort, a batch out of retries) waits at most one chunk; the merge sees the swept source when it retakes the lock and discards itself. Merged batches keep their compressed blob, so retries do not recompress. Only never-frozen batches merge, so idempotent and transactional sequencing is unchanged. Merging starts once a sink knows its produce version, so the first request per broker is unmerged. A merged batch keeps the codec and format it was sealed with; if its partition later moves to a broker that cannot take it (below produce v3, or below v7 for zstd), createReq fails those records back to the application with an error rather than retrying a request the broker rejects, once nothing is inflight ahead of the batch. Merging itself waits for a sink's first produce response, so this only arises for a partition that moves after merging.

Thanks to @scunningham, whose #1421 split the original merge into the steal, merge, and splice steps and cleaned up the stream compressor; this revision keeps that shape.

Changes from the earlier revision of this PR: sources are frozen instead of carrying a per-batch merging flag (a per-partition flag makes createReq skip the partition), the splice finds the span by pointer instead of validating drain indexes, the codec is closed inside the merge so at most one is open per sink, the worst-case bound accounts for codec raw block overhead (the fixed 96 byte slack did not hold for gzip stored blocks past ~500KB of pending data), the blob buffer is pooled, and the merge runs before the inflight wait so compression overlaps requests in flight. kgo is about +400 lines instead of +710. Tests: direct merge tests in kgo over every codec and incompressible data, a discarded merge with and without a tail, a swept source, a sweep that must not wait past one chunk of compression, a head record too big to merge, an unsupported merged batch waiting for the inflight batch ahead of it, `renumber` against recomputation with timestamps spread over days, and v8 plus v9 encoding; in kfake, one table-driven end to end test (every codec, retry, live drain, purge, close) plus a random workload test that checks every batch written stays within the limit compressed; a partition moved between brokers mid merge (which hangs without the drain trigger the merge fires), a broker too old for the codec (no merges), a merged batch staged on a sink that has not yet seen a produce response, a transaction, and a 4KB batch limit; and `StreamingCompression` is on for the producers in TestGroupETL and TestTxnEtl, which the real broker CI jobs run.

Benchmarks, blackholed kfake, 512B ~5x compressible values, 4 partitions, medians of 7 interleaved passes:

| codec | legacy ns/rec | streaming ns/rec | rec/req legacy | rec/req streaming |
|---|---|---|---|---|
| gzip | 1417 | 1460 (+3.0%) | 7317 | 100000 (13.7x) |
| snappy | 479 | 461 | 3488 | 5660 (1.6x) |
| lz4 | 387 | 402 (+3.8%) | 4615 | 37500 (8.1x) |
| zstd | 388 | 485 (+25%) | 6000 | 50000 (8.3x) |
| none | 511 | 517 | 3896 | 4615 |

Snappy compresses too fast for a backlog to build against a blackholed broker, so its density gain there is small; the end to end test, with a real backlog, merges snappy 8x. The zstd delta is klauspost's streaming encoder against EncodeAll; a follow-up could write one zstd frame per chunk with EncodeAll (decoders accept concatenated frames) and skip the flushes entirely.

A final commit speeds up compression on both paths: gzip comes from klauspost/compress (same format, already a dependency; level for level 1.1 to 1.2x the stdlib encoder, and its default level 5 is 1.7x the stdlib default 6 with a slightly better ratio), zstd no longer writes its frame checksum (the batch CRC covers the bytes; 4% off encoding), and the merge path grows its merged record slice once to the span's size, pre-grows the blob buffer to the span's bytes, pools its serialization buffers, and adjusts stamped record lengths instead of recomputing them. Both paths run at zero allocations per record; before this commit the merge path churned about 200 bytes per record growing its record slice. Before and after, same setup, medians of 5:

| codec | legacy ns/rec | streaming ns/rec |
|---|---|---|
| gzip | 1398 to 687 | 1486 to 738 |
| zstd | 380 to 386 (noise) | 455 to 377 |
| lz4 | 370 to 374 (noise) | 404 to 373 |
| snappy | 464 to 457 (noise) | 460 to 441 |

Streaming zstd is now faster than legacy zstd. The gzip decoder moved with the import and is a wash (71us per batch either way, 2 B/op).

## Changelog

- kgo: add `StreamingCompression`, a producer option that merges a partition's backlog into batches bounded by their compressed size.
- kgo: gzip compression now uses klauspost/compress/gzip (same format): 2x faster at the default level, and `WithLevel` now selects klauspost's level of that number.
- kgo: zstd batches no longer carry a frame checksum; the record batch CRC already covers them.

### DIFF
--- a/pkg/kgo/compression.go
+++ b/pkg/kgo/compression.go
@@ -1,6 +1,7 @@
 package kgo
 
 import (
+	"bufio"
 	"bytes"
 	"compress/gzip"
 	"encoding/binary"
@@ -317,6 +318,154 @@ func (c *compressor) Compress(dst *bytes.Buffer, src []byte, flags ...CompressFl
 	}
 
 	return out, use
+}
+
+// streamableCodec reports whether a codec supports the incremental Flush
+// needed for compressed-size-bound batching. gzip, zstd, and lz4 stream;
+// snappy (one-shot block s2) does not, so its batches stay bound by
+// uncompressed size.
+func streamableCodec(codec CompressionCodecType) bool {
+	return codec == CodecGzip || codec == CodecZstd || codec == CodecLz4
+}
+
+// pickCodec mirrors Compress's codec selection for a produce version: the
+// first configured option, skipping zstd when the version is below 7
+// (KIP-110: zstd requires produce v7+). An unknown version (-1, not yet
+// negotiated) optimistically assumes a modern broker; if the negotiated
+// version turns out to be ancient, sealStream demotes the batch to the
+// legacy path, which re-picks with the real version.
+func (c *compressor) pickCodec(produceVersion int32) CompressionCodecType {
+	disableZstd := produceVersion >= 0 && produceVersion < 7
+	for _, option := range c.options {
+		if option == CodecZstd && disableZstd {
+			continue
+		}
+		return option
+	}
+	return CodecNone
+}
+
+// streamStagePool holds the buffered writers that sit in front of a
+// streamCompressor's codec. Records are written one at a time (a few hundred
+// bytes each) but the codecs have real per-Write overhead; staging feeds them
+// in large chunks, matching the efficiency of the legacy whole-batch Compress.
+var streamStagePool = sync.Pool{New: func() any { return bufio.NewWriterSize(nil, 32<<10) }}
+
+// streamCompressor incrementally compresses a batch's records into dst. It lets
+// the buffering path decide batch membership by compressed size (cut a batch
+// just before the compressed bytes would exceed the wire limit) instead of the
+// conservative uncompressed bound, packing ~ratio-x more records per request.
+type streamCompressor struct {
+	c     *compressor
+	codec CompressionCodecType
+	dst   *bytes.Buffer
+	stage *bufio.Writer
+	gz    *gzip.Writer
+	zstd  *zstdEncoder
+	lz4   *lz4.Writer
+}
+
+// stream returns a streaming compressor writing into dst for a streamable
+// codec, or nil otherwise. release must be called when finished.
+func (c *compressor) stream(codec CompressionCodecType, dst *bytes.Buffer) *streamCompressor {
+	sc := &streamCompressor{c: c, codec: codec, dst: dst}
+	switch codec {
+	case CodecGzip:
+		sc.gz = c.gzPool.Get().(*gzip.Writer)
+		sc.gz.Reset(dst)
+	case CodecZstd:
+		sc.zstd = c.zstdPool.Get().(*zstdEncoder)
+		sc.zstd.inner.Reset(dst)
+	case CodecLz4:
+		sc.lz4 = c.lz4Pool.Get().(*lz4.Writer)
+		sc.lz4.Reset(dst)
+	default:
+		return nil
+	}
+	sc.stage = streamStagePool.Get().(*bufio.Writer)
+	switch codec {
+	case CodecGzip:
+		sc.stage.Reset(sc.gz)
+	case CodecZstd:
+		sc.stage.Reset(sc.zstd.inner)
+	case CodecLz4:
+		sc.stage.Reset(sc.lz4)
+	default: // unreachable: the switch above returned nil
+	}
+	return sc
+}
+
+func (sc *streamCompressor) write(p []byte) error {
+	_, err := sc.stage.Write(p)
+	return err
+}
+
+// flush flushes buffered input to dst (retaining the compression window) so
+// that len reflects everything written so far.
+func (sc *streamCompressor) flush() error {
+	if err := sc.stage.Flush(); err != nil {
+		return err
+	}
+	switch sc.codec {
+	case CodecGzip:
+		return sc.gz.Flush()
+	case CodecZstd:
+		return sc.zstd.inner.Flush()
+	case CodecLz4:
+		return sc.lz4.Flush()
+	default: // unreachable: stream() only builds streamable codecs
+		return nil
+	}
+}
+
+// len is the compressed byte count; meaningful only right after flush/finish.
+func (sc *streamCompressor) len() int { return sc.dst.Len() }
+
+// finish finalizes the stream (emits the codec footer) and returns the full
+// compressed records blob. The compressor holds only accepted records, so this
+// is always a clean, well-formed frame.
+func (sc *streamCompressor) finish() ([]byte, error) {
+	if err := sc.stage.Flush(); err != nil {
+		return nil, err
+	}
+	switch sc.codec {
+	case CodecGzip:
+		if err := sc.gz.Close(); err != nil {
+			return nil, err
+		}
+	case CodecZstd:
+		if err := sc.zstd.inner.Close(); err != nil {
+			return nil, err
+		}
+	case CodecLz4:
+		if err := sc.lz4.Close(); err != nil {
+			return nil, err
+		}
+	default: // unreachable: stream() only builds streamable codecs
+	}
+	return sc.dst.Bytes(), nil
+}
+
+// release returns the underlying writer to its pool. The zstd encoder is Reset
+// to a nil writer first: the pool's other consumer (Compress) uses EncodeAll
+// without Resetting, exactly matching how the pool's New creates encoders
+// (zstd.NewWriter(nil, ...)), so we must hand back an encoder in that same
+// EncodeAll-usable state rather than one with a closed stream attached. gzip
+// needs no equivalent because Compress always Resets it before use.
+func (sc *streamCompressor) release() {
+	sc.stage.Reset(nil)
+	streamStagePool.Put(sc.stage)
+	sc.stage = nil
+	switch sc.codec {
+	case CodecGzip:
+		sc.c.gzPool.Put(sc.gz)
+	case CodecZstd:
+		sc.zstd.inner.Reset(nil)
+		sc.c.zstdPool.Put(sc.zstd)
+	case CodecLz4:
+		sc.c.lz4Pool.Put(sc.lz4)
+	default: // unreachable: stream() only builds streamable codecs
+	}
 }
 
 type decompressor struct {
--- a/pkg/kgo/config.go
+++ b/pkg/kgo/config.go
@@ -143,6 +143,8 @@ type cfg struct {
 	partitioner Partitioner
 	compressor  Compressor
 
+	streamCompression bool
+
 	stopOnDataLoss bool
 	onDataLoss     func(string, int32)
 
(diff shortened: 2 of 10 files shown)


twmb (commented):
-

pkg/kgo/compression.go
```diff
@@ -319,6 +312,107 @@ func (c *compressor) Compress(dst *bytes.Buffer, src []byte, flags ...CompressFl
...
+	if _, ok := sc.w.(*xerialWriter); ok {
+		return n + n/6 + 36*(n/xerialBlockSize+1) + 16
+	}
+	return n + n>>10 + 64
+}
+
+// write hands buffered records to the codec, waiting for a chunk to collect
+// unless forced.
+func (sc *streamCompressor) write(force bool) error {
+	if len(sc.buf) < 32<<10 && !force {
```
twmb, line 387: why is a chunk 32K? why not more, why not less?

pkg/kgo/sink.go
```diff
@@ -2435,13 +2498,250 @@ func (b *recBatch) tryBuffer(pr promisedRec, produceVersion, maxBatchBytes int32
...
 
+// mergeBacklogs merges every partition's backlog of unsent batches into
+// compressed-size-bound batches. Compression runs on the drain goroutine
+// holding no locks, so producing never waits on it.
+func (s *sink) mergeBacklogs() {
+	cc, codec := s.streamCodec()
+	if cc == nil {
+		return
+	}
+	for i := 0; ; i++ { // a concurrent add or remove may skip or repeat one; both are harmless
```
twmb: How is this harmless?

pkg/kgo/sink.go
```diff
@@ -2435,13 +2498,250 @@ func (b *recBatch) tryBuffer(pr promisedRec, produceVersion, maxBatchBytes int32
...
+// moves that batch's remaining records to a tail batch. Back under mu, we
+// replace the consumed batches with the merged batch and tail. If a failure
+// sweep took the batches while we were merging, it owns every promise and
+// we discard the merge.
+//
+// Only never-frozen batches merge. A batch is frozen when it is first put in
+// a request, so a frozen batch may have been sent and must be resent as is.
+func (recBuf *recBuf) mergeBacklog(cc *compressor, codec CompressionCodecType) {
+	recBuf.mu.Lock()
+	var span []*recBatch
```
twmb: This should start with a default size of like 8 or something - which should stay stack allocated 90+% of the time.

pkg/kgo/sink.go
```diff
@@ -123,7 +124,7 @@ func (s *sink) createReq(id int64, epoch int16) (*produceRequest, *kmsg.AddParti
 		recBuf.mu.Lock()
 		// A pending seq reset applies only to the first batch: stage
 		// nothing more until everything inflight finishes.
-		if recBuf.failing || len(recBuf.batches) == recBuf.batchDrainIdx || recBuf.needSeqReset && recBuf.inflight != 0 || recBuf.inflightOnSink != nil && recBuf.inflightOnSink != s || recBuf.inflight != 0 && !recBuf.okOnSink {
+		if recBuf.failing || recBuf.merging || len(recBuf.batches) == recBuf.batchDrainIdx || recBuf.needSeqReset && recBuf.inflight != 0 || recBuf.inflightOnSink != nil && recBuf.inflightOnSink != s || recBuf.inflight != 0 && !recBuf.okOnSink {
```
twmb: probably high time to put each || condition on its own line

pkg/kgo/sink.go
```diff
@@ -270,6 +271,10 @@ func (s *sink) clearBackoff() {
 func (s *sink) drain() {
 	again := true
 	for again {
+		// Merging compresses backlogs while earlier requests are still
+		// in flight, so we do it before waiting on an inflight slot.
```
twmb: Reword:
// We merge before waiting on an inflight sem spot. When we saturate the wire,
// we spend time waiting on the sem; compressing while we wait increases throughput
// rather than compressing after the sem slot.
//
// More batches can build up after our initial compression attempt (which does nothing
// if only one batch exists right now). Ideally, producing takes long enough that the
// NEXT time we loop here, we have multiple batches to compress again. In the
// steady state, one merge here amortizes to be the single one good option needed.

pkg/kgo/config.go
```diff
@@ -1447,6 +1441,32 @@ func ManualFlushing() ProducerOpt {
...
+// ProducerBatchMaxBytes. Merged batches keep their compressed bytes, so
+// retries do not recompress.
+//
+// Records are buffered exactly as before; merging only happens on the drain
+// goroutine, for partitions that have a backlog, and only for batches that
+// have never been sent, so idempotent and transactional sequencing is
+// unchanged. The compression cost moves from the connection goroutine to
+// the drain goroutine. This option requires the default compressor. Merged
+// snappy batches use the framing the Java producer writes, which every
+// consumer already reads.
```
twmb: Wayyyyy overcomplicated documentation.

Just:

StreamingCompression opts the client into compressing as many batches as possible
together when producing. By default, the client creates batches internally up to
ProducerBatchMaxBytes in size, and each of these batches goes into their own
produce request. With streaming compression, the client compresses all batches
available when a produce request is about to be cut up until the point that the next
batch would exceed ProducerBatchMaxBytes. This effectively changes the max bytes option from
bounding uncompressed bytes to bounding compressed bytes.

The two potential downsides of this option are:
* Consumers cannot bound how much memory they use as well, because a consumer
  cannot ask "I want only 1MiB per partition *uncompressed*" - consumers can only
  ask for batch sizes and do not know how large a batch will inflate to.
* The client hard codes compression overhead; if the client gets it wrong and the compressed batch exceeds
  ProducerBatchMaxBytes, the client backs out of streaming compression permanently and logs a warning for you to create an issue.

Streaming compression is always recommended.

pkg/kgo/sink.go
```diff
@@ -2435,13 +2497,250 @@ func (b *recBatch) tryBuffer(pr promisedRec, produceVersion, maxBatchBytes int32
...
+		total += len(b.records)
+	}
+	recBuf.merging = true
+	recBuf.mu.Unlock()
+
+	m, tail, consumed, ok := recBuf.mergeSpan(span, total, cc, codec)
+
+	recBuf.mu.Lock()
+	defer recBuf.mu.Unlock()
+	recBuf.merging = false
```
twmb, line 2626: I think we need to defer drain, don't we?

pkg/kgo/sink.go
```diff
@@ -2435,13 +2497,250 @@ func (b *recBatch) tryBuffer(pr promisedRec, produceVersion, maxBatchBytes int32
...
+	at := slices.Index(recBuf.batches, span[0])
+	if !ok || at < 0 {
+		for _, b := range span { // records are nil, and this a no-op, if swept
```
twmb: "if swept"?

pkg/kgo/sink.go
```diff
@@ -2563,6 +2862,13 @@ func (b seqRecBatch) appendTo(
 	transactional bool,
 	compressor Compressor,
 ) (dst []byte, m ProduceBatchMetrics) { // named return so that our defer for flexible versions can modify it
+	// A merged zstd batch can reach a broker too old for zstd if the
+	// partition moved brokers after merging. We fall back to serializing
+	// and compressing the records, which picks a codec for this version.
+	if st := b.stream; st != nil && st.codec == CodecZstd && version < 7 {
+		b.wireLength = recordBatchOverhead + int32(st.uncompressed)
+		b.releaseStream()
+	}
```
twmb: How does this work? uncompressed will surely be too large.

pkg/kgo/compression.go
twmb: Measured against 4KB, 16KB, 64KB, 128KB, and 256KB for snappy and zstd: 4KB and 16KB are slower (more codec calls), 32KB through 128KB are the same, and 256KB is slower again (the chunk no longer sits in cache next to the codec's own buffers); gzip does not care. 32KB is the smallest size on the plateau and is also the xerial block size, so for snappy one Write is one block. It is now the `streamChunk` constant with that in its comment.

pkg/kgo/sink.go
twmb: A remove swaps the last recBuf into the hole. If the hole is behind the cursor, that recBuf is skipped this round and merged next round, and skipping only delays a merge: createReq sends a head batch merged or not. If the hole is ahead, we visit it at the hole. Repeats cannot happen with swap-remove. The comment now says this, and the walk is by index to avoid an allocation per request.

pkg/kgo/sink.go
twmb: Done: `make([]*recBatch, 0, 8)`.

pkg/kgo/sink.go
twmb: Done.

pkg/kgo/sink.go
twmb: Took your wording.

pkg/kgo/config.go
twmb: Replaced with your text, plus one sentence that a custom Compressor is unaffected.

pkg/kgo/sink.go
twmb: Yes, and it was a real hang. If the partition moves to another sink mid merge, that sink drains once on arrival (addRecBuf clears failing), skips the partition because `merging` is set, and then nothing wakes it. mergeBacklog now takes the merging sink and defers maybeTriggerDrain when recBuf.sink differs at splice time (before the unlock). TestStreamingCompressionMigration in kfake moves the partition mid merge and waits on the promises without a Flush: it times out after 15s without the trigger and passes in 1s with it.

pkg/kgo/sink.go
twmb: Rewritten. A sweep nils the records and owns their promises, so restoring stamps over nil restores nothing; otherwise the sources drain as they were and need their own stamps back.

pkg/kgo/sink.go
twmb: Right: the merged batch was bounded by compressed size, so its uncompressed size can exceed the limit, and the request was sized with the compressed length too. Removed. A merged batch keeps its codec; a broker that cannot read it rejects it, as it would any batch it cannot read. streamCodec's doc says so, and TestStreamingCompressionOldBroker pins that a broker below produce v7 with zstd only gets no merged batches at all.

PR #1368, opened 2026-07-09
test(kbin): cover Reader byte-array null/empty/short-read paths

nikolauspschuetz:
## What

`pkg/kbin` had tests for the varint/uvarint/varlong helpers but none for the `Reader` byte-array decoders. This adds a focused, test-only file covering their null / empty / short-read behavior.

## Why

The most valuable case is the intentional, **non-spec** EventHubs workaround documented on `Reader.Bytes` / `Reader.CompactBytes`:

```go
// ... Microsoft EventHubs fails here. -1 means null, which should throw an
// exception. EventHubs uses -1 to mean "does not exist" ...
// Until EventHubs is fixed, we return an empty byte slice for null.
if l == -1 {
    return []byte{}
}
```

That behavior is subtle (return empty, **never nil**, instead of erroring) and currently unpinned by any test — easy to "fix" away by accident. The new tests lock it in, alongside the nullable variants and the buffer-overrun path.

## Coverage added

- `Bytes` / `CompactBytes`: `-1`/`0` length → empty, **never nil** (EventHubs workaround).
- `NullableBytes` / `CompactNullableBytes`: `-1` length → `nil`.
- `CompactBytes`: `0` length (uvarint `1`) → non-nil empty.
- `Bytes` / `NullableBytes`: a length that overruns the source marks the `Reader` bad → `Complete()` returns `ErrNotEnoughData` (no read past the buffer).

Test-only; no production changes. `go test ./pkg/kbin/`, `go vet`, and `gofmt` all clean.

### DIFF
--- a/pkg/kbin/primitives_reader_test.go
+++ b/pkg/kbin/primitives_reader_test.go
@@ -1,0 +1,109 @@
+package kbin
+
+import (
+	"bytes"
+	"testing"
+)
+
+// TestReaderBytesNullAndEmpty pins the byte-array readers' handling of the null
+// (-1) length and the empty (0) length, including the intentional, non-spec
+// EventHubs workaround where the non-nullable Bytes/CompactBytes readers return
+// an empty (never nil) slice for a -1 length instead of erroring. See the
+// comment on Reader.Bytes.
+func TestReaderBytesNullAndEmpty(t *testing.T) {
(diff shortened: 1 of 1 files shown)


twmb:
Not really critical but it doesn't hurt I suppose, thanks.

PR #1369, opened 2026-07-13
kgo: add NewRecordAttrs constructor

pracucci:
### What

Adds an exported `NewRecordAttrs` constructor to `kgo`, mirroring the existing
`CompressionType` / `TimestampType` / `IsTransactional` / `IsControl` accessors:

```go
func NewRecordAttrs(compressionType uint8, timestampType int8, isTransactional, isControl bool) RecordAttrs
```

The `attrs` bits stay unexported; the constructor round-trips through the
accessors. Includes a unit test. No existing behaviour changes.

### Why

We're building [warpstream-go](https://github.com/grafana/warpstream-go) — a
produce-only Kafka client for WarpStream's stateless-agent architecture — on top
of franz-go, and we'd like it to be a drop-in for `kgo` on the produce path. When
a produce completes we populate the returned `*kgo.Record` with the same metadata
franz-go sets in `finishPromises`, including `Record.Attrs` (the batch's
compression codec + timestamp type).

That one field can't be set from outside `kgo` today: `RecordAttrs`'s only field
is unexported and there's no constructor, so external code can *read* attrs (via
the accessors) but never *build* one. It's the last produce-result field we can't
match.

So we're gently asking for this small, additive constructor to close the gap. The
name and signature are of course open to whatever you prefer — happy to adjust.

### DIFF
--- a/pkg/kgo/record_and_fetch.go
+++ b/pkg/kgo/record_and_fetch.go
@@ -67,6 +67,33 @@ func (a RecordAttrs) IsControl() bool {
 	return a.attrs&0b0010_0000 != 0
 }
 
+// NewRecordAttrs returns a RecordAttrs built from its components. It exists so
+// code outside this package can populate Record.Attrs to match what the client
+// sets: the attrs bits are otherwise unexported and only readable via the
+// accessors below.
+//
+// The arguments mirror those accessors and round-trip through them:
+// compressionType uses the CompressionType codes (0 none, 1 gzip, 2 snappy,
+// 3 lz4, 4 zstd) and timestampType uses the TimestampType convention (0 for
+// client-side CreateTime, 1 for broker-side LogAppendTime, -1 for no timestamp
+// as on v0 message sets). Both are masked to their documented bit ranges.
+func NewRecordAttrs(compressionType uint8, timestampType int8, isTransactional, isControl bool) RecordAttrs {
+	attrs := compressionType & 0b0000_0111
+	switch {
+	case timestampType < 0:
+		attrs |= 0b1000_0000 // no timestamp type
+	case timestampType > 0:
+		attrs |= 0b0000_1000 // LogAppendTime
+	}
+	if isTransactional {
+		attrs |= 0b0001_0000
+	}
+	if isControl {
+		attrs |= 0b0010_0000
+	}
+	return RecordAttrs{attrs}
+}
+
 // Record is a record to write to Kafka.
 type Record struct {
 	// Key is an optional field that can be used for partition assignment.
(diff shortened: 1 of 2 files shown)


pracucci:
@twmb Could you consider this PR, please? Just let me know if it's feasible and I will work around it differently. Thanks!

twmb:
Yes I've been thinking about this on and off (including yesterday since I was working in the repo a bunch). I'm on board with the concept and can't think of a better API. Timeline wise, I was / am still hoping to review my claude-generated open PRs, but the streaming one and topic recreation ones will require a lot of thought so I haven't prioritized full days of reviews yet.

The one thing that worries me is that attrs is private deliberately and theoretically, Kafka could add another attr down the line that would require another different function signature (i.e. this isn't forward compatible). That said, Kafka hasn't added an attr since they introduced RecordBatch's, and there's no open KIP that mentions a format change, so I don't think that's a huge risk. The only alternative I've thought of is to add setters to an existing attrs. It'd be more awkward to use, but it's forward compatible because you could just add another method. I'll keep brainstorming but _currently_ my disposition is to merge.

twmb:
So, spent a bit more time on this. What do you think about an Opts based approach?

```go
// RecordAttrsOpts contains the fields of a [RecordAttrs]. The zero value is an
// uncompressed, non-transactional, non-control record with a client-generated
// timestamp.
type RecordAttrsOpts struct {
	// Codec is the codec the record was compressed with, overriding the
	// default of no compression. The codec is stored in three bits; a
	// codec that does not fit is ignored.
	Codec CompressionCodecType

	// TimestampType is how the record's timestamp was determined; see
	// [RecordAttrs.TimestampType] for what the values mean.
	TimestampType int8

	// Transactional sets the record as a part of a transaction.
	Transactional bool

	// Control sets the record as a control record (ABORT or COMMIT).
	Control bool
}

// NewRecordAttrs returns the attrs corresponding to opts. This exists so you
// can populate [Record.Attrs] yourself; the client sets that field for you
// when producing and consuming.
func NewRecordAttrs(opts RecordAttrsOpts) RecordAttrs
```

I was wrong in the last message where I said Kafka hasn't added anything - they did in 2021 and I missed it, some random bit I don't know about (a delete horizon bit?). The opts based approach is forward compatible for adding new fields while still preserving the ease of your API.

Wdyt? I can also do this PR myself; I'm a bit behind due to an unfun life thing.

pracucci:
> What do you think about an Opts based approach?

@twmb It works to me, thanks!

> I'm a bit behind due to an unfun life thing.

Sorry to hear that.

pracucci:
The Opts approach LGTM — cleaner and forward-compatible, thanks. Pushed it in 8b060c9: `NewRecordAttrs(RecordAttrsOpts{Codec, TimestampType, Transactional, Control})`, with the codec ignored when it does not fit in three bits. No rush on your end, and happy to tweak naming/docs if you want anything changed.

twmb (approved):

twmb:
Ideally release coming within the week, merging now in prep.

PR #1370, opened 2026-07-15
deps: replace golang.org/x/crypto/pbkdf2 with stdlib crypto/pbkdf2

macdewee:
Go 1.24 added `crypto/pbkdf2` to the standard library. This PR swaps the three call sites that used `golang.org/x/crypto/pbkdf2` over to the stdlib version, dropping `golang.org/x/crypto` from `go.mod` entirely.

The motivation is FIPS 140 compliance. Go's FIPS-validated crypto module (`GOFIPS140`) only covers stdlib primitives — anything from `golang.org/x/crypto` is outside that boundary, so the dependency had to go.

The stdlib API is slightly different: the hash constructor moves to the first argument, the password becomes a plain `string` instead of `[]byte`, and the function now returns `([]byte, error)`. The error path is FIPS-only mode (`GOFIPS140` enforced), which rejects salts shorter than 16 bytes. `kfake` was generating 10-byte salts, so its salt is bumped to 16 bytes as part of this change. With that fix all three call sites are provably error-free: salts are at least 16 bytes, key lengths are 32 or 64 bytes (well above the 14-byte minimum), and both SHA-256 and SHA-512 are approved hashes.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,6 @@ require (
 	github.com/klauspost/compress v1.18.6
 	github.com/pierrec/lz4/v4 v4.1.26
 	github.com/twmb/franz-go/pkg/kmsg v1.13.1
-	golang.org/x/crypto v0.51.0
 )
 
 retract v1.11.4 // This version is actually a breaking change and requires a major version change.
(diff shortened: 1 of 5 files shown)


twmb:
Sounds good to me. I'll merge this now, it'll piggyback on whatever the next release is (patch or minor). Might be around a month as I have a bit going on and want to deeply review whatever Claude came up with for the topic recreation handling PR

PR #1371, opened 2026-07-16
kgo: classify retired broker reads as broker dead

tomplarge:
## Problem

The broker list in a metadata response is authoritative. When a broker is removed or replaced, `updateBrokers` retires it through `stopForever`, which marks each connection dead and closes its socket.

If a request is waiting for its first response at that point, the blocked read wakes with the platform's local socket-close error. `handleResp` currently treats that as an unexpected first-read failure, emits the `is SASL missing?` warning, and returns the raw network error. Requests queued behind the active read already receive `errChosenBrokerDead`, and `updateBrokers` documents that in-flight requests should receive the same retryable result.

This is especially visible with deployments whose advertised broker endpoints change frequently, but the lifecycle mismatch is generic to kgo.

## Change

- Return `errChosenBrokerDead` for a non-context socket-read failure when another goroutine has already marked the connection dead and the client remains open.
- Perform the normalization after raw read hooks and per-request debug logging, preserving the actual I/O error for low-level observability.
- Avoid both the misleading first-read SASL warning and the cancellation debug message for an intentional retirement.
- Preserve response-header and correlation errors even if the connection is retired concurrently.
- Preserve the existing warning and `ErrFirstReadEOF` behavior for unexpected first-read disconnects.

## Validation

- The retirement regression failed before the fix with the local socket-close error and SASL warning.
- The protocol-error regression failed before the refinement because `errChosenBrokerDead` masked `errCorrelationIDMismatch`.
- The logging regression failed before the refinement because retirement emitted the cancellation debug message.
- `go test -race ./pkg/kgo -run '^(TestRetiredBrokerConnectionClassifiedAsDead|TestRetiredBrokerDoesNotMaskResponseError|TestUnexpectedFirstReadStillWarns)$' -count=100`
- `go test -race ./pkg/kgo -run '^(TestRetiredBrokerConnectionClassifiedAsDead|TestRetiredBrokerDoesNotMaskResponseError|TestUnexpectedFirstReadStillWarns|TestAuditApiVersionsResetRetryClosesConns)$' -count=1`
- `go build ./pkg/kgo`
- `go vet ./pkg/kgo`
- `gofmt` and `git diff --check`

The full integration suite was not run locally because it requires a broker; the repository's PR workflow covers Kafka, kfake, and Redpanda.

### DIFF
--- a/pkg/kgo/broker.go
+++ b/pkg/kgo/broker.go
@@ -1571,6 +1571,14 @@ func (cxn *brokerCxn) readResponse(
 	}
 
 	if readErr != nil {
+		// A metadata update can retire this broker while a response read is
+		// blocked. stopForever calls die, which marks the connection dead
+		// before closing it; the read then wakes with a platform-specific
+		// socket error. Return the same retryable error as requests queued
+		// behind this read, unless the client or request itself was canceled.
+		if cxn.dead.Load() && cxn.cl.ctx.Err() == nil && !isContextErr(readErr) {
+			return nil, errChosenBrokerDead
+		}
 		return nil, readErr
 	}
 	if len(buf) < 4 {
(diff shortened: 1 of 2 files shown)


pkg/kgo/broker.go
```diff
@@ -1850,22 +1858,24 @@ func (cxn *brokerCxn) handleResp(pr promisedResp) {
...
-			} else {
-				cxn.b.cl.cfg.logger.Log(LogLevelWarn, "read from broker errored, killing connection after 0 successful responses (is SASL missing?)", "req", kmsg.Key(pr.resp.Key()).Name(), "addr", cxn.b.addr, "broker", logID(cxn.b.meta.NodeID), "err", err)
-				if err == io.EOF || err == io.ErrUnexpectedEOF { // specifically avoid checking errors.Is to ensure this is not already wrapped
-					// ErrUnexpectedEOF gets the same first-read
-					// pessimism: now that it is retryable in
-					// isRetryableBrokerErr, a mid-frame close on
-					// the very first read (bad SASL cutting us off
-					// mid-response) must not retry forever either.
-					err = &ErrFirstReadEOF{kind: firstReadSASL, err: err, retry: cxn.b.cl.cfg.alwaysRetryEOF}
+		if !errors.Is(err, errChosenBrokerDead) {
```
twmb, line 1861: a few more levels of conditionals here and I think the codebase may finally start looking like some real solid enterprise code

twmb (approved):
looks good, this will piggyback on the next patch or minor, whatever comes first

PR #1372, opened 2026-07-21
kgo: flatten topic/partition maps in group rebalance logs

constanca-m:
On rebalance, `kgo` logs affected partitions as a raw `map[string][]int32` keyed by topic. A structured (JSON) logger turns each map key into its own field, so with high-cardinality topics (for example one topic per tenant) the field count grows without bound on the log backend.

`kgo` is inconsistent about this. It already wraps the same data in the `mtps` stringer on "new group session begun" and "synced":

- https://github.com/twmb/franz-go/blob/dd56926b16a14d119f6c4c0212051b79c9f9aed8/pkg/kgo/consumer_group.go#L1045
- https://github.com/twmb/franz-go/blob/dd56926b16a14d119f6c4c0212051b79c9f9aed8/pkg/kgo/consumer_group.go#L1774

But it passes the raw map on "calling onRevoke at the end of a session" (`lost`), both "revoking prior assigned partitions" lines (`revoking`), and the debug "entering <callback>" lines (`with`):

- https://github.com/twmb/franz-go/blob/dd56926b16a14d119f6c4c0212051b79c9f9aed8/pkg/kgo/consumer_group.go#L846
- https://github.com/twmb/franz-go/blob/dd56926b16a14d119f6c4c0212051b79c9f9aed8/pkg/kgo/consumer_group.go#L749-L753
- https://github.com/twmb/franz-go/blob/dd56926b16a14d119f6c4c0212051b79c9f9aed8/pkg/kgo/consumer_group.go#L401-L405

A `map[string][]int32` has no `String()`, so it reflects into nested keys. [`mtps`](https://github.com/twmb/franz-go/blob/dd56926b16a14d119f6c4c0212051b79c9f9aed8/pkg/kgo/topics_and_partitions.go#L97-L118) implements `fmt.Stringer`, so wrapping the value emits a single string. This change wraps those values in `mtps`; only log formatting changes, the same information is preserved.

Example now:

```json
"lost": {
  "events.tenant-a1": [0],
  "metrics.tenant-a1": [0]
}
```

Example after this change:

```json
"lost": "events.tenant-a1[0], metrics.tenant-a1[0]"
```

### DIFF
--- a/pkg/kgo/consumer_group.go
+++ b/pkg/kgo/consumer_group.go
@@ -399,9 +399,9 @@ func (c *consumer) initGroup() {
 			default:
 			}
 			if ctxExpired {
-				cl.cfg.logger.Log(LogLevelDebug, "entering "+name, "with", m, "context_expired", ctxExpired)
+				cl.cfg.logger.Log(LogLevelDebug, "entering "+name, "with", mtps(m), "context_expired", ctxExpired)
 			} else {
-				cl.cfg.logger.Log(LogLevelDebug, "entering "+name, "with", m)
+				cl.cfg.logger.Log(LogLevelDebug, "entering "+name, "with", mtps(m))
 			}
 			if user != nil {
 				dup := make(map[string][]int32)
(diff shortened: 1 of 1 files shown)


twmb:
Sure makes sense, this will piggyback on the next patch or minor.

PR #1374, opened 2026-07-27
kgo: drop the topic partition lock before buffering a record

twmb:
`doPartition` held `parts.partsMu` across both the partitioner call **and**
`bufferRecord`. There is one `partsMu` per topic, so every produce to a topic
serialized on one mutex no matter how many partitions the topic had — producing
to a single topic scaled *negatively* with cores.

A mutex profile of 8 goroutines producing keyed records to one 64-partition
topic put **97% of all mutex delay in `doPartition`** (the global `producer.mu`
was ~2%).

### Why the lock can be dropped

`partsMu` guards exactly two things: `parts.partitioner` (lazy init plus the
partitioner's own mutable state) and `parts.lb`, the `leastBackupInput` fed to
`TopicBackupPartitioner`. It guards nothing `bufferRecord` touches, and
`bufferRecord` takes the target `recBuf`'s own mutex (`sink.go`, `recBuf.mu`).
Once the partitioner has handed back a pick, the lock has no further work.

`mapping` is not protected by it either: `partsData` is an immutable snapshot
loaded *before* `partsMu` is taken, so a metadata refresh can already swap it
under the lock today. Buffering into a `recBuf` from a stale snapshot is the
existing design — `migrateProductionTo` reuses the same `recBuf` pointer in the
new `topicPartition` under `recBuf.mu`.

### Why the KIP-480 path keeps the old scope

For a partitioner implementing `TopicPartitionerOnNewBatch`, a buffer that would
start a new batch aborts, and the recovery calls `OnNewBatch` and re-picks. That
trio (pick, abort, re-pick) must be atomic against the partitioner's state:
`OnNewBatch` un-pins the sticky partition, so a concurrent produce landing in
between would either be handed the partition we are abandoning, or would re-pin
one that our `OnNewBatch` then clobbers. Those partitioners keep the lock across
the buffering.

Shipped partitioners that implement it — `StickyPartitioner`,
`StickyKeyPartitioner` (embeds `stickyTopicPartitioner`), `LeastBackupPartitioner`.
Ones that do not, and so take the narrowed scope — the **default**
`UniformBytesPartitioner`, `RoundRobinPartitioner`, `BasicConsistentPartitioner`,
`ManualPartitioner`.

### Benchmarks

One topic, 64 partitions, 128-byte values, `ManualFlushing`, `b.RunParallel`,
`benchstat` over 12 runs per point (each run in its own process — see the
benchmark comment for why).

| GOMAXPROCS | keyed, spread over partitions | keyless, default sticky partitioner | topic per goroutine (control) |
|---|---|---|---|
| 1 | 327n -> 262n | 239n -> 233n (~) | ~ |
| 2 | 369n -> 276n (-25%) | 334n -> 292n (~) | ~ |
| 4 | 561n -> 347n (-38%) | 431n -> 400n (-7%) | ~ |
| 8 | 732n -> 504n (-31%) | 569n -> 542n (-5%) | ~ |

A second full run on a busier machine reproduced -11% / -23% / -34% at 2/4/8 for
the keyed shape and -6.5% at 8 for the keyless shape. A focused 20-sample
`GOMAXPROCS=1` rerun showed **no** significant difference (p=0.165), which is the
expected result with no contention — the apparent gain in the table's first row
did not survive resampling.

The keyless case gains less because the default partitioner is sticky: it pins
one partition at a time, so the goroutines converge on one `recBuf` and the
contention only moves from `partsMu` to `recBuf.mu`. Total mutex delay for that
shape still drops ~25% (582ms -> 438ms) because the two locks now pipeline.

Aggregate mutex delay for the contended shape drops correspondingly; the control
benchmark (one topic per goroutine, no shared `partsMu`) is unchanged, which is
the point.

### Testing

`TestConcurrentProducePartitioning` produces concurrently through every shipped
partitioner with a batch cap small enough that batches roll — and so
`OnNewBatch` fires — constantly, asserting each record is promised exactly once
with an in-range partition. It needs no broker.

It is a real guard, not a formality: reverting just the KIP-480 exception (so
the lock is dropped for `OnNewBatch` partitioners too) makes it fail with 69
data races, all on `stickyTopicPartitioner.OnNewBatch`.

- `gofmt -l`, `go vet ./...`, `go build ./...`, `golangci-lint run ./...` — clean
- `go test -race ./pkg/kgo/internal/...` — pass
- `go test -race -count=4` over the new test and the partitioner tests — pass
- The broker-backed `pkg/kgo` suite was **not** run (no broker on :9092 available here)

### Checked against #1347

#1347 (KIP-1123 rack-aware partitioning) also edits `doPartition`, but only the
`mapping` selection *before* the pick — it does not touch the lock scope,
`bufferRecord`, or the `OnNewBatch` interaction, and nothing in its diff or its
review discussion states an invariant requiring `partsMu` to span the buffering.
The two merge cleanly (verified with a trial merge: no conflicts, builds, both
test sets pass), and the composition is correct — the rack filter narrows
`mapping` while still under `partsMu`, well before any unlock point.

### DIFF
--- a/pkg/kgo/producer.go
+++ b/pkg/kgo/producer.go
@@ -887,8 +887,13 @@ func (cl *Client) doPartition(parts *topicPartitions, partsData *topicPartitions
 		return
 	}
 
+	// partsMu exists only to serialize access to the topic's partitioner
+	// (and the least-backup input we feed it): it guards nothing that
+	// bufferRecord touches, and bufferRecord takes the target recBuf's own
+	// mutex. Every exit below therefore unlocks as soon as the partitioner
+	// is done being used -- see the comment above the buffering below for
+	// why the KIP-480 path is the one exception.
 	parts.partsMu.Lock()
-	defer parts.partsMu.Unlock()
 	if parts.partitioner == nil {
 		parts.partitioner = cl.cfg.partitioner.ForTopic(pr.Topic)
 	}
@@ -910,6 +915,7 @@ func (cl *Client) doPartition(parts *topicPartitions, partsData *topicPartitions
 		mapping = partsData.partitions
 	}
 	if len(mapping) == 0 {
+		parts.partsMu.Unlock()
 		cl.producer.promiseRecord(pr, errors.New("unable to partition record due to no usable partitions"))
 		return
 	}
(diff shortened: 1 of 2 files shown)


twmb:
Not worth it: contention on the outer mutex (partsMu) drops, but then we just replace that by contention on the inner mutex (the one on recBuf). The results aren't impressive. Apparently this helps more for keyed records, but by like, 300ns per record and only when producing very concurrently to different keys.

PR #1375, opened 2026-07-27
kgo: skip metadata refresh work that nothing changed

twmb:
Three independent optimizations to the periodic metadata refresh. Each removes
work a refresh did unconditionally but that only matters when the cluster
actually changed. Behavior is unchanged in every case; the slow paths still run
exactly as before.

Based on `master`. This does not touch the per-partition debug logs in
`mergeTopicPartitions` that #1373 guards, so the two should not conflict.

### 1. Rebuild the topic ID map only when a topic ID changed

`updateMetadata` cloned `id2t` and built a throwaway reverse name set on every
refresh. Topic IDs only move when a topic is created, deleted, or recreated, so
the extracted `mergeTopicIDs` first walks the response: if every non-zero ID
already maps to that same name, the merge below provably cannot change anything
(the name is necessarily in `knownNames`, since it is a value in the map we
would clone, and the map already holds the exact entry the merge would write),
so we skip the rebuild and leave the stored map in place. The walk short
circuits on the first difference, so a response that does change something pays
at most one extra map lookup per topic before it finds one.

### 2. Sort partitions only when the response is out of order

`fetchTopicMetadata` sorted every topic's partitions and then verified they are
exactly `0..n-1`. That verification proves sortedness when it passes -- partition
`i` sitting at index `i` is sorted by definition -- so `ensurePartitionsOrdered`
runs the check first and only sorts when it fails, then re-checks. An
out-of-order or incomplete response takes the identical old path (sort, verify,
same error text). A passing check also rules out duplicate partitions, so there
is no input where the old unstable sort and the new skip could produce different
orders.

### 3. Set the recBuf linger callback on first linger

`newPartition` stored `r.lingerFn = r.unlingerAndManuallyDrain` on every recBuf
it built. A method value is an allocation, and a refresh builds a recBuf for
every partition it sees only to discard the ones that already existed, while few
partitions ever linger. It is now set in `lockedMaybeLinger`, next to the
`time.AfterFunc` that consumes it.

`lingerFn` has exactly one reader (that `AfterFunc` call), which is where the
lazy set lives, and both call sites that reach it hold `recBuf.mu`, so the write
is under the same mutex as the read. The only other `&recBuf{}` in the package
(`topics_and_partitions.go`, `modifyP`) is a field-carrying template whose
`sink` is copied into the live recBuf before it is discarded -- it never
lingers, and it did not set `lingerFn` before this change either.

## Benchmarks

New in `pkg/kgo/metadata_test.go`; they need no broker. The baseline is `master`
with the two extracted functions lifted verbatim out of `updateMetadata` /
`fetchTopicMetadata` into a (throwaway, uncommitted) test file, so the same
benchmarks run against the old algorithms. Apple M1, `-count 5` for
allocations, interleaved base/new rounds for time (`-count 6`, `-count 10` for
`NewPartitions`).

Allocation counts are deterministic and are the primary result:

| benchmark | before | after |
| --- | --- | --- |
| `MergeTopicIDs/unchanged/100` | 12.75 KiB, 7 allocs | 0 B, 0 allocs |
| `MergeTopicIDs/unchanged/1000` | 213.5 KiB, 15 allocs | 0 B, 0 allocs |
| `MergeTopicIDs/unchanged/10000` | 1.668 MiB, 99 allocs | 0 B, 0 allocs |
| `MergeTopicIDs/changed/10000` | 1.668 MiB, 99 allocs | 1.668 MiB, 99 allocs |
| `EnsurePartitionsOrdered/ordered/100` | 168 B, 3 allocs | 0 B, 0 allocs |
| `EnsurePartitionsOrdered/ordered/1000` | 168 B, 3 allocs | 0 B, 0 allocs |
| `EnsurePartitionsOrdered/ordered/10000` | 168 B, 3 allocs | 0 B, 0 allocs |
| `EnsurePartitionsOrdered/shuffled/10000` | 168 B, 3 allocs | 168 B, 3 allocs |
| `NewPartitions/produce/100` | 29.98 KiB, 303 allocs | 28.42 KiB, 203 allocs |
| `NewPartitions/produce/1000` | 297.4 KiB, 3003 allocs | 281.7 KiB, 2003 allocs |
| `NewPartitions/produce/10000` | 2.903 MiB, 30003 allocs | 2.750 MiB, 20003 allocs |
| `NewPartitions/consume/10000` (control) | 2.140 MiB, 20003 allocs | 2.140 MiB, 20003 allocs |

`NewPartitions/produce` drops exactly one allocation per partition; the consume
path, which this PR does not touch, is identical on both sides.

Time, medians:

| benchmark | before | after | delta |
| --- | --- | --- | --- |
| `MergeTopicIDs/unchanged/100` | 16.02 us | 3.96 us | -75% (p=0.002) |
| `MergeTopicIDs/unchanged/1000` | 336.4 us | 40.0 us | -88% (p=0.002) |
| `MergeTopicIDs/unchanged/10000` | 3.800 ms | 466.6 us | -88% (p=0.002) |
| `MergeTopicIDs/changed/10000` | 3.750 ms | 3.574 ms | ~ (p=0.937) |
| `EnsurePartitionsOrdered/ordered/100` | 868 ns | 157 ns | -82% (p=0.002) |
| `EnsurePartitionsOrdered/ordered/1000` | 6.25 us | 1.51 us | -76% (p=0.002) |
| `EnsurePartitionsOrdered/ordered/10000` | 67.9 us | 17.9 us | -74% (p=0.002) |
| `EnsurePartitionsOrdered/shuffled/10000` | 2.540 ms | 2.871 ms | ~ (p=0.394) |
| `NewPartitions/produce/100` | 14.50 us | 12.45 us | -14% (p=0.001) |
| `NewPartitions/produce/1000` | 171.5 us | 148.6 us | ~ (p=0.143) |
| `NewPartitions/produce/10000` | 2.233 ms | 1.740 ms | ~ (p=0.218) |
| `NewPartitions/consume/100` (control) | 11.01 us | 11.18 us | ~ (p=0.912) |

Caveat on the timing column: another process was running benchmarks on the same
machine throughout, so the confidence intervals are wide (frequently +/-50%) and
benchstat calls several plausible improvements insignificant. The untouched
`NewPartitions/consume` control drifts by up to ~15% between runs, which is the
noise floor here; read anything smaller as noise, and treat the allocation table
above as the reliable measurement.

## Testing

Ran: `gofmt -l` (clean over `pkg/`; `examples/bench/main.go` is already
unformatted on `master` and is untouched here), `go vet ./...`,
`go build ./...`, `go test -race ./pkg/kgo/internal/...`, and `go test -race` on
the new `TestMergeTopicIDs` / `TestEnsurePartitionsOrdered` plus
`TestAuditTxnRecheckRewindUnmarksAddedToTxn` -- the one existing test in a file
this PR touches, since it mirrored the recBuf construction that change 3
removes.

**Not run: `go test ./pkg/kgo/`.** That suite needs a Kafka broker on :9092,
which was not available in this environment. The full refresh path
(`updateMetadata` / `fetchTopicMetadata` end to end against a live broker) is
therefore unverified here; the new unit tests cover the two extracted functions
directly, including the recreated-topic case and the missing / duplicate /
out-of-order / negative partition cases.

### DIFF
--- a/pkg/kgo/metadata.go
+++ b/pkg/kgo/metadata.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/twmb/franz-go/pkg/kerr"
 	"github.com/twmb/franz-go/pkg/kgo/internal/xsync"
+	"github.com/twmb/franz-go/pkg/kmsg"
 )
 
 type metawait struct {
@@ -90,6 +91,69 @@ func (cl *Client) id2tMap() map[[16]byte]string {
 		return noid2t
 	}
 	return m
+}
+
+// mergeTopicIDs merges the topic ID mappings in a metadata response into the
+// existing id2t map. We clone rather than rebuild so that topics missing from
+// this particular metadata response (transient broker omission, non-all
+// request) retain their ID mapping from prior responses.
+//
+// If a topic was deleted and recreated, the broker returns a new ID for the
+// same name. We do NOT add the new ID if the old ID is still present - the old
+// mapping is preserved until the user explicitly purges via
+// PurgeTopicsFromClient. This avoids having two IDs for the same topic name.
+func (cl *Client) mergeTopicIDs(latest map[string]*metadataTopic) {
+	old := cl.id2tMap()
+
+	// Steady state is that every ID in the response is already mapped to
+	// the same name: IDs only change when a topic is created, deleted, or
+	// recreated. Cloning the map and building the reverse name set costs
+	// two allocations sized by the number of topics, on every refresh, so
+	// we first walk the response to see if there is anything to do. If
+	// every ID maps to the name we already have, the merge below cannot
+	// change anything: the name is necessarily in knownNames (it is a
+	// value in the map we would clone) and the map already holds the
+	// entry the merge would write.
+	var changed bool
+	for _, mt := range latest {
+		if mt.id == ([16]byte{}) {
+			continue
+		}
+		if name, ok := old[mt.id]; !ok || name != mt.topic {
+			changed = true
+			break
+		}
+	}
+	if !changed {
+		return
+	}
+
+	merged := make(map[[16]byte]string, len(old)+len(latest))
+	maps.Copy(merged, old)
+
+	// Build the set of topic names that already have an ID.
+	knownNames := make(map[string]struct{}, len(merged))
+	for _, name := range merged {
+		knownNames[name] = struct{}{}
+	}
+
+	for _, mt := range latest {
+		if mt.id == ([16]byte{}) {
+			continue
+		}
+		if _, exists := knownNames[mt.topic]; exists {
+			// This name already has an ID in the map. Only update
+			// if it's the same ID (normal case), skip if it's a
+			// different ID (recreated topic).
+			if _, sameID := merged[mt.id]; sameID {
+				merged[mt.id] = mt.topic
+			}
+			continue
+		}
+		merged[mt.id] = mt.topic
+		knownNames[mt.topic] = struct{}{}
+	}
+	cl.id2t.Store(merged)
 }
 
 // waitmeta returns immediately if metadata was updated within the last second,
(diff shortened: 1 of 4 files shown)


twmb:
Closing as overcomplicated: it's more difficult code, and while current code has more allocations, I assume that's fine because we're talking one independent isolated goroutine every 5min.

PR #1379, opened 2026-07-28
kgo: exact sticky balancing on balance, rack, and stickiness; BalanceRacks opt in

twmb:
Makes the sticky balancers exactly optimal on balance, then rack placement, then stickiness, makes rack aware balancing opt in, and speeds sticky balancing up. Rebuilt from the earlier version of this PR as six commits that each build and pass; every claim below is verified in the tree.

**Rack aware balancing is opt in.** Since v1.21.0, setting `Rack` also made the range and sticky balancers prefer partitions whose leader is in the member's rack. That only saves anything when fetches go to the leader, which is the broker default; with a rack aware replica selector and a replica in every rack, assigning by leader rack gains nothing. Kafka's client bundles the two and assumes the selector; this client asks, via the new `BalanceRacks` option, and warns if it sees preferred read replicas while `BalanceRacks` is on. Partition racks are still built and handed to every `ConsumerBalancer` whenever a member reports a rack, so custom balancers keep `PartitionRacks()`; `BalanceRacks` only decides whether the range and sticky balancers use them. **Rack-only configs lose rack aware placement on upgrade unless they add `BalanceRacks()`.**

**A repair pass after balancing settles balance, rack, and stickiness exactly.** `balance()` was already load optimal but moved whichever partition its search reached first and never revisited a carried-over partition, so stickiness fell short of what the same counts allowed, and rack locality decayed to random as leaders moved (96% to 33% over 25 rebalances at 10% leader movement). The repair collapses the plan onto a table of interchangeable partitions by members and cancels improving rotations until none remain: a min-cost flow on the plan itself, with load shifts allowed only between members one level apart through a node per load level, so nothing is built and no large weights are needed. Rack outranks stickiness on purpose: a misplaced partition costs cross zone traffic for as long as it stays misplaced, moving it is paid once.

**Verified against the true optimum.** Brute force over 89,898 random small instances at 100,000 seeds, a quarter of them uniform, a quarter without racks, and some with over 64 members so the subscriber bitsets span two words: load vector optimal, then the fewest partitions off rack, then stickiness at the maximum in every one (`STICKY_SEEDS` raises the count). The same brute force confirms that a uniform group without racks, which skips the repair, has nothing for it to do. Rack locality after 25 rebalances under leader movement and member churn equals a fresh balance of the final cluster exactly. The pre-repair stickiness heuristics in the steal search were removed once the brute force showed they no longer change the result.

**Faster.** Per partition work that only needed doing per topic or per member is gone (subscription bitset, in-place compaction, per topic heaps, sorting only steal sources, average sized reservations), the steal search walks partition owners rather than partitions, and the complex path no longer scans every member per partition for a rack match. Numbers are in the commit messages; measured on this branch against master:

| shape | master | this branch |
|---|---|---|
| rejoin, 100 members, 1600 topics x 100 partitions | 351ms | 24ms |
| regex shaped, 500 members, 40 of 20,000 single partition topics | 176ms, 810MB | 12ms, 9MB |
| 2001 members, 500 topics x 2000 partitions, one narrow subscriber | 3.4s | 0.3s |
| same, with racks | 4.4s | 0.47s |
| 1001 members, 50 topics x 1000 partitions, one narrow subscriber | 94ms | 11ms |
| fresh uniform, 2000 members, 500 topics x 2000 partitions | 139ms | 141ms |
| same, with racks | 145ms | 169ms |

The last row is the repair's table build and rewrite over a million partitions, paid only with `BalanceRacks` on.

**Where to look hardest**, in order:

- `repair.go`: the level node model in `stickyFinder.find` and the rack weight on `stickyTable`, which is the partition count plus one.
- `repair.go`: the worked example at the top of the file, then `stickyTable`: a member has a slot in every rack's row of each of its classes, so rows and slots are arithmetic, and the finder only sees rows that hold partitions.
- `repair.go`: the `!moved && !t.crossed` early return, which skips rewriting the plan when no rotation fired and everybody holds what it arrived with.
- `sticky.go`: `origOwner` is captured after `tryRestickyStales`, since giving a partition back to an older generation's claimant changes who counts as having arrived with it.

**Changelog notes**, for the release changelog, which is not part of this PR:

* Rack aware assignment (KIP-881), which v1.21.0 turned on whenever `Rack` was set, now requires the new `BalanceRacks` option. A rack asks brokers to serve fetches from a nearby replica, which only does anything if the brokers run a rack aware replica selector; if they do and every rack has a replica, assigning partitions by the leader's rack gains nothing. The two are separate decisions now, so if you were relying on rack aware assignment, add `BalanceRacks()`. Partition racks are still built and handed to every `ConsumerBalancer` whenever a member reports a rack. The client warns at balance time if `BalanceRacks` is on while brokers are returning preferred read replicas.

* The sticky balancers now keep as many partitions in place as the balance allows, exactly. Balancing was already load optimal, but it moved whichever partition its search found first and never revisited a partition carried over from the previous assignment. With `BalanceRacks`, a partition whose leader moved to another rack was also never re-examined, so rack locality decayed toward what random placement gives as leadership moved. A repair pass after balancing now settles all three at once by cancelling improving rotations through a table of interchangeable partitions by members, which is a min-cost flow. Brute force over 90k small groups, some with racks and some past 64 members, finds the result optimal on all three. Rack placement outranks stickiness, so turning `BalanceRacks` on in a running group reassigns, at its next rebalance, every partition held by a member in a different zone from the partition's leader.

* Sticky balancing is faster, most of all on rejoins and on groups whose members subscribe to different topics. Against master, a rejoin of 100 members over 1600 topics of 100 partitions goes from 351ms to 24ms, a regex shaped group of 500 members over 20,000 topics from 176ms and 810MB to 12ms and 9MB, and 2001 members over 500 topics of 2000 partitions where one member subscribes narrowly from 3.4s to 0.3s. Fresh balances with uniform subscriptions are unchanged; with `BalanceRacks`, a million partitions across 2000 members pay about 15% for the repair.

* `AdjustCooperative` was quadratic in the number of partitions one member is planned of one topic. It is now linear.

### DIFF
--- a/pkg/kgo/config.go
+++ b/pkg/kgo/config.go
@@ -165,6 +165,7 @@ type cfg struct {
 	isolationLevel int8
 	keepControl    bool
 	rack           string
+	balanceRacks   bool
 	preferLagFn    PreferLagFn
 	decompressor   Decompressor
 
@@ -1669,14 +1670,45 @@ func ConsumeResetOffset(offset Offset) ConsumerOpt {
 	return consumerOpt{func(cfg *cfg) { cfg.resetOffset, cfg.setResetOffset = offset, true }}
 }
 
-// Rack specifies where the client is physically located and changes fetch
-// requests to consume from the closest replica as opposed to the leader
-// replica.
+// Rack specifies where the client is physically located.
 //
-// Consuming from a preferred replica can increase latency but can decrease
-// cross datacenter costs. See KIP-392 for more information.
+// This asks brokers to serve fetches from the closest replica rather than the
+// leader, which can increase latency but can decrease cross datacenter costs.
+// Brokers only honor the request if they are configured with a rack matching
+// replica.selector.class; by default a broker serves every fetch from the
+// leader and this has no effect on where fetches go. See KIP-392 for more
+// information.
+//
+// See [BalanceRacks] to also take racks into account when assigning
+// partitions to members.
 func Rack(rack string) ConsumerOpt {
 	return consumerOpt{func(cfg *cfg) { cfg.rack = rack }}
+}
+
+// BalanceRacks makes sticky balancing prefer giving a member partitions whose
+// leader is in the member's own rack, so that fetching them does not cross
+// zones. This requires [Rack] to also be specified; without a rack, there is
+// nothing to prefer and this does nothing.
+//
+// The preference is by partition leader because, by default, a broker serves
+// every fetch from the leader, so the leader's rack is what decides whether a
+// fetch crosses zones.
+//
+// If your brokers are configured with a rack matching replica.selector.class
+// (KIP-392), consumers instead fetch from a rack local replica. If your
+// replication factor is at least your rack count, every partition then already
+// has a replica local to every consumer, and balancing by rack moves
+// partitions without saving anything. Leave this off in that case.
+//
+// Kafka's Java client turns the equivalent on automatically whenever a rack is
+// configured, assuming a rack is only ever configured alongside a rack
+// matching replica selector (KIP-881). Nothing checks that assumption and it
+// does not hold for a default broker, so this client asks instead of assuming.
+//
+// Only the group leader's setting takes effect, as the leader is what computes
+// the assignment for the group.
+func BalanceRacks() ConsumerOpt {
+	return consumerOpt{func(cfg *cfg) { cfg.balanceRacks = true }}
 }
 
 // IsolationLevel controls whether uncommitted or only committed records are
@@ -1953,6 +1985,9 @@ func ShareAckCallback(fn func(*Client, ShareAckResults)) GroupOpt {
 // downgrade, the client revokes all partitions and re-consumes from
 // committed offsets, which can result in duplicates. It is not recommended
 // to downgrade once a group is cooperative.
+//
+// The sticky balancers can additionally take racks into account; see
+// [BalanceRacks].
 func Balancers(balancers ...GroupBalancer) GroupOpt {
 	return groupOpt{func(cfg *cfg) { cfg.balancers = balancers }}
 }
--- a/pkg/kgo/config_test.go
+++ b/pkg/kgo/config_test.go
@@ -134,3 +134,28 @@ func TestConfigRejectsInt32MillisOverflow(t *testing.T) {
 		}
 	}
 }
+
+// A rack says where the client is, which is what brokers need to serve a
+// closer replica. Chasing rack locality when assigning partitions is a
+// separate decision with its own cost, so it takes its own opt in -- unlike
+// Kafka's client, which turns it on whenever a rack is set.
+func TestConfigBalanceRacksIsOptIn(t *testing.T) {
+	cfg, _, err := validateCfg(ConsumerGroup("g"), Rack("az-1"))
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if cfg.rack != "az-1" {
+		t.Errorf("rack is %q, want az-1", cfg.rack)
+	}
+	if cfg.balanceRacks {
+		t.Error("a rack alone opted the group into rack aware balancing")
+	}
+
+	cfg, _, err = validateCfg(ConsumerGroup("g"), Rack("az-1"), BalanceRacks())
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if !cfg.balanceRacks {
+		t.Error("BalanceRacks did not opt the group into rack aware balancing")
+	}
+}
--- a/pkg/kgo/group_balancer.go
+++ b/pkg/kgo/group_balancer.go
@@ -457,10 +457,10 @@ func (g *groupConsumer) balanceGroup(proto string, members []kmsg.JoinGroupRespo
 		g.initExternal(topicPartitionCount)
 	}
 
-	// KIP-881: build partition rack info for rack-aware assignment.
-	// We use cached broker racks and partition leaders from local
-	// metadata. This requires no extra fetches.
-	if cb, ok := memberBalancer.(*ConsumerBalancer); ok {
+	// KIP-881: build partition rack info for rack-aware assignment, if
+	// the user asked for it. We use cached broker racks and partition
+	// leaders from local metadata, so this requires no extra fetches.
+	if cb, ok := memberBalancer.(*ConsumerBalancer); ok && g.cfg.balanceRacks {
 		cb.partitionRacks = g.buildPartitionRacks(cb, topicPartitionCount)
 	}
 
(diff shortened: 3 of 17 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,54 @@
+v1.22.0 (unreleased)
```
twmb: Drop the changelog, note it in the PR desc itself; I'll write the changelog before release.

pkg/kgo/config_test.go
```diff
@@ -134,3 +134,24 @@ func TestConfigRejectsInt32MillisOverflow(t *testing.T) {
 		}
 	}
 }
+
+func TestConfigBalanceRacksIsOptIn(t *testing.T) {
```
twmb: useless test?

pkg/kgo/internal/sticky/sticky_test.go
```diff
@@ -2610,3 +2611,21 @@ func TestRackAwareSparseRacksComplex(t *testing.T) {
 		t.Errorf("total partitions %d, want 8", total)
 	}
 }
+
+// Rack aware pre-assignment caps each member at the floor of an even share,
+// so that filling the rest by load leaves the plan within one level and
+// balancing has nothing to move.
+func TestRackAwareAssignStaysWithinOneLevel(t *testing.T) {
```
twmb: is this test useless?

pkg/kgo/internal/sticky/repair.go
```diff
@@ -0,0 +1,608 @@
+package sticky
+
+import (
+	"math"
+	"slices"
+)
+
```
twmb, line 7: Add a caveat at the top of this:
```
// NOTE: The repair algorithm and explanation below is LLM written;
// this was pursued via a session that explored open research on optimal
// ways to represent flows & stickiness. The test suite ensuring correctness
// and speed is ...extensive...

File: pkg/kgo/config.go [hunk 5]
```diff
@@ -1655,16 +1656,33 @@ func ConsumeResetOffset(offset Offset) ConsumerOpt {
...
-// cross datacenter costs. See KIP-392 for more information.
+// cross datacenter costs. See KIP-392 for more information, and see
+// [BalanceRacks] to also take racks into account when assigning partitions.
 func Rack(rack string) ConsumerOpt {
 	return consumerOpt{func(cfg *cfg) { cfg.rack = rack }}
 }
 
+// BalanceRacks makes the range and sticky balancers prefer giving a member
+// partitions whose leader is in the member's rack (KIP-881). [Rack] must also
+// be set, and only the group leader's setting matters since the leader
```
twmb: nix "only the group leader's setting matters" because the leader could be anybody.

pkg/kgo/config.go
```diff
@@ -1655,16 +1656,33 @@ func ConsumeResetOffset(offset Offset) ConsumerOpt {
...
+// computes the assignment.
+//
+// Brokers serve every fetch from the leader by default, so the leader's rack
+// decides whether a fetch crosses zones. If your brokers instead use a rack
+// aware replica selector (KIP-392), fetches are already served from a local
+// replica where one exists and rack aware balancing likely moves partitions
+// for nothing. Kafka's Java client balances by rack whenever a rack is set;
+// this client requires opting in.
```
twmb: Delete final sentence about java. Also, we may want a sentence on when a person wants this option, when a person wants to set only rack, and when a person wants to set neither.

pkg/kgo/group_balancer.go
```diff
@@ -466,11 +466,18 @@ func (g *groupConsumer) balanceGroup(proto string, members []kmsg.JoinGroupRespo
...
 
-	// KIP-881: build partition rack info for rack-aware assignment.
-	// We use cached broker racks and partition leaders from local
-	// metadata, which we refreshed above if we had to.
-	if cb, ok := memberBalancer.(*ConsumerBalancer); ok {
+	// KIP-881: build partition rack info for rack-aware assignment if
+	// the user opted in. We use cached broker racks and partition leaders
+	// from local metadata, which we refreshed above if we had to.
+	if cb, ok := memberBalancer.(*ConsumerBalancer); ok && g.cfg.balanceRacks {
 		cb.partitionRacks = g.buildPartitionRacks(cb, topicPartitionCount)
```
twmb, line 476: shouldn't we always set partitionRacks? balanceRacks is different.

pkg/kgo/group_balancer.go
```diff
@@ -466,11 +466,18 @@ func (g *groupConsumer) balanceGroup(proto string, members []kmsg.JoinGroupRespo
...
+
+		// A preferred read replica means the brokers run a rack aware
+		// replica selector and fetches are already rack local, so
+		// balancing by leader rack is moving partitions for nothing.
+		if g.cl.sawPreferredReplica.Load() {
+			g.cl.cfg.logger.Log(LogLevelWarn, "BalanceRacks is on but brokers are returning preferred read replicas; fetches are already rack local and rack aware balancing is likely moving partitions for nothing", "group", g.cfg.group)
```
twmb: "likely moving partitions" - clients aren't moving partitions; I think this means to say: "and rack aware balancing is assigning rack-local leader partitions for nothing"

pkg/kgo/group_balancer.go
```diff
@@ -560,6 +567,9 @@ func (b *ConsumerBalancer) anyMemberRack() bool {
 // updates keep those topics cached, since we ask for them, but a user's
 // metadata request for other topics evicts them (see storeCachedMeta).
 func (g *groupConsumer) needRackMeta(memberBalancer GroupMemberBalancer, topics map[string]struct{}) bool {
+	if !g.cfg.balanceRacks {
```
twmb: similar to above. Isn't this wrong? a balancer may want rack information for any odd reason. we should give the balancer the rack information even if balancers dont use it. the difference for us should come in whether we call Balance or whether we call BalanceWithRacks, I think...

CHANGELOG.md
twmb: Dropped the changelog commit. The notes are in the PR description under Changelog notes.

pkg/kgo/config_test.go
twmb: Agreed, it only checked that an option setter sets a field. Removed.

pkg/kgo/internal/sticky/sticky_test.go
twmb: Agreed. It pinned a performance property of the pre-assignment cap that balancing fixes anyway, and the brute force covers correctness. Removed.

pkg/kgo/internal/sticky/repair.go
twmb: Added at the top of the file, in your words plus what the tests check.

pkg/kgo/config.go
twmb: Removed. It now says to set both on every member, since whichever member leads the group computes the assignment.

pkg/kgo/config.go
twmb: Removed the Java sentence and added the three cases: neither, Rack only, and both.

pkg/kgo/group_balancer.go
twmb: Yes. Partition racks are now built whenever any member reports a rack, as on master, so custom balancers keep PartitionRacks(). balanceRacks is a field on ConsumerBalancer that only the range and sticky balancers check.

pkg/kgo/group_balancer.go
twmb: Reworded to your phrasing: rack aware balancing is assigning rack-local leader partitions for nothing. The comment above it and the option doc say the assignment gains nothing, for the same reason.

pkg/kgo/group_balancer.go
twmb: Yes, same fix. The gate is gone from needRackMeta, so rack metadata loads as it did on master since v1.21.7, and the difference is whether the sticky balancer calls Balance or BalanceWithRacks and whether the range balancer runs its rack phase.

PR #1387, opened 2026-08-17
Randomize initial seed broker selection

chailuecha:
## Summary

Randomize the starting seed broker index for each client.

Before this change, every new client started `anySeedIdx` at `0`, so clients with the same `SeedBrokers(...)` list all sent their first seed-broker request to the first configured broker. In large fleets restarting at roughly the same time, this can create a TCP connection spike on that first bootstrap broker.

I’m currently seeing this issue through the OpenTelemetry Collector Kafka exporter, which uses franz-go underneath. When many collectors restart around the same time with the same Kafka bootstrap broker list, the first configured broker receives a spike of TCP connections.

This keeps the configured seed broker order unchanged, but starts the private seed cursor at a random index per client. After that first pick, the existing round-robin behavior over seed brokers is unchanged.

## Test

Added a deterministic test that simulates 1000 clients using the same 4 seed brokers, and verifies the first selected seed is spread roughly evenly instead of always selecting `seed_0`.
```bash
go test ./pkg/kgo -run TestSeedBrokersStartDistributedAcrossClients

### DIFF
--- a/pkg/kgo/client.go
+++ b/pkg/kgo/client.go
@@ -597,6 +597,7 @@ func NewClient(opts ...Opt) (*Client, error) {
 		b := cl.newBroker(unknownSeedID(i), seed.host, seed.port, nil)
 		seedBrokers = append(seedBrokers, b)
 	}
+	cl.anySeedIdx = cl.randomSeedIdx(len(seedBrokers))
 	cl.seeds.Store(seedBrokers)
 	go cl.updateMetadataLoop()
 	go cl.reapConnectionsLoop()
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by chailuecha** [comment]
Hi @twmb , not sure if I could get your help in triggering review process. 
Let me know if the MR is not right or I need to follow any process / doc. 

Thanks a ton.
Chatchai C.

**Comment by twmb** [comment]
I'm fine to merge. Mind dropping the test? It doesn't add much here. Also in UpdateSeedBrokers, this line of the doc comment can now be deleted:

```
we do not need
	// the lock for what this usually guards.
```

twmb:
it'll be in the next release.

PR #1388, opened 2026-08-18
deps: bump github.com/klauspost/compress to v1.19.2

GavinDmello:
Bumping compress as it has a fix for an issue with zstd compression

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/twmb/franz-go
 go 1.25.0
 
 require (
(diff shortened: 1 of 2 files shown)


twmb:
I can merge now, sure - but I do bump all deps before releases. Since this is a library, individual dep bumps aren't necessary here itself, since end users can set the min version in their bins. But I also get that this is probably showing up in some vuln scanner because vuln scanners aren't always great about min version bumping in the binary...

PR #1392, opened 2026-08-26
kgo: expose BalanceInfo for custom balancer implementations

michaelwilner:
This is the `FirstBalancer`, an eager group balancer (protocol `"first"`) that assigns all partitions to the group leader. Everyone else joins as a hot standby and gets nothing. When the active member leaves, everything moves to the newly elected leader. Since the coordinator keeps the same leader until it leaves, standbys joining or leaving never migrate partitions. Failover only happens when the active member itself goes away.

We run this in our system in production and have for several months. Some of our services need exactly one instance consuming all partitions (stateful processing where the one instance processing messages has to control the universe) with standbys ready for fast failover where rebalancing is cheap and fast but startup cost is slow by comparison. The group protocol already does leader election, so we lean on that instead of running an external lock service. We built it on franz-go's public balancer interfaces (modeled closely on the roundrobin balancer) and figured the community would benefit from having it upstream.

For this changeset to work, balancers need to see the leader ID through the `GroupBalancer` interface, so `balanceGroup` now passes it into an unexported field on `ConsumerBalancer` (same spot as the KIP-881 rack injection). No public API changes. If no leader is known (external use for example), it falls back to the first member sorted by instance ID then member ID which is always at index 0. That way there's always a member that has all the partitions even if the leader cannot be dereferenced.

### DIFF
--- a/pkg/kgo/group_balancer.go
+++ b/pkg/kgo/group_balancer.go
@@ -6,6 +6,7 @@ import (
 	"slices"
 	"sort"
 	"strings"
+	"sync"
 
 	"github.com/twmb/franz-go/pkg/kerr"
 	"github.com/twmb/franz-go/pkg/kgo/internal/sticky"
@@ -84,6 +85,45 @@ type GroupMemberBalancerOrError interface {
 	BalanceOrError(topics map[string]int32) (IntoSyncAssignment, error)
 }
 
+// BalanceInfo contains information about the group being balanced and the
+// cluster it is in. We set this on balancers that implement
+// GroupMemberBalancerInfo, before we balance.
+type BalanceInfo struct {
+	// Group is the group being balanced.
+	Group string
+
+	// Generation is the generation we are balancing into. Each member's
+	// metadata contains the generation that member last knew of.
+	Generation int32
+
+	// LeaderID is the group leader's member ID, which is always you: the
+	// coordinator only asks the leader to balance. The coordinator keeps
+	// the same leader until that member leaves the group, so pinning work
+	// to the leader is stable while other members join and leave.
+	LeaderID string
+
+	// Topics returns metadata for every topic being balanced. The first
+	// call builds the result, later calls return the same map.
+	//
+	// We do not request metadata to answer this: topics we have not
+	// loaded yet are missing, and what we have can be up to MetadataMaxAge
+	// old.
+	Topics func() map[string]TopicMetadata
+
+	// Brokers returns every broker we know of, by node ID. Use this to map
+	// a partition's leader or replicas to a rack.
+	Brokers func() map[int32]BrokerMetadata
+}
+
+// GroupMemberBalancerInfo is an optional extension interface for
+// GroupMemberBalancer. If your balancer implements this, we set the info for
+// the group being balanced before balancing. ConsumerBalancer implements
+// this; see ConsumerBalancer.Info.
+type GroupMemberBalancerInfo interface {
+	GroupMemberBalancer
+	SetBalanceInfo(BalanceInfo)
+}
+
 // IntoSyncAssignment takes a balance plan and returns a list of assignments to
 // use in a kmsg.SyncGroupRequest.
 //
@@ -105,6 +145,12 @@ type ConsumerBalancer struct {
 	// partitionRacks maps topic => partition index => leader rack.
 	// nil when rack-aware assignment is not active.
 	partitionRacks map[string][]string
+
+	// info is everything we know about the group and cluster at balance
+	// time, set by the client before balancing (see balanceGroup). Zero
+	// when the balancer is constructed directly (NewConsumerBalancer) and
+	// balanced without a client.
+	info BalanceInfo
 
 	err error
 }
@@ -149,6 +195,20 @@ func (b *ConsumerBalancer) SetError(err error) {
 // if no rack info is available.
 func (b *ConsumerBalancer) PartitionRacks() map[string][]string {
 	return b.partitionRacks
+}
+
+// Info returns information about the group being balanced and the cluster it
+// is in. The client sets the info before balancing; if you construct a
+// ConsumerBalancer directly with NewConsumerBalancer and balance without a
+// client, the info is empty: LeaderID is unset and the Topics and Brokers
+// functions are nil.
+func (b *ConsumerBalancer) Info() BalanceInfo {
+	return b.info
+}
+
+// SetBalanceInfo implements GroupMemberBalancerInfo.
+func (b *ConsumerBalancer) SetBalanceInfo(info BalanceInfo) {
+	b.info = info
 }
 
 // MemberTopics returns the unique set of topics that all members are
@@ -386,7 +446,7 @@ func (g *groupConsumer) findBalancer(from, proto string) (GroupBalancer, error) 
 // returns all topics and partitions; the leader will then periodically do its
 // own metadata update to see if partition counts have changed for these random
 // topics.
-func (g *groupConsumer) balanceGroup(proto string, members []kmsg.JoinGroupResponseMember, skipBalance bool) ([]kmsg.SyncGroupRequestGroupAssignment, error) {
+func (g *groupConsumer) balanceGroup(proto string, resp *kmsg.JoinGroupResponse) ([]kmsg.SyncGroupRequestGroupAssignment, error) {
 	g.cl.cfg.logger.Log(LogLevelInfo, "balancing group as leader")
 
 	b, err := g.findBalancer("balance group", proto)
@@ -394,6 +454,7 @@ func (g *groupConsumer) balanceGroup(proto string, members []kmsg.JoinGroupRespo
 		return nil, err
 	}
 
+	members := resp.Members
 	sortJoinMembers(members)
 
 	memberBalancer, topics, err := b.MemberBalancer(members)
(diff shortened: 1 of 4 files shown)


twmb:
I'm wary to add the "first" balancer to the package directly, since it's a non-standard balancer with a niche use case. That said, I also see that you can't implement it as is with the current API, since you need the leader. I've been workshopping some API change proposals with claude for a bit, the next comment will include a proposed API that should make what you want to do possible. Please review and let me know what you think - and, if you're on board, what you think about not merging the first balancer.

twmb:
Here is the proposed API. It is purely additive: no existing interface or
signature changes, and every balancer that exists today keeps working untouched.

**The problem.** A balancer sees the group's members and their metadata, and a
`map[string]int32` of topic to partition count. Anything we have added since has
been bolted onto `ConsumerBalancer` as an unexported field with an accessor.
That is what rack info for KIP-881 is today. The leader ID would be the second
one, replicas the third. Rather than add another, we give balancers one place to
receive everything we know at balance time.

```go
// BalanceInfo contains information about the group being balanced and the
// cluster it is in. We set this on balancers that implement
// GroupMemberBalancerInfo, before we balance.
type BalanceInfo struct {
	// Group is the group being balanced.
	Group string

	// Generation is the generation we are balancing into. Each member's
	// metadata contains the generation that member last knew of.
	Generation int32

	// LeaderID is the group leader's member ID, which is always you: the
	// coordinator only asks the leader to balance. The coordinator keeps
	// the same leader until that member leaves the group, so pinning work
	// to the leader is stable while other members join and leave.
	LeaderID string

	// Topics returns metadata for every topic being balanced. The first
	// call builds the result, later calls return the same map.
	//
	// We do not request metadata to answer this: topics we have not
	// loaded yet are missing, and what we have can be up to MetadataMaxAge
	// old.
	Topics func() map[string]TopicMetadata

	// Brokers returns every broker we know of, by node ID. Use this to map
	// a partition's leader or replicas to a rack.
	Brokers func() map[int32]BrokerMetadata
}

// GroupMemberBalancerInfo is an optional extension interface for
// GroupMemberBalancer. If your balancer implements this, we set the info for
// the group being balanced before balancing. ConsumerBalancer implements
// this; see ConsumerBalancer.Info.
type GroupMemberBalancerInfo interface {
	GroupMemberBalancer
	SetBalanceInfo(BalanceInfo)
}

func (b *ConsumerBalancer) Info() BalanceInfo
func (b *ConsumerBalancer) SetBalanceInfo(info BalanceInfo)
```

`TopicMetadata` and `PartitionMetadata` are new general types that mirror
`kadm.TopicDetail` and `kadm.PartitionDetail`, alongside the existing
`kgo.BrokerMetadata`:

```go
type PartitionMetadata struct {
	Topic     string
	Partition int32

	Leader          int32   // the broker leader, if there is one, otherwise -1
	LeaderEpoch     int32
	Replicas        []int32
	ISR             []int32
	OfflineReplicas []int32

	Err error // non-nil if the partition currently has a load error
}

type TopicMetadata struct {
	Topic      string
	ID         [16]byte // all zero if the broker does not support topic IDs
	Partitions []PartitionMetadata
	Err        error
}
```

**What this means for your balancer.** You get the leader, so the balancer works
unchanged outside the package, and it gets simpler than the version here: we
always set `LeaderID`, so the sorted-first-member fallback is only reachable if
you build a `ConsumerBalancer` with `NewConsumerBalancer` and balance without a
client.

```go
func (*firstBalancer) Balance(b *kgo.ConsumerBalancer, topics map[string]int32) kgo.IntoSyncAssignment {
	plan := b.NewPlan()
	leader := b.Info().LeaderID
	idx := slices.IndexFunc(b.Members(), func(m kmsg.JoinGroupResponseMember) bool {
		return m.MemberID == leader
	})
	if idx < 0 {
		return plan // no leader known: assign nothing and rebalance again
	}
	member, meta := b.MemberAt(idx)
	for _, topic := range meta.Topics {
		for p := int32(0); p < topics[topic]; p++ {
			plan.AddPartition(member, topic, p)
		}
	}
	return plan
}
```

`Topics` and `Brokers` are functions because building the snapshot costs roughly
1.3MB and 700us for a group with 10k partitions, and sticky, range, and
roundrobin never read it; balancers that do not ask should not pay. Neither
issues a metadata request. Balancing runs inside the join, so you get what the
client has already cached.

Does this cover what you need, and is anything missing for the standby setup you
described? If you are on board, are you alright with the balancer itself staying
in your repo? And if you would like to implement this change, it is yours.

michaelwilner:
> I'm wary to add the "first" balancer to the package directly, since it's a non-standard balancer with a niche use case. That said, I also see that you can't implement it as is with the current API, since you need the leader. I've been workshopping some API change proposals with claude for a bit, the next comment will include a proposed API that should make what you want to do possible. Please review and let me know what you think - and, if you're on board, what you think about not merging the first balancer.

Hey @twmb thanks for the feedback! This makes sense, I'll shift gears here.

michaelwilner:
@twmb updated, the API you proposed is implemented. Let me know if the underlying implementation is kosher here. Pulled the fork and things work as anticipated here.

pkg/kgo/group_balancer.go
```diff
@@ -464,6 +525,33 @@ func (g *groupConsumer) balanceGroup(proto string, members []kmsg.JoinGroupRespo
...
 	}
 
+	// Balancers that opt in receive everything we know about the group and
+	// cluster at balance time. Topics and Brokers are functions so that
+	// balancers that never read them (sticky, range, roundrobin) do not
+	// pay for building the snapshots.
+	if ib, ok := memberBalancer.(GroupMemberBalancerInfo); ok {
+		var balanceInfoTopicsBuilder, balanceInfoBrokersBuilder sync.Once
+		var balanceInfoTopics map[string]TopicMetadata
+		var balanceInfoBrokers map[int32]BrokerMetadata
```
twmb, line 535: nit:
```
var (
    topicsBuilder, brokersBuilder ...
    topics ...
    brokers ...
)
````
no prefix needed esp. since these are method local

File: pkg/kgo/group_balancer.go [hunk 2]
```diff
@@ -587,6 +675,56 @@ func (g *groupConsumer) buildPartitionRacks(b *ConsumerBalancer, topicPartitionC
...
+		for i := range ct.t.Partitions {
+			p := &ct.t.Partitions[i]
+			t.Partitions = append(t.Partitions, PartitionMetadata{
+				Topic:           topic,
+				Partition:       p.Partition,
+				Leader:          p.Leader,
+				LeaderEpoch:     p.LeaderEpoch,
+				Replicas:        slices.Clone(p.Replicas),
+				ISR:             slices.Clone(p.ISR),
+				OfflineReplicas: slices.Clone(p.OfflineReplicas),
```
twmb, line 704: You could create one slice that is the len of replicas + isr + offline replicas, and then sub-slice that (with cap) to reduce allocs here and speed it up a little bit.

pkg/kgo/group_balancer.go
```diff
@@ -587,6 +675,56 @@ func (g *groupConsumer) buildPartitionRacks(b *ConsumerBalancer, topicPartitionC
 	return nil
 }
 
+// balanceInfoTopics builds the BalanceInfo.Topics map of topic metadata
+func (cl *Client) balanceInfoTopics(topics map[string]struct{}) map[string]TopicMetadata {
+	m := make(map[string]TopicMetadata, len(topics))
+	cl.metaCache.mu.Lock()
```
twmb, line 681: This holds `metaCache.mu` for the whole deep copy, which is fine, but could be a bit more optimized: releasing the mu early is safe because the cache replaces entries wholesale rather than modifying them, so the slices captured here cannot change while we copy. This first form is slightly more alloc efficient and, heck while we're having an llm write it, why not:

```go
// balanceInfoTopics returns the topic metadata for BalanceInfo.Topics. We
// copy cache entries under the lock and deep copy outside it: the cache
// replaces entries wholesale rather than modifying them, so what we captured
// cannot change while we copy.
func (cl *Client) balanceInfoTopics(topics map[string]struct{}) map[string]TopicMetadata {
	cached := make(map[string]cachedMetaTopic, len(topics))
	cl.metaCache.mu.Lock()
	for topic := range topics {
		if ct, ok := cl.metaCache.topics[topic]; ok {
			cached[topic] = ct
		}
	}
	cl.metaCache.mu.Unlock()

	var nreplicas, nparts int
	for _, ct := range cached {
		nparts += len(ct.t.Partitions)
		for i := range ct.t.Partitions {
			p := &ct.t.Partitions[i]
			nreplicas += len(p.Replicas) + len(p.ISR) + len(p.OfflineReplicas)
		}
	}

	// One backing slice for every partition's replica lists, and one for
	// every topic's partitions. Both are capped when sub-sliced, so a
	// caller appending to one copies rather than overwriting a neighbor.
	replicas := make([]int32, 0, nreplicas)
	parts := make([]PartitionMetadata, 0, nparts)
	sub := func(src []int32) []int32 {
		if len(src) == 0 {
			return nil
		}
		at := len(replicas)
		replicas = append(replicas, src...)
		return replicas[at:len(replicas):len(replicas)]
	}

	m := make(map[string]TopicMetadata, len(cached))
	for topic, ct := range cached {
		at := len(parts)
		for i := range ct.t.Partitions {
			p := &ct.t.Partitions[i]
			parts = append(parts, PartitionMetadata{
				Topic:           topic,
				Partition:       p.Partition,
				Leader:          p.Leader,
				LeaderEpoch:     p.LeaderEpoch,
				Replicas:        sub(p.Replicas),
				ISR:             sub(p.ISR),
				OfflineReplicas: sub(p.OfflineReplicas),
				Err:             kerr.ErrorForCode(p.ErrorCode),
			})
		}
		tps := parts[at:len(parts):len(parts)]
		sort.Slice(tps, func(i, j int) bool { return tps[i].Partition < tps[j].Partition })
		m[topic] = TopicMetadata{
			Topic:      topic,
			ID:         ct.id,
			Partitions: tps,
			Err:        kerr.ErrorForCode(ct.t.ErrorCode),
		}
	}
	return m
}
```

slightly simpler, without as much subslicing:

```go
func (cl *Client) balanceInfoTopics(topics map[string]struct{}) map[string]TopicMetadata {
	cached := make(map[string]cachedMetaTopic, len(topics))
	cl.metaCache.mu.Lock()
	for topic := range topics {
		if ct, ok := cl.metaCache.topics[topic]; ok {
			cached[topic] = ct
		}
	}
	cl.metaCache.mu.Unlock()

	m := make(map[string]TopicMetadata, len(cached))
	for topic, ct := range cached {
		// ... unchanged from here
	}
	return m
}
```

michaelwilner: Opted for the simpler form for brevity https://github.com/twmb/franz-go/pull/1392/commits/ca1ce3c999ac8d40a2b2c06d4652fffcf5d9dcb7

pkg/kgo/group_balancer.go
michaelwilner: https://github.com/twmb/franz-go/pull/1392/commits/ca1ce3c999ac8d40a2b2c06d4652fffcf5d9dcb7

michaelwilner: `topics` was already in use, used topicsMdBuilder, brokersMdBuilder, topicsMd, brokersMd

michaelwilner:
@twmb should be ready for another pass, rebased

twmb:
lgtm, will wait for a minor release (trying to review my bigger claude-generated PRs and improve them)

twmb (approved):

twmb:
Ideally release coming within the week, merging now in prep

PR #1364, opened 2026-07-04
kgo: handle topic recreation on brokers 2.8 and above

twmb:
A topic deleted and recreated under the same name is a new topic that happens to share the name. This PR makes the client treat it that way, by default, on every broker from 2.8 on: nothing from the old topic, consume positions, produce sequences, transaction membership, or share acquisition state, is reused against the new one. Eight commits, each reviewable on its own; the README gains a short user-facing section and the CHANGELOG the full description.

## Detection

Topic IDs are random and never reused, so metadata reporting an ID this partition has never held can only mean a recreation, and the metadata merge adopts it at once. An ID the partition previously held is a lagging broker's view and is refused: a broker rejecting the ID we hold says only that its own view differs, which a leader that has not yet learned a recreation says about the correct ID too. Only after the held ID has been rejected for a whole grace of paced retries, longer than metadata takes to propagate, is a prior ID that metadata keeps reporting believed instead.

- **3.1+**: fetches carry the topic ID, so a fetch of the old topic is rejected rather than misread. No silent window.
- **2.8 through 3.0**: fetches go by name and can read the new topic until the next metadata update. Fetched records carrying a leader epoch below what was already consumed are withheld, paced, until metadata decides; the withhold is bounded so a rolled-back log still delivers.
- **Below 2.8**: no topic IDs; behavior unchanged and documented as unsupported.

A partition with no position yet waits for the position before swapping, so a racing old-topic committed offset is rejected rather than applied to the new topic.

## On detection

- **Consumers restart from the new topic's beginning**: a subscription is a point in time and everything after it. `NoResetOffset` instead surfaces an error and waits for `SetOffsets`. An unconsumed restart re-resolves the beginning on a further out of range (recreate then truncate).
- **The group's uncommitted offsets for the old topic are dropped at the swap**, so a commit after the swap never carries them, and a fetch of the old topic taken just before the swap is skipped by topic ID. The broker deletes the old commits with the topic.
- **Idempotent producers** adopt the new ID and restart their sequence chain with no sequence error surfaced and no duplicate possible; when a produce response proves the chain already landed in the new log, the chain continues instead. A by-name batch whose outcome is unknowable fails loudly. A produce flipping back to a prior ID re-initializes the producer ID rather than restarting sequences, since the old log remembers the chain.
[...]

PR #1399, opened 2026-09-01
DESIGN: rewrite in plain prose

twmb:
Stacked on #1364 so the diff is only DESIGN.md; merge after it, or retarget
to master once #1364 lands.

Every line of DESIGN.md blamed to 2026 and read as generated. This cuts it
from 1419 lines to 818 in our own voice: no reading guide, no table of
contents, no bold step labels, no narration. The mermaid diagrams (11 of 14;
three that restated a sentence are gone), the connection and file tables, and
the invariants list stay, since those carry the information. The topic
recreation section is trimmed to match.

I wrote this from the generated text, not from a fresh read of every code
path it describes, so it deserves the read the original never got.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,59 @@
+Unreleased
+===
+
+This release handles topic recreation, a topic deleted and recreated under
+the same name, across the whole client, by default, at every broker version.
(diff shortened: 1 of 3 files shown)


twmb:
Superseded: DESIGN.md is being removed instead (see the remove-design PR).

PR #1400, opened 2026-09-01
remove DESIGN.md

twmb:
Every line of it was generated and none of it was ever reviewed, yet pkg/kgo/CLAUDE.md pointed every session at it as the design reference and told them to keep it updated. The invariants it listed already live as comments beside the code they describe. #1364 no longer touches the file, so it rebases cleanly over this.


twmb:
Deleting because it was entirely LLM generated, and I don't think it's useful to maintain. Could be added later, but by hand with LLM guidance would be better.

PR #1403, opened 2026-09-03
kotel: preallocate trace attribute slices

yianni:
The trace hooks build their fixed attributes in a slice with no spare capacity, then append optional attributes. The first append allocates a larger backing array and copies the slice.

This change preallocates the required capacity and shares attribute construction between receive and process. The API and emitted telemetry are unchanged.

These benchmarks use a no-op tracer provider, a 19-byte key, a client ID, and a consumer group:

| Path | Before | After |
|---|---:|---:|
| Produce | 1,056 B/op, 9 allocs/op | 672 B/op, 8 allocs/op |
| Fetch | 1,584 B/op, 10 allocs/op | 944 B/op, 9 allocs/op |
| Process | 1,440 B/op, 8 allocs/op | 800 B/op, 7 allocs/op |

A ten-minute randomized A/B test against a local Kafka broker ran 243 pairs. Each run produced and consumed 20,000 records using `AlwaysSample` and a `BatchSpanProcessor`. The change allocated about 20.5 MB less and made about 40,000 fewer allocations per run. Median paired elapsed time was 1.75% lower, although timing will vary by workload.

### DIFF
--- a/plugin/kotel/tracer.go
+++ b/plugin/kotel/tracer.go
@@ -99,24 +99,99 @@ func NewTracer(opts ...TracerOpt) *Tracer {
 	return t
 }
 
-func (t *Tracer) maybeKeyAttr(attrs []attribute.KeyValue, r *kgo.Record) []attribute.KeyValue {
+func (t *Tracer) keyAttr(r *kgo.Record) (attribute.KeyValue, bool) {
 	if r.Key == nil {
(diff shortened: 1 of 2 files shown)


twmb (approved):

twmb:
thanks! released in kotel v1.7.1

PR #1405, opened 2026-09-03
kfake: per-partition fault injection

twmb:
## AddFault

`c.AddFault(kfake.Fault{...})` answers matching partitions of matching requests with a chosen error and answers the rest of each request normally.

A request matches on its key (`Keys`, nil for any supported kind) and on the node it arrived at (`Node`, -1 for any); a partition matches on `Topic` or `TopicID` plus `Partition` (-1 for any). The zero values of `Node` and `Partition` are node 0 and partition 0, not "any", since kfake uses raw IDs everywhere. `Count` is a budget of requests: a matching request consumes one and fails every partition in it that matches, and an exhausted fault removes itself. The returned `*FaultHandle` has `Fired()` (requests answered, safe from any goroutine) and `Remove()` (safe twice, and after exhaustion).

Faults apply at the point each handler emits a per-partition error code, so the rest of the request stays real: Produce, Fetch, ListOffsets, OffsetForLeaderEpoch, ShareFetch, and ShareAcknowledge. A faulted partition also completes a fetch or share fetch at once instead of holding it for MaxWait, which is what any errored partition does. Because faults live inside handlers, a `ControlKey` control that answers a request itself bypasses them, while one that passes through sees them.

Why: matching by topic ID. After a delete and recreate, a client can keep addressing the stale ID, and a test has to fail only the requests that use it; until now the only way to fail one partition was a control function rebuilding the whole response by hand.

## Converting existing tests

Six controls in the kfake tests existed only to answer a request's partitions with an error code, each rebuilding a whole response by hand to say that one thing; they are now `AddFault` calls. That removes 121 lines and adds 57, and 68 of the removed lines were building kmsg response entries. Two responses change shape and nothing asserted on the difference: the fenced fetch no longer carries the high watermark, last stable offset, and log start offset the control put on an errored partition, and the not-leader share fetch is now the handler's own response, whose leader hint is the -1/-1 the control was setting by hand. The controls left alone are not per-partition errors: they inject preferred read replicas, a negative offset, or the KIP-320 undefined-epoch sentinel, they fail a whole request or close the connection, they alternate the error code per request, or the test synchronizes on a channel the control closes.

## Tested

New tests: a produce budget of 2 (the third attempt succeeds, `Fired() == 2`, the record lands once), a fetch fault on one topic ID alongside an unfaulted topic in the same request (faulted partition errors, the other returns records, the response is immediate, exhaustion frees it), `Keys` restricting a fault to produce while a fetch of the same partition succeeds, `Remove` stopping a budgetless fault, and a node-pinned fault firing only at that node versus `Node: -1` firing at both.

`gofmt -l pkg/kfake` clean, `go vet`, the fault tests and every converted test with `-race -count=10`, the persist tests with `-race`, and the full package with `-race -skip ETL`.

This branch previously also carried a FreezeNode/UnfreezeNode feature; that work was dropped and only the fault injection remains.

### DIFF
--- a/pkg/kfake/00_produce.go
+++ b/pkg/kfake/00_produce.go
@@ -95,6 +95,7 @@ func (c *Cluster) handleProduce(creq *clientReq) (kmsg.Response, error) {
 		return toresp(), nil
 	}
 
+	fc := c.faultsFor(creq)
 	now := time.Now().UnixMilli()
 	for _, rt := range req.Topics {
 		if req.Version >= 13 {
@@ -111,6 +112,10 @@ func (c *Cluster) handleProduce(creq *clientReq) (kmsg.Response, error) {
 		}
 		maxMessageBytes := c.data.maxMessageBytes(rt.Topic)
 		for _, rp := range rt.Partitions {
+			if e := fc.err(rt.Topic, rt.TopicID, rp.Partition); e != nil {
+				donep(rt, rp, e.Code, e.Message)
+				continue
+			}
 			pd, ok := c.data.tps.getp(rt.Topic, rp.Partition)
 			if !ok {
 				donep(rt, rp, kerr.UnknownTopicOrPartition.Code, "Unknown topic or partition.")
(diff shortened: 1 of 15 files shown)


twmb:
Superseded by #1406: the branch was renamed to `kfake-faults` (the rename closed this PR), and the FreezeNode work was dropped. Same two commits, rebuilt on master.

PR #1406, opened 2026-09-04
kfake: per-partition fault injection

twmb:
## Fault

`c.Fault(faults...)` installs faults and returns one handle for the batch. Every field of a `Fault` is an AND filter over what it applies to: `Keys` (request kinds), `Nodes`, `Topic` or `TopicID`, `Partitions`, `Group`, `TxnID`, `Resource` (a config resource, quota entity, SCRAM user, log dir, feature, LeaveGroup member, or ACL resource name), and `TopLevel` to fail the request's top-level error code instead of its entities. Every zero value means "all", so node 0 and partition 0 are named the same way as any other, and a selector a request does not carry never matches: a fault with a `Topic` does not fail a heartbeat, and one with `Partitions` does not match an entity that has no partition. `Fault` may be called from a control function.

`Count` is a budget of requests: 0 is one, -1 is until `Remove`, N is N. Faults are checked in the order added and the first match answers an entity; a `ControlKey` control that answers a request bypasses them. The handle has `Hits`, `Wait(ctx, n)`, and `Remove`.

`REQUEST_TIMED_OUT` is answered after the work rather than in place of it on exactly the requests a real broker can answer that way, and `NOT_ENOUGH_REPLICAS_AFTER_APPEND` on Produce and WriteTxnMarkers. A table in faults.go holds the set, all per entity: Produce, WriteTxnMarkers, DeleteRecords, OffsetCommit, TxnOffsetCommit, DeleteGroups, AlterShareGroupOffsets, DeleteShareGroupOffsets, and the forwarded admin requests whose controller work can finish after the broker's forwarding budget expires (CreateTopics, DeleteTopics, CreatePartitions, ElectLeaders, AlterPartitionAssignments, AlterConfigs, IncrementalAlterConfigs, CreateACLs, DeleteACLs, AlterClientQuotas, AlterUserSCRAMCredentials, UpdateFeatures). Everywhere else -- the group protocol, the transaction coordinator, the share requests, every read -- the code is an ordinary rejection. So a faulted produce still appends, a timed out delete still deletes, a timed out commit is still stored, and a timed out create still creates, each with a test.

**Coverage.** Faults are checked wherever authorization is (one helper beside every ACL check, in the numbered handlers and in groups.go, share_groups.go and txns.go), plus per partition inside every request that carries partitions, and at dispatch for the eight requests that carry nothing to select on (the entityless set in faults.go). A table test walks every request key kfake registers, sends a minimal request of that kind, and asserts the response carries the injected code and that the fault fired -- so no kind can silently miss.

Why: matching by topic ID, group, or transaction. After a delete and recreate a client can keep addressing the stale ID, and a test has to fail only the requests that use it; until now the only way to fail one partition was a control function rebuilding the whole response by hand.

## Converting existing tests

Nine controls in the kfake tests existed only to answer a request with an error code; they are now faults. That removes 216 lines and adds 120, and 97 of the removed lines were building kmsg response entries. Five become per-partition faults (produce, fetch, list offsets, offset for leader epoch, share fetch), two become `TopLevel` faults with the control kept only for its timestamps, and two that blocked on a channel the control closed now wait on the handle with `Wait`. The controls left alone are not error injection: preferred read replicas, a negative offset, the KIP-320 undefined-epoch sentinel, a duplicated partition entry, a closed connection, an error code that alternates per request, a request held in flight, and pure observers.

## Tested

The coverage table over every registered key; a partition selector that fails one partition and leaves the rest; a group selector across commits, heartbeats, and a describe; `TopLevel` failing a fetch at once with no partitions; a transaction selector across InitProducerID, EndTxn, and TxnOffsetCommit; a node selector on a two node cluster; FindCoordinator matching a group key against `Group` and a transaction key against `TxnID`; `Resource` on DescribeConfigs and on CreateACLs; a topic fault that does not touch a heartbeat; the `Count` budget and `Remove`; a batch handle summing hits and removing both faults; the earliest fault winning; the after-apply cases (produce three ways, delete records, offset and transactional commits, delete groups, create topics, alter configs); faults installed from inside a control function, plain and sleeping; and `Wait` before, during, at exhaustion, and on a canceled context.

`gofmt -l pkg/kfake` clean, `go vet`, the fault tests and every converted test with `-race -count=10`, the persist tests with `-race`, and the full package with `-race -skip ETL`.

## Maintenance

pkg/kfake/CLAUDE.md gains a section on adding a request handler: where the deny and check calls go, what the fault key names, the answer-then-skipsWork emission shape, the after-apply table, and the tests that must pass. nilerr is excluded for the numbered handlers: they answer an error code inside the response and return nil to the connection, which the linter reads as a swallowed error.

### DIFF
--- a/pkg/kfake/00_produce.go
+++ b/pkg/kfake/00_produce.go
@@ -45,14 +45,21 @@ func (c *Cluster) handleProduce(creq *clientReq) (kmsg.Response, error) {
 		return nil, err
 	}
 
+	// A fault can answer a partition before the work runs. The work's own
+	// answer for that partition must not add an entry or replace the code.
+	answered := make(map[tp]int)
 	donep := func(t kmsg.ProduceRequestTopic, p kmsg.ProduceRequestTopicPartition, errCode int16, errMsg string) *kmsg.ProduceResponseTopicPartition {
+		ps := tdone[id(t)]
+		if i, ok := answered[tp{t.Topic, p.Partition}]; ok {
+			return &ps[i]
+		}
+		answered[tp{t.Topic, p.Partition}] = len(ps)
 		sp := kmsg.NewProduceResponseTopicPartition()
 		sp.Partition = p.Partition
 		sp.ErrorCode = errCode
 		if req.Version >= 8 && errMsg != "" {
 			sp.ErrorMessage = &errMsg
 		}
-		ps := tdone[id(t)]
 		ps = append(ps, sp)
 		tdone[id(t)] = ps
 		return &ps[len(ps)-1]
@@ -105,12 +112,20 @@ func (c *Cluster) handleProduce(creq *clientReq) (kmsg.Response, error) {
 			}
 			rt.Topic = topic
 		}
-		if !c.allowedACL(creq, rt.Topic, kmsg.ACLResourceTypeTopic, kmsg.ACLOperationWrite) {
-			donet(rt, kerr.TopicAuthorizationFailed.Code, kerr.TopicAuthorizationFailed.Message)
+		tk := faultKey{topic: rt.Topic, topicID: rt.TopicID}
+		if e := c.deny(creq, rt.Topic, kmsg.ACLResourceTypeTopic, kmsg.ACLOperationWrite, tk); e != nil && creq.skipsWork(e) { // a timed-out append falls through to the per-partition checks
+			donet(rt, e.Code, e.Message)
 			continue
 		}
 		maxMessageBytes := c.data.maxMessageBytes(rt.Topic)
 		for _, rp := range rt.Partitions {
+			e := creq.faults.check(tk.part(rp.Partition))
+			if e != nil {
+				donep(rt, rp, e.Code, e.Message)
+				if creq.skipsWork(e) { // a timed-out append still appends
+					continue
+				}
+			}
 			pd, ok := c.data.tps.getp(rt.Topic, rp.Partition)
 			if !ok {
 				donep(rt, rp, kerr.UnknownTopicOrPartition.Code, "Unknown topic or partition.")
--- a/pkg/kfake/78_share_fetch.go
+++ b/pkg/kfake/78_share_fetch.go
@@ -48,10 +48,11 @@ func (c *Cluster) handleShareFetch(creq *clientReq, w *watchShareFetch) (kmsg.Re
 	}
 
 	resp.AcquisitionLockTimeoutMillis = c.shareRecordLockDurationMs()
+	fc := creq.faults
 
 	// ACL: require GROUP READ.
-	if !c.allowedACL(creq, groupID, kmsg.ACLResourceTypeGroup, kmsg.ACLOperationRead) {
-		resp.ErrorCode = kerr.GroupAuthorizationFailed.Code
+	if e := c.deny(creq, groupID, kmsg.ACLResourceTypeGroup, kmsg.ACLOperationRead, faultKey{group: groupID}); e != nil {
+		resp.ErrorCode = e.Code
 		return resp, nil
 	}
 
@@ -286,13 +287,18 @@ func (c *Cluster) handleShareFetch(creq *clientReq, w *watchShareFetch) (kmsg.Re
 			}
 			continue
 		}
-		if !c.allowedACL(creq, topicName, kmsg.ACLResourceTypeTopic, kmsg.ACLOperationRead) {
+		tk := faultKey{topic: topicName, topicID: topicID}
+		if e := c.deny(creq, topicName, kmsg.ACLResourceTypeTopic, kmsg.ACLOperationRead, tk); e != nil {
 			for p := range parts {
-				donep(topicID, p, kerr.TopicAuthorizationFailed.Code)
+				donep(topicID, p, e.Code)
 			}
 			continue
 		}
 		for p := range parts {
+			if e := fc.check(tk.part(p)); e != nil {
+				donep(topicID, p, e.Code)
+				continue
+			}
 			pd, ok := c.data.tps.getp(topicName, p)
 			if !ok {
 				donep(topicID, p, kerr.UnknownTopicOrPartition.Code)
@@ -445,7 +451,7 @@ func (c *Cluster) handleShareFetch(creq *clientReq, w *watchShareFetch) (kmsg.Re
 	// fireShareWatchers fires when acks free the window, so the watcher
 	// wakes up promptly instead of the client busy-looping with empty
 	// fetches.
-	if totalRecords == 0 && w == nil {
+	if totalRecords == 0 && w == nil && !fc.anyHit() { // a faulted partition completes the fetch at once
 		wait := time.Duration(req.MaxWaitMillis) * time.Millisecond
 		deadline := creq.at.Add(wait)
 		remaining := time.Until(deadline)
--- a/pkg/kfake/faults.go
+++ b/pkg/kfake/faults.go
@@ -1,0 +1,401 @@
+package kfake
+
+import (
+	"context"
+	"reflect"
+	"slices"
+	"sync"
+	"sync/atomic"
+
+	"github.com/twmb/franz-go/pkg/kerr"
+	"github.com/twmb/franz-go/pkg/kmsg"
+)
+
+// We check faults wherever we check authorization, and per partition in
+// every request that carries partitions.
+
+// Fault fails matching requests with Err in place of the real answer, with
+// every field being an AND filter of what the fault should apply to. Faults
+// are an easier way to inject targeted failures than Control.
+//
+// A selector that a request does not carry never matches: a fault with a
+// Topic does not fail a heartbeat. Requests with a top-level ErrorCode can
+// have that error code set by setting TopLevel to true (some requests set
+// either a top-level code OR a per-resource code).
+//
+// A faulted entity is rejected before the broker acts on it, except with
+// REQUEST_TIMED_OUT, on the requests a real broker can answer that way after
+// applying them, or NOT_ENOUGH_REPLICAS_AFTER_APPEND on produce: those are
+// answered and the work still happens. Faults are checked in the order added
+// and the first match answers; a ControlKey control that answers a request
+// bypasses them.
+type Fault struct {
+	Keys  []kmsg.Key // which requests this should apply to; nil means all
+	Nodes []int32    // which nodes this should apply to; nil means all
+
+	Topic      string   // "" means all
+	TopicID    [16]byte // zero means all
+	Partitions []int32  // which partitions this should apply to; nil means all
+	Group      string   // "" means all; a FindCoordinator group key matches this
+	TxnID      string   // "" means all; a FindCoordinator transaction key matches this
+	TopLevel   bool     // fail the request's top-level ErrorCode rather than its entities, ignored for requests without one
+
+	// Resource names an entity no selector above reaches: a config
+	// resource, a client quota entity, a SCRAM user, a log dir, a feature,
+	// a member in a LeaveGroup, or the resource name of an ACL creation or
+	// filter. "" means all.
+	Resource string
+
+	Err   *kerr.Error // nil defaults to UNKNOWN_SERVER_ERROR
+	Count int         // requests to fault; 0 means one, -1 means until Remove is called
+}
+
+// FaultHandle refers to the faults installed by one Fault call.
+type FaultHandle struct {
+	c  *Cluster
+	fs []*fault
+}
+
+type fault struct {
+	keys       []kmsg.Key
+	nodes      []int32
+	topic      string
+	topicID    uuid
+	partitions []int32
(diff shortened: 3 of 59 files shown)


pkg/kfake/00_produce.go
```diff
@@ -45,14 +45,25 @@ func (c *Cluster) handleProduce(creq *clientReq) (kmsg.Response, error) {
 		return nil, err
 	}
 
+	answered := make(map[tp]bool)
```
twmb: Can't this be `map[tp]*kmsg.ProduceResponseTopicPartition`? deletes the loop below.

pkg/kfake/faults.go
```diff
@@ -0,0 +1,405 @@
...
+func (h *FaultHandle) Wait(ctx context.Context, n int) error {
+	for {
+		h.c.faultsMu.Lock()
+		hit := h.c.faultHit
+		h.c.faultsMu.Unlock()
+		if h.Hits() >= n {
+			return nil
+		}
+		select {
+		case <-hit:
```
twmb: A sync.Cond is more resilient here - a channel only notifies one waiter. I have a similar pattern handling this in kgo producer.go; we use a function-local goroutine.

pkg/kfake/faults.go
```diff
@@ -0,0 +1,405 @@
...
+	if len(fs) == 0 {
+		return nil
+	}
+	return &faultCheck{c: c, key: key, fs: fs}
+}
+
+// check returns the error to answer the entity named by k with, if a fault
+// matches it, and whether the caller must still do the entity's work. The
+// bool only matters where the site applies something: for a read, rejecting
+// and answering after the read are the same answer.
```
twmb: "The bool only matters" - sentence is useless.

pkg/kfake/faults.go
```diff
@@ -0,0 +1,405 @@
...
+}
+
+// faultCheck is the faults that can match one request. A request consumes at
+// most one unit from each fault it hits, no matter how many of its entities
+// match, so we remember what we charged.
+type faultCheck struct {
+	c       *Cluster
+	key     int16
+	fs      []*fault
+	charged []*fault
```
twmb: I don't like this charged terminology, it doesn't make sense. What's it doing? What's an alternative?

pkg/kfake/share_groups.go
```diff
@@ -1154,6 +1154,31 @@ func (g *shareGroup) processShareAcks(
 	onPartition func(tid uuid, p int32, ec int16),
 	onNotLeader func(tid uuid, p int32, pd *partData),
 ) (toFire []*partData) {
+	// A fault answered before the ack still lets the ack apply; the
+	// answer it gave stands, so we drop what the ack itself would
+	// answer with.
```
twmb: Confusing doc line

pkg/kfake/groups.go
```diff
@@ -437,9 +437,11 @@ func (gs *groups) handleDescribe(creq *clientReq) *kmsg.DescribeGroupsResponse {
 	for _, rg := range req.Groups {
 		sg := doneg(rg)
 		// ACL check: DESCRIBE on Group
-		if !gs.c.allowedACL(creq, rg, kmsg.ACLResourceTypeGroup, kmsg.ACLOperationDescribe) {
-			sg.ErrorCode = kerr.GroupAuthorizationFailed.Code
-			continue
+		if e, after := gs.c.deny(creq, rg, kmsg.ACLResourceTypeGroup, kmsg.ACLOperationDescribe, faultKey{group: rg}); e != nil {
+			sg.ErrorCode = e.Code
+			if !after {
```
twmb: I don't like reading "if not after" - it's confusing to always read in the negative. We need a better system for this.
I'm also confused why we're applying this before/after logic to requests that, within standard kafka, can't time out. We should inventory requests that actually CAN have RequestTimedOut, and only apply it to those.

pkg/kfake/groups.go
```diff
@@ -761,12 +777,21 @@ func (g *group) handleOffsetDelete(creq *clientReq) *kmsg.OffsetDeleteResponse {
 
 	for _, t := range req.Topics {
 		for _, p := range t.Partitions {
+			e, after := creq.faults.check(faultKey{group: req.Group, topic: t.Topic}.part(p.Partition))
```
twmb: Why the before AND after check here?

pkg/kfake/groups.go
```diff
@@ -1269,11 +1306,23 @@ func (g *group) handleLeave(creq *clientReq) kmsg.Response {
 
 		r := &resp.Members[len(resp.Members)-1]
 
+		// A fault answered before the leave still lets the member
+		// leave; the answer it gave stands.
+		e, after := creq.faults.check(faultKey{group: req.Group, resource: rm.MemberID})
```
twmb: I think every time an "after" (or whatever we rename this, due to my not wanting to read "if not after") CAN be handled, we should add a doc about it. Short, one sentence.

pkg/kfake/00_produce.go
twmb: Done as an index rather than a pointer: donep appends to the slice the pointer would name, so a stored pointer dangles after the next append. The scan is gone.

pkg/kfake/faults.go
twmb: Renamed to the API's own word: hit(f), hits, anyHit. What it does: a request counts against a fault's Count at most once however many of its entities match, and the request remembers which faults it already hit.

pkg/kfake/groups.go
twmb: Sites now read `if fc.rejects(e) { continue }`, no negations. After-apply is restricted to the kinds where a real broker can answer REQUEST_TIMED_OUT after applying (traced in the Kafka source): produce, WriteTxnMarkers, DeleteRecords, OffsetCommit and TxnOffsetCommit (classic coordinator up to 3.9), DeleteGroups, the share offset alter and delete, and the forwarded admin requests. Everything else, including the whole group protocol and the transaction coordinator requests, rejects. Note that coordinator write timeouts reach clients as COORDINATOR_NOT_AVAILABLE / NOT_COORDINATOR / REBALANCE_IN_PROGRESS with the record already written; not modeled.

pkg/kfake/groups.go
twmb: Gone: the site answers first with the fault's code, does the delete, and the emission dedupes, one check like produce.

pkg/kfake/faults.go
```diff
@@ -0,0 +1,400 @@
...
+	int16(kmsg.IncrementalAlterConfigs):   {kerr.RequestTimedOut},
+	int16(kmsg.CreateACLs):                {kerr.RequestTimedOut},
+	int16(kmsg.DeleteACLs):                {kerr.RequestTimedOut},
+	int16(kmsg.AlterClientQuotas):         {kerr.RequestTimedOut},
+	int16(kmsg.AlterUserSCRAMCredentials): {kerr.RequestTimedOut},
+	int16(kmsg.UpdateFeatures):            {kerr.RequestTimedOut},
+}
+
+// rejects reports whether e stops the entity's work. Every denial does, and so
+// does every fault but one afterApply allows for this request.
```
twmb: "but one afterApply allows for this request" -> confusing. Reword.

Should this function be named `blocksWork`?
Why do we start with a nil check?

pkg/kfake/faults.go
```diff
@@ -0,0 +1,400 @@
...
+}
+
+// check returns the error to answer the entity named by k with, if a fault
+// matches it.
+func (fc *faultCheck) check(k faultKey) *kerr.Error {
+	if fc == nil {
+		return nil
+	}
+	for _, f := range fc.fs {
+		if f.topLevel || !f.matches(k) {
```
twmb, line 251: why filter out toplevel here?

pkg/kfake/faults.go
```diff
@@ -0,0 +1,400 @@
...
+
+// rejects reports whether e stops the entity's work. Every denial does, and so
+// does every fault but one afterApply allows for this request.
+func (fc *faultCheck) rejects(e *kerr.Error) bool {
+	return fc == nil || !slices.Contains(afterApply[fc.key], e)
+}
+
+// fail answers the request with a matching fault's code set on its top-level
+// error code, or nil if no fault matches.
+func (fc *faultCheck) fail(kreq kmsg.Request, k faultKey) kmsg.Response {
```
twmb: All calls to `fail` use an empty faultKey. If it's not needed, it shouldn't be an argument here. Below in the check, what will an empty faultKey match, when used?

pkg/kfake/faults.go
```diff
@@ -0,0 +1,400 @@
...
+	resp := kreq.ResponseKind()
+	if code, ok := topLevelCode(resp); ok {
+		code.SetInt(int64(e.Code))
+	}
+	return resp
+}
+
+// topLevel answers the request itself when a fault sets a top-level error
+// code.
+func (fc *faultCheck) topLevel(kreq kmsg.Request) kmsg.Response {
```
twmb, line 309: How is this meaningfully different from `fail`? The doc is not clear enough as to why one or the other.

pkg/kfake/faults.go
```diff
@@ -0,0 +1,400 @@
...
+	if !code.IsValid() || code.Kind() != reflect.Int16 {
+		return reflect.Value{}, false
+	}
+	return code, true
+}
+
+// hit takes a unit of f's budget for this request, if this request has not
+// taken one already. A fault with nothing left removes itself.
+func (fc *faultCheck) hit(f *fault) bool {
+	if slices.Contains(fc.hits, f) {
```
twmb, line 350: In what case is a fault being checked twice?

pkg/kfake/faults.go
twmb: Now creq.blocksWork(e): reports whether answering e stops the entity's work, false only for the afterApply codes of this request kind. Off faultCheck, so no nil check.

pkg/kfake/faults.go
twmb: A TopLevel fault is answered only at dispatch by topLevel; if check matched it too, a TopLevel fault on a kind without a top-level code would fail its entities instead of being ignored. One-line comment added.

pkg/kfake/faults.go
twmb: Parameter gone, and fail itself is gone: merged into topLevel. An empty key matches exactly the faults with no entity selectors, since a set selector never matches a key that lacks it.

pkg/kfake/faults.go
twmb: Merged. topLevel now runs before every handler and answers a TopLevel fault on any request with a top-level code, and on the eight entityless kinds any fault with no entity selectors. The eight first-line hooks are deleted; the coverage test still passes for all 62 keys.

pkg/kfake/faults.go
twmb: A request with several matching entities: each entity's check finds the same fault, the first takes the unit, later ones are still answered. Doc says so now.

pkg/kfake/faults.go
twmb: Wait now uses a sync.Cond on faultsMu, the same shape as Client.Flush: a local goroutine waits on the cond while hits < n and closes done; the caller selects on done or ctx.Done(), and on ctx sets quit under the lock and broadcasts. hit broadcasts after incrementing. The per-hit channel is gone.

pkg/kfake/faults.go
twmb: Sentence removed; check's doc is now the one line saying what it returns.

pkg/kfake/share_groups.go
twmb: Reworded to one line: an ack answered by a fault keeps that answer; its own outcome is dropped.

pkg/kfake/groups.go
twmb: Every after-apply site now carries one sentence saying what still applies there (a timed-out commit is still stored, timed-out deletions still happen, and so on), and after-apply itself is limited to the kinds where a broker can answer REQUEST_TIMED_OUT after applying, per the Kafka source inventory.

pkg/kfake/78_share_fetch.go
```diff
@@ -445,7 +451,7 @@ func (c *Cluster) handleShareFetch(creq *clientReq, w *watchShareFetch) (kmsg.Re
 	// fireShareWatchers fires when acks free the window, so the watcher
 	// wakes up promptly instead of the client busy-looping with empty
 	// fetches.
-	if totalRecords == 0 && w == nil {
+	if totalRecords == 0 && w == nil && !fc.anyHit() {
```
twmb: Why the !fc.anyHit()? add doc.

twmb: It gates the long poll: a share fetch with no records parks in a watcher until data or MaxWait, and a partition a fault answered must complete at once instead, the same rule Fetch applies through returnEarly. Inline note added on that line.

PR #1408, opened 2026-09-05
kgo: rack aware assignment for topics we do not consume; evict cached metadata less

twmb:
Two things, the second noticed while writing the first.

**Rack aware assignment skipped the topics the leader does not consume.** `buildPartitionRacks` looked each topic up in `tps`, which only holds the topics this client consumes. A group can be interested in topics its leader does not, and those came back with no rack at all, so range, sticky, and cooperative sticky matched nothing for them: the leader assigned them without regard to rack, silently. Partition leaders now come from the metadata cache, which the metadata loop keeps for the group's whole topic set, with `tps` as the fallback for our own. A topic in neither still yields empty racks, so today's behavior is the floor.

Balancing also refetches when the cache cannot answer rack placement, but only when a member reported a rack and only for topics we do not consume ourselves. A group with no racks never pays, and neither does one whose topics are all cached.

**The metadata cache was evicting topics for requests that say nothing about topics.** `storeCachedMeta` evicts every cached topic older than `MetadataMinAge` that the response it stores does not contain. Loading brokers asks for *no* topics, so it dropped everything; resolving topic IDs, creating unknown produce topics, and loading a group's topics to balance each ask only about their own. We load brokers whenever we want a broker we do not know, whenever we want the controller and do not know it, and on every request that fans out to all brokers, so a client that also does admin work was throwing its topic cache away regularly.

Eviction now happens only for requests that say something about what we want cached: the metadata loop, a user's own metadata request, and `RequestCachedMetadata`. The loop still bounds the cache, since it asks only for the topics we produce to, consume, or balance.

No API change. #1379 stacks on this and gates rack aware balancing behind `BalanceRacks`.

### DIFF
--- a/pkg/kgo/client.go
+++ b/pkg/kgo/client.go
@@ -1067,7 +1067,7 @@ func (cl *Client) fetchBrokerMetadata(ctx context.Context) error {
 			}()
 			req := kmsg.NewPtrMetadataRequest()
 			req.Topics = []kmsg.MetadataRequestTopic{}
-			_, _, wait.err = cl.fetchMetadata(cl.ctx, req, true, nil)
+			_, _, wait.err = cl.fetchMetadata(cl.ctx, req, true, false, nil) // prune: no; brokers only, no opinion on topics
 		}()
 	}
 	cl.fetchingBrokersMu.Unlock()
@@ -1080,7 +1080,7 @@ func (cl *Client) fetchBrokerMetadata(ctx context.Context) error {
 	}
 }
 
-func (cl *Client) fetchMetadataByName(ctx context.Context, all bool, topics []string, results map[string]cachedMetaTopic) (*broker, *kmsg.MetadataResponse, error) {
+func (cl *Client) fetchMetadataByName(ctx context.Context, all bool, topics []string, prune bool, results map[string]cachedMetaTopic) (*broker, *kmsg.MetadataResponse, error) {
 	req := kmsg.NewPtrMetadataRequest()
 	req.AllowAutoTopicCreation = cl.cfg.allowAutoTopicCreation
 	if all {
@@ -1094,7 +1094,7 @@ func (cl *Client) fetchMetadataByName(ctx context.Context, all bool, topics []st
 			req.Topics = append(req.Topics, reqTopic)
 		}
 	}
-	return cl.fetchMetadata(ctx, req, true, results)
+	return cl.fetchMetadata(ctx, req, true, prune, results)
 }
 
 // resolveTopicMetaByID fetches metadata by TopicID and caches the results.
@@ -1121,11 +1121,11 @@ func (cl *Client) resolveTopicMetaByID(ctx context.Context, ids [][16]byte) (map
 		req.Topics = append(req.Topics, reqTopic)
 	}
 	results := make(map[string]cachedMetaTopic)
-	_, _, err := cl.fetchMetadata(ctx, req, true, results)
+	_, _, err := cl.fetchMetadata(ctx, req, true, false, results) // prune: no; only the ids we could not resolve
 	return results, err
 }
 
-func (cl *Client) fetchMetadata(ctx context.Context, req *kmsg.MetadataRequest, limitRetries bool, results map[string]cachedMetaTopic) (*broker, *kmsg.MetadataResponse, error) {
+func (cl *Client) fetchMetadata(ctx context.Context, req *kmsg.MetadataRequest, limitRetries, prune bool, results map[string]cachedMetaTopic) (*broker, *kmsg.MetadataResponse, error) {
 	r := cl.retryable()
 
 	var rebootstrapped bool
@@ -1160,7 +1160,7 @@ start:
 		cl.updateMetadataBrokers(meta)
 
 		// Cache the metadata, and potentially store each topic in the results.
-		cl.storeCachedMeta(meta, req.Topics == nil, results)
+		cl.storeCachedMeta(req, meta, prune, results)
 	}
 	return r.last, meta, err
 }
@@ -1563,7 +1563,7 @@ func (cl *Client) RequestCachedMetadata(ctx context.Context, req *kmsg.MetadataR
 			rt.TopicID = id
 			idReq.Topics = append(idReq.Topics, rt)
 		}
-		_, meta, err := cl.fetchMetadata(ctx, idReq, true, nil)
+		_, meta, err := cl.fetchMetadata(ctx, idReq, true, false, nil) // prune: no; only the ids we could not resolve
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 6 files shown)


pkg/kgo/group_balancer.go
```diff
@@ -430,8 +430,18 @@ func (g *groupConsumer) balanceGroup(proto string, members []kmsg.JoinGroupRespo
 		})
 	}
 
+	// KIP-881: we build rack info below from the metadata cache, which
+	// prunes a topic whenever it stores a response that does not cover it.
```
twmb: "that does not cover it" -> "that is missing the topic."

pkg/kgo/group_balancer.go
```diff
@@ -534,19 +544,47 @@ func (g *groupConsumer) balanceGroup(proto string, members []kmsg.JoinGroupRespo
...
 		}
 	}
-	if !hasRack {
+	return false
+}
+
+// needRackMeta returns whether we must refresh metadata before we can build
+// rack info for the topics we balance. We read partition leaders from the
+// metadata cache, which prunes a topic whenever it stores a response that
+// does not cover it, so the group's topics can be gone from it between
```
twmb: This entire comment can be worded:

"""
We always have racks for topics we consume. But, other group members can consume topics we aren't
consuming, and we need to load their racks. We use groupExternal to load their metadata, which populates
the meta cache, BUT the meta cache prunes topics on every metadata update - so, our own metadata update
will remove previously cached info about other group topics. This function returns if we need to populate racks
for other group member topics.
"""

A bit wordy, but clearer. Maybe can be tightened a little bit, but the existing wording is confusing.

pkg/kgo/group_balancer.go
```diff
@@ -556,13 +594,35 @@ func (g *groupConsumer) buildPartitionRacks(b *ConsumerBalancer, topicPartitionC
...
-	// partition's rack is determined by its leader broker's rack.
+	// Each partition's rack is its leader broker's rack. We prefer the
+	// metadata cache: it has every topic the group is interested in,
+	// including topics we do not consume and so never have in tps. The
+	// cache prunes entries whenever a request that does not cover them
+	// stores its response, so for our own topics we fall back to tps.
+	cached := make(map[string]cachedMetaTopic, len(topicPartitionCount))
+	g.cl.metaCache.mu.Lock()
+	for topic := range topicPartitionCount {
+		if ct, ok := g.cl.metaCache.topics[topic]; ok {
```
twmb, line 604: We should ensure the cache has a MINIMUM cache time so that we don't load external, then have an internal metadata update within 1s which prunes what external we just loaded. Does it?

pkg/kgo/group_balancer.go
twmb: Reworded, here and in the commit message. "a broker only metadata request covers nothing at all" became "is missing every topic" for the same reason.

pkg/kgo/group_balancer.go
twmb: Taken, tightened a little:

```go
// needRackMeta returns whether we need to load metadata before we can rack
// balance topics that other group members consume. We always have racks for
// the topics we consume ourselves; for the rest, we load metadata through
// groupExternal, which populates the metadata cache. The cache prunes a
// topic whenever it stores a response that is missing the topic, so our own
// metadata updates drop what we cached for other members' topics.
```

pkg/kgo/group_balancer.go
twmb: Yes, `MetadataMinAge` is that minimum. The prune only deletes entries older than it:

```go
if when.Sub(ct.when) > cl.cfg.metadataMinAge {
```

so anything just stored is protected for at least that long. It defaults to 5s and is validated to at least 10ms (`config.go:352`), with `MetadataMaxAge` validated to be no smaller. So the case you describe, loading external and then pruning it with an internal update a second later, cannot happen.

There is a second reason it does not happen: while `groupExternal` exists, its topics are in the metadata loop's own request set (`metadata.go:349`), so the loop covers them and refreshes their timestamps rather than pruning them.

What is left is a store that is missing those topics arriving more than `MetadataMinAge` after the last refresh: `fetchBrokerMetadata` sends an empty topic list, and ID resolution, `RequestCachedMetadata`, and produce auto-create each carry only their own topics. That is the case this function catches, and by then we are minutes past the load rather than one second.

pkg/kgo/group_balancer.go
```diff
@@ -430,8 +430,19 @@ func (g *groupConsumer) balanceGroup(proto string, members []kmsg.JoinGroupRespo
 		})
 	}
 
+	// KIP-881: we build rack info below from the metadata cache, which
+	// prunes a topic whenever it stores a response that is missing the
+	// topic. If the topics we cannot answer from tps are gone, fetch them
```
twmb: "if the topics we cannot answer" sentence - delete.

twmb: Deleted; the comment now stops at the pruning fact, and `needRackMeta`'s doc carries the rest.

pkg/kgo/client.go
```diff
@@ -3158,19 +3158,35 @@ func (cl *Client) resolveTopicMeta(ctx context.Context, topics []string, useCach
 	if results == nil {
 		results = make(map[string]cachedMetaTopic)
 	}
-	_, _, err := cl.fetchMetadataByName(ctx, all, needed, results)
+	_, _, err := cl.fetchMetadataByName(ctx, all, needed, true, results) // prune: yes; our caller wants these cached
```
twmb: `prune: yes; our caller wants only these cached`

pkg/kgo/group_balancer.go
```diff
@@ -430,14 +430,23 @@ func (g *groupConsumer) balanceGroup(proto string, members []kmsg.JoinGroupRespo
...
+			why = "our metadata cache no longer has all group topics and we balance by rack"
+		}
+		g.cl.cfg.logger.Log(LogLevelInfo, "fetching metadata for all group topics", "why", why)
 		var metaTopics []string
 		for topic := range topics {
 			metaTopics = append(metaTopics, topic)
 		}
 
-		_, resp, err := g.cl.fetchMetadataByName(g.ctx, false, metaTopics, nil)
+		_, resp, err := g.cl.fetchMetadataByName(g.ctx, false, metaTopics, false, nil) // prune: no; only the group's topics
```
twmb: What does "only the group's topics" mean here? Clarify, shortly.

pkg/kgo/metadata.go
```diff
@@ -369,7 +369,7 @@ func (cl *Client) updateMetadata() (retryWhy multiUpdateWhy, err error) {
 				unknownTopics = append(unknownTopics, unknown)
 			}
 			var err error
-			unknownCreateResp, err = cl.fetchTopicMetadata(false, unknownTopics)
+			unknownCreateResp, err = cl.fetchTopicMetadata(false, unknownTopics, false) // prune: no; only produce topics we do not know
```
twmb: reword comment beter.

pkg/kgo/client.go
twmb: Applied verbatim.

pkg/kgo/group_balancer.go
twmb: Reworded: `prune: no; the group's topics are not all we cache` - pruning here would drop the topics we produce to, which this fetch says nothing about.

pkg/kgo/metadata.go
twmb: Reworded: `prune: no; unknown produce topics only, the fetch below covers the rest` - it is a side fetch, and the loop's real request follows a few lines down.

PR #1409, opened 2026-09-07
kgo: only log regex-consumed topics as added if not excluded

lahsivjar:
filterMetadataAllTopics logged a topic as "added" as soon as an include
regex matched, before the exclude regexes were evaluated. An excluded
topic therefore appeared in both the "added" and "evaluated_and_skipped"
fields of the "consumer regular expressions evaluated on new topics"
log line, implying it would be consumed even though it never is.

Log the topic as added only after it also passes the exclude regexes.


twmb:
good find, thanks, this will ride along with the next release whenever that is.

PR #1410, opened 2026-09-07
kgo: never regex-match internal topics; align 848 regex with the broker

twmb:
Follow-up to #1409. Reviewing that change surfaced three pre-existing divergences in regex consuming, all in or around `filterMetadataAllTopics`.

## Internal topics

Classic and direct regex consumers have skipped internal topics since 2019, but the skip lived at the assignment points, not in the filter. So:

- a share consumer in regex mode sent internal topics by name in `SubscribedTopicNames` and consumed them;
- a classic consumer logged a matching internal topic as `added` and returned it from `GetConsumeTopics` while never assigning it.

The filter now skips internal topics itself, so every path that subscribes from `tps` agrees, and the three per-path checks are gone.

## KIP-848 regex

The broker resolves `SubscribedTopicRegex` against the entire topic name (re2j `matches()`), while kgo matches anywhere in the name (`MatchString`) on every other path. `ConsumeTopics("orders")` therefore consumed `orders` alone on 848 but `orders`, `orders-v2`, and `reorders` everywhere else, including 848 the moment any `ConsumeExcludeTopics` was set. We now wrap the regex sent to the broker in `.*` so it resolves the same names we would. Anchoring with `^` and `$` still gives an exact match.

The broker also does not skip internal topics when resolving a regex (only a DESCRIBE ACL filter applies). With the filter change above, a broker-assigned internal topic would be missing from `tps`. The heartbeat now adopts such a topic into `tps` and parks the assignment until metadata has loaded it, reusing the unresolved-ID mechanism. The same wait covers an assigned topic whose ID resolved via `id2t` before its partitions were stored.

Verified against Kafka 4.2.0 with an 848 consumer on `.*`:

```
[INFO] adopting a topic the broker assigned that our regex filter skipped; group: verify848-regex, topic: __consumer_offsets
[INFO] consumer group heartbeat detected an updated assignment; ... now_assigned: map[__consumer_offsets:[0 1 2 3 4] verify848-topic:[0]]
FETCHED topic=__consumer_offsets records=11
```

## Log

`consumer regular expressions evaluated on new topics` now carries `excluded` (keyed by the exclude regex that matched) and `skipped_internal` alongside `added` and `evaluated_and_skipped`, so an excluded topic is distinguishable from one that matched nothing. Include regexes are evaluated in sorted order so attribution is stable. The log runs after the consumer lock is released and is skipped below Info.

## kfake

kfake now anchors the 848 regex the way Kafka does, so the kgo tests exercise the wrap.

## Tests

- `TestConsumeRegex` asserts the log buckets and that an internal topic is neither added nor returned.
- `TestConsumeRegexShareSkipsInternal` covers the share path.
- `Test848RegexBrokerResolves` covers a bare-name include matching a longer topic name and adoption of a broker-assigned internal topic.

The internal-topic helper creates one on kfake through `kfake.is_internal` and, on a real broker, nudges a coordinator lookup so `__consumer_offsets` exists. All three pass 30x under `-race` against Kafka 4.2.0 and run unchanged on kfake.

### DIFF
--- a/pkg/kgo/config.go
+++ b/pkg/kgo/config.go
@@ -1725,6 +1725,16 @@ func ConsumePartitions(partitions map[string]map[int32]Offset) ConsumerOpt {
 // regular expressions. You can further use ConsumeExcludeTopics to exclude
 // topics that would match any ConsumeTopics regex.
 //
+// A regular expression matches anywhere in a topic name, the same as Go's
+// MatchString: "foo" matches "foo", "foobar", and "barfoo". Anchor with ^ and
+// $ to match an entire name. A regex client never consumes internal topics
+// such as __consumer_offsets; consume those from a client without
+// ConsumeRegex. The one exception to this is the next-gen consumer group
+// protocol, where if you don't use ConsumeExcludeTopics, the broker resolves
+// the regex and may include internal topics. ConsumeExcludeTopics forces
+// client-side regex evaluation because the next-gen protocol does not yet
+// support exclude regexes.
+//
 // When consuming via regex, every metadata request loads *all* topics, so that
 // all topics can be passed to any regular expressions. Every topic is
 // evaluated only once ever across all regular expressions; either it
--- a/pkg/kgo/consumer.go
+++ b/pkg/kgo/consumer.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 	"math"
 	"slices"
 	"sort"
@@ -1355,13 +1356,18 @@ func (c *consumer) assignPartitions(assignments map[string]map[int32]Offset, how
 }
 
 // filterMetadataAllTopics, called BEFORE doOnMetadataUpdate, evaluates
-// all topics received against the user provided regex.
-func (c *consumer) filterMetadataAllTopics(topics []string) []string {
-	c.mu.Lock()
-	defer c.mu.Unlock()
-
+// all topics received against the user provided regex and returns the
+// topics we will consume.
+//
+// We never match internal topics. The broker can still assign one to us
+// when it resolves the regex itself (KIP-848 with no excludes); see
+// adoptAssignedTopic.
+func (c *consumer) filterMetadataAllTopics(latest map[string]*metadataTopic) []string {
 	var rns reNews
 	defer rns.log(&c.cl.cfg)
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
 
 	var reSeen map[string]bool
 	if c.g != nil {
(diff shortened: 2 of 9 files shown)


pkg/kgo/consumer.go
```diff
@@ -1372,32 +1380,47 @@ func (c *consumer) filterMetadataAllTopics(topics []string) []string {
 		reSeen = c.d.reSeen
 	}
 
-	keep := topics[:0]
-	for _, topic := range topics {
+	// We evaluate regexes in sorted order so that a topic matching
+	// multiple regexes is always logged under the same one.
```
twmb: 'always logged under the same one' what?

pkg/kgo/config.go
```diff
@@ -1725,6 +1725,14 @@ func ConsumePartitions(partitions map[string]map[int32]Offset) ConsumerOpt {
 // regular expressions. You can further use ConsumeExcludeTopics to exclude
 // topics that would match any ConsumeTopics regex.
 //
+// A regular expression matches anywhere in a topic name, the same as Go's
+// MatchString: "foo" matches "foo", "foobar", and "barfoo". Anchor with ^ and
+// $ to match an entire name. We never match internal topics such as
+// __consumer_offsets; to consume one, name it in ConsumeTopics without
+// ConsumeRegex. The exception is the next-gen consumer group protocol with no
```
twmb: The one exception to this is the next-gen consumer group protocol, where
if you don't use ConsumeExcludeTopics, the broker resolves the regex and
may include internal topics. ConsumeExcludeTopics forces client-side regex
evaluation because the "next-gen" protocol does not yet support exclude regexes.

pkg/kgo/consumer.go
```diff
@@ -1355,14 +1356,21 @@ func (c *consumer) assignPartitions(assignments map[string]map[int32]Offset, how
...
-
+// all topics received against the user provided regex and returns the
+// topics we will consume.
+//
+// We never match internal topics. The broker can still assign one to us
+// when it resolves the regex itself (KIP-848 with no excludes); see
+// adoptAssignedTopic.
+func (c *consumer) filterMetadataAllTopics(latest map[string]*metadataTopic) []string {
+	// We log after unlocking so that the user's logger never runs under
+	// our lock.
```
twmb: Unnecessary doc comment, obvious

pkg/kgo/consumer_group.go
```diff
@@ -2412,9 +2412,6 @@ func (g *groupConsumer) findNewAssignments() {
 		// want to load the metadata", but the topic was not returned
 		// in the metadata (or it was returned with an error).
 		if useTopic && numPartitions > 0 {
-			if g.cfg.regex && parts.isInternal {
-				continue
-			}
```
twmb, line 2417: Why was this conditional deleted?

pkg/kgo/consumer_group_848.go
```diff
@@ -18,6 +18,18 @@ import (
 	"github.com/twmb/franz-go/pkg/kmsg"
 )
 
+// adoptAssignedTopic adds a topic the broker assigned us that our regex
+// filter skipped, so that the next metadata update loads it and we can
+// fetch from it.
+func (g *groupConsumer) adoptAssignedTopic(topic string) {
+	g.c.mu.Lock()
+	defer g.c.mu.Unlock()
+	g.cfg.logger.Log(LogLevelInfo, "adopting a topic the broker assigned that our regex filter skipped", "group", g.cfg.group, "topic", topic)
```
twmb: log can come outside lock

pkg/kgo/metadata.go
```diff
@@ -450,24 +450,12 @@ func (cl *Client) updateMetadata() (retryWhy multiUpdateWhy, err error) {
...
-		for topic, mt := range latest {
-			// loadErr should only be non-nil when requesting all
-			// topics if this is with auto-topic-creation && the
-			// creation failed. That is, we should not consume the
-			// topic since we just tried creating it and creating
-			// it failed.
-			if mt.loadErr == nil {
-				allTopics = append(allTopics, topic)
-			}
-		}
```
twmb, line 463: Why was this deleted?

pkg/kgo/consumer.go
twmb: Reworded in a8258bd0: sorted so the log is deterministic when a topic matches multiple regexes.

pkg/kgo/config.go
twmb: Took your wording in a8258bd0, and the lead-in now reads: a regex client never consumes internal topics such as __consumer_offsets; consume those from a client without ConsumeRegex.

pkg/kgo/consumer.go
twmb: Removed in a8258bd0.

pkg/kgo/consumer_group.go
twmb: Unreachable now: the filter marks internal topics reSeen=false, so useTopic is already false for them in regex mode. The one way an internal topic reaches this loop is 848 adoption, which sets reSeen=true on purpose so the topic lands in using.

pkg/kgo/consumer_group_848.go
twmb: Moved above the lock in a8258bd0.

pkg/kgo/metadata.go
twmb: Moved, not dropped: the filter needs isInternal, so it takes the metadata map now and does the loadErr skip itself, with the same comment.

PR #1414, opened 2026-09-08
kgo: respect produce retry backoff for repeated leader hints

### LINKED ISSUE #1412: Repeated leader hints bypass Produce retry backoff
## Background

We found this issue while investigating Produce failures during Kafka rolling restarts / upgrades: when `NOT_LEADER_OR_FOLLOWER` responses repeatedly report the leader ID and epoch already known to the client, franz-go bypasses `RetryBackoffFn` and rapidly exhausts the retry budget. Other scenarios producing the same responses may also be affected.

This path requires Produce API v10 or later, which introduced the KIP-951 leader hints. The original Kafka cluster was running 4.1.2, which supports Produce v3–v13; we did not capture the exact negotiated Produce version during the incident.

- We did not capture the leader and epoch in each response during the original incident, so its exact trigger remains unconfirmed. The independent kfake reproduction below demonstrates the repeated-hint behavior.

## Expected and actual behavior

According to [KIP-951](https://cwiki.apache.org/confluence/spaces/KAFKA/pages/263424791/KIP-951+Leader+discovery+optimizations+for+the+client), a higher leader epoch permits an immediate retry; subsequent failures at the same epoch should respect the client's backoff policy.

With `RecordRetries(5)` and a fixed 200ms `RetryBackoffFn`, all six Produce attempts fail rapidly, without calling the backoff function:

| franz-go version | Total elapsed time | Produce requests | Backoff calls |
| --- | --- | --- | --- |
| `v1.21.6` | 58.06ms | 6 | 0 |
| `6b0b6509f117` (master, 2026-09-08) | 57.25ms | 6 | 0 |

Both runs return `record failed after being retried too many times`. The tested master commit includes [PR #1407](https://github.com/twmb/franz-go/pull/1407); this issue remains reproducible.

## Minimal reproduction

Environment: Go `1.25.14 darwin/arm64`, kmsg `v1.13.1`; other dependency versions are pinned below. No external Kafka cluster is required.

Save the program as `main.go` in an empty directory and run:

```sh
go mod init retry-backoff-repro
go get github.com/twmb/franz-go@v1.21.6 github.com/twmb/franz-go/pkg/kfake@v0.0.0-20260727183601-4176fc0fcaf7
go run .
# Check the tested master commit:
go get github.com/twmb/franz-go@6b0b6509f117452fe2812c68a5489f5429246b1a
go run .
```

After a successful warmup, every Produce response is injected with `NOT_LEADER_OR_FOLLOWER`, repeating the already known leader 0 and epoch 0. Counters start after warmup.

```go
package main

import (
	"context"
	"fmt"
	"net"
	"strconv"
	"sync/atomic"
	"time"

	"github.com/twmb/franz-go/pkg/kerr"
)

func main() {
	c := kfake.MustCluster(kfake.NumBrokers(1), kfake.SeedTopics(1, "retry-probe"))
	defer c.Close()
	var backoffCalls atomic.Int32
	cl, err := kgo.NewClient(
		kgo.RetryBackoffFn(func(int) time.Duration { backoffCalls.Add(1); return 200 * time.Millisecond }),
		kgo.SeedBrokers(c.ListenAddrs()...),
		kgo.DisableIdempotentWrite(),
		kgo.MaxProduceRequestsInflightPerBroker(1),
		kgo.RecordRetries(5),
	)
	if err != nil {
		panic(err)
	}
	defer cl.Close()
	ctx, cancel := context.WithTimeout(context.Background(), 40*time.Second)
	defer cancel()
	if err := cl.ProduceSync(ctx, &kgo.Record{Topic: "retry-probe", Value: []byte("warmup")}).FirstErr(); err != nil {
		panic(err)
	}
	var metadata, produces atomic.Int32
	c.ControlKey(int16(kmsg.Metadata), func(req kmsg.Request) (kmsg.Response, error, bool) {
		c.KeepControl()
		metadata.Add(1)
		return nil, nil, false
	})
	c.ControlKey(int16(kmsg.Produce), func(req kmsg.Request) (kmsg.Response, error, bool) {
		c.KeepControl()
		produces.Add(1)
		r := req.ResponseKind().(*kmsg.ProduceResponse)
		t := kmsg.NewProduceResponseTopic()
		t.Topic = "retry-probe"
		t.TopicID = req.(*kmsg.ProduceRequest).Topics[0].TopicID
		p := kmsg.NewProduceResponseTopicPartition()
		p.Partition = 0
		p.ErrorCode = kerr.NotLeaderForPartition.Code
		p.CurrentLeader.LeaderID = 0
		p.CurrentLeader.LeaderEpoch = 0
		host, portText, _ := net.SplitHostPort(c.ListenAddrs()[0])
		port, _ := strconv.Atoi(portText)
		r.Brokers = []kmsg.ProduceResponseBroker{{NodeID: 0, Host: host, Port: int32(port)}}
		t.Partitions = append(t.Partitions, p)
		r.Topics = append(r.Topics, t)
		return r, nil, true
	})
	start := time.Now()
	err = cl.ProduceSync(ctx, &kgo.Record{Topic: "retry-probe", Value: []byte("probe")}).FirstErr()
	fmt.Printf("elapsed=%s produce_requests=%d metadata_requests=%d backoff_calls=%d error=%v\n", time.Since(start), produces.Load(), metadata.Load(), backoffCalls.Load(), err)
}
```

## Code path

In `v1.21.6`, [`maybeAddProducePartition`](https://github.com/twmb/franz-go/blob/v1.21.6/pkg/kgo/topics_and_partitions.go#L758) accepts hints that do not advance the epoch → [`handleRetryBatches`](https://github.com/twmb/franz-go/blob/v1.21.6/pkg/kgo/sink.go#L1321) skips the normal retry wait for batches in `kmove` → [`doMove`](https://github.com/twmb/franz-go/blob/v1.21.6/pkg/kgo/topics_and_partitions.go#L919) calls `clearFailing()` even when the route is unchanged, scheduling another send.

jjj-n:
Repeated `NOT_LEADER_OR_FOLLOWER` responses with an unchanged leader epoch currently take the KIP-951 immediate retry path and can exhaust `RecordRetries` without calling `RetryBackoffFn`.

Related: #1413 was opened while this patch was being prepared. This is an independently prepared alternative for #1412, and the core fix overlaps. The additional kfake cases here may also be useful for that PR, particularly idempotent producers, older epochs, and an advancing epoch followed by repeated hints.

Only use that path when the hinted epoch advances the record buffer's known epoch. Route repeated or stale hints through the existing produce backoff path, refresh metadata asynchronously, and retain retry-limit handling. A higher epoch, including a move to another broker, still retries immediately.

Also check sink backoff after acquiring an inflight slot. A failure response can request backoff while the drain loop is waiting for that slot; checking before the wait lets the next send miss it.

Add kfake regressions for equal and stale epochs, an advancing epoch followed by repeated hints, ordinary request timeouts, and a real leader move. Cover both idempotent and non-idempotent producers where applicable, and disable client telemetry in these tests so its requests do not affect the shared backoff counter.

Validation:

- Confirmed the equal-epoch regression fails on unmodified master: six Produce attempts in about 59 ms, zero backoff calls, with five retries and a 50 ms backoff.
- `go test -race -run '^TestIssue1412' -count=20 -timeout=2m .` in `pkg/kfake`.
- `go test -timeout=5m ./...` and `go test -race -timeout=5m ./...` in `pkg/kfake`.
- `go test -timeout=5m ./...` in the root module against kfake configured as Kafka 4.1.
- `golangci-lint run --timeout=5m ./pkg/kgo/...` in the root module and `golangci-lint run --timeout=5m ./...` in `pkg/kfake`.

Fixes #1412.

### DIFF
--- a/pkg/kfake/issue_1412_test.go
+++ b/pkg/kfake/issue_1412_test.go
@@ -1,0 +1,162 @@
+package kfake
+
+import (
+	"context"
+	"errors"
+	"net"
+	"strconv"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/twmb/franz-go/pkg/kerr"
+	"github.com/twmb/franz-go/pkg/kgo"
+	"github.com/twmb/franz-go/pkg/kmsg"
+)
+
+// Repeated KIP-951 hints for the current leader must not bypass the produce
+// retry backoff and exhaust the retry budget in a tight loop.
+func TestIssue1412ProduceRetryBackoff(t *testing.T) {
+	t.Parallel()
+	for _, idempotent := range []bool{false, true} {
+		for _, test := range []struct {
+			name       string
+			epochDelta int32
+			timeout    bool
+			immediate  int32
+		}{
+			{name: "same_epoch"},
+			{name: "stale_epoch", epochDelta: -1},
+			{name: "new_epoch_then_repeated", epochDelta: 1, immediate: 1},
+			{name: "request_timeout", timeout: true},
+		} {
+			// An idempotent producer intentionally keeps retrying a timed-out
+			// append whose outcome is unknown, even beyond RecordRetries.
+			if test.timeout && idempotent {
+				continue
+			}
+			t.Run(test.name+"/idempotent="+strconv.FormatBool(idempotent), func(t *testing.T) {
+				t.Parallel()
+				const topic = "retry-backoff"
+				const retries = 5
+				const backoff = 50 * time.Millisecond
+				c := newCluster(t, NumBrokers(1), SeedTopics(1, topic))
+				// Start at epoch 1 so the stale hint still has a valid epoch.
+				if err := c.MoveTopicPartition(topic, 0, 0); err != nil {
+					t.Fatal(err)
+				}
+				var backoffs atomic.Int32
+				var produces atomic.Int32
+				opts := []kgo.Opt{
+					// Telemetry requests share RetryBackoffFn; count only produce retries.
+					kgo.DisableClientMetrics(),
+					kgo.RecordRetries(retries),
+					kgo.RetryBackoffFn(func(int) time.Duration {
+						backoffs.Add(1)
+						return backoff
+					}),
+				}
+				if !idempotent {
+					opts = append(opts, kgo.DisableIdempotentWrite(), kgo.MaxProduceRequestsInflightPerBroker(1))
+				}
+				cl := newPlainClient(t, c, opts...)
+				ctx, cancel := context.WithTimeout(t.Context(), 10*time.Second)
+				defer cancel()
+				if err := cl.ProduceSync(ctx, &kgo.Record{Topic: topic, Value: []byte("warmup")}).FirstErr(); err != nil {
+					t.Fatal(err)
+				}
+				backoffs.Store(0)
+				pi := c.PartitionInfo(topic, 0)
+				host, portText, err := net.SplitHostPort(c.ListenAddrs()[0])
+				if err != nil {
+					t.Fatal(err)
+				}
+				port, err := strconv.Atoi(portText)
+				if err != nil {
+					t.Fatal(err)
+				}
+				c.ControlKey(int16(kmsg.Produce), func(req kmsg.Request) (kmsg.Response, error, bool) {
+					c.KeepControl()
+					n := produces.Add(1)
+					if n > retries+1 {
+						// Allow recovery so a retry-limit regression cannot hang.
+						return nil, nil, false
+					}
+					if want := n - 1 - test.immediate; backoffs.Load() < want {
+						t.Errorf("produce %d: backoff calls = %d, want at least %d", n, backoffs.Load(), want)
+					}
+					if n == 2 && test.immediate != 0 && backoffs.Load() != 0 {
+						t.Error("advancing leader epoch did not retry immediately")
+					}
+					resp := req.ResponseKind().(*kmsg.ProduceResponse)
+					rt := kmsg.NewProduceResponseTopic()
+					rt.Topic = topic
+					rt.TopicID = req.(*kmsg.ProduceRequest).Topics[0].TopicID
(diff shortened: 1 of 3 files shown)


twmb:
Thanks for the PR - I workshopped a few extra improvements into my own PR. You caught largely the same things, with some small implementation differences, and added more tests. I've copied your tests (and added you as co-authored-by) into my PR as a separate commit. The differences between your implementation and mine are,

Yours:
* The `<=` comparison: a hint has to carry a newer epoch to count as a move. I'd started from `doMove`'s existing check, where a same-epoch hint at a different leader still moves, but two brokers naming each other at one epoch would move us back and forth with no wait between attempts, so I switched to yours and made `doMove` match. Same rule as the Java client, which is some reassurance.
* A table of tests over idempotent and non-idempotent producers, each with the same epoch, an older epoch, and a hint that moves once and then repeats at the new epoch; plus a real leader move test pinning that a moving hint retries at once.
* Bumping the partition to epoch 1 first so an older hint is still a valid epoch, which is what lets the backwards-hint case run in kfake at all.
* `DisableClientMetrics` in the tests so produce is the only request that can wait on `RetryBackoffFn`.

Mine:
* Also fixes the literal `0` backoff seq in the retry and missing-reply paths, which could back off once per sink lifetime and then never again.
* Also applies the guard on the fetch side, where a stale hint was silently swallowed with no metadata refresh.
* On exhaustion it surfaces the broker's NOT_LEADER_OR_FOLLOWER, the same as a hint-less NOT_LEADER, instead of ErrRecordRetries.
* Refreshes metadata opportunistically from the backoff rather than forcing one per stale response; both are held to `MetadataMinAge`, so no behavior difference.

Good spot, good work. As mentioned I merged your extra stuff into my PR, I'll close this and merge the first.

twmb:
Folded yours in (With co-authored-by in the commit), merged mine, closing this one ; it'll be out in the next patch

PR #1416, opened 2026-09-10
kgo: never skip records when resetting after OFFSET_OUT_OF_RANGE

twmb:
An OFFSET_OUT_OF_RANGE while consuming can skip records that are still in the log. It comes from resetting by timestamp alone.

### What the reset does today

`source.go` resets to `AfterMilli(lastConsumedTime)`. The broker answers with the earliest offset at or after `logStartOffset` whose record timestamp is at or after ours (`UnifiedLog.searchOffsetInLocalLog` passes `logStartOffset` as the starting offset; `FileRecords.searchForTimestamp` scans forward for the first record at or after it), or -1 if there is none. On -1 we substitute the end listing.

Timestamps are producer-set CreateTime, so offset order does not imply timestamp order.

### The rule

With N the offset that was out of range, start and end the log bounds, and byTime the by-time answer:

```
x = byTime if byTime >= 0 else N
resume = clamp(min(x, N), start, end)
```

| N vs the log | by-time | today | now |
|---|---|---|---|
| below the start | any | first survivor as new as ours, **skipping the rest**; -1 → **the end, skipping the backlog** | the start |
| in range (the log lost data and regrew under us) | at or below N | by-time | by-time: one duplicate on an intact log, the replacement records after a loss |
| in range | above N | by-time, **skipping N through byTime-1** | N |
| in range | -1 | **the end, skipping N through end** | N |
| past the end | a record | by-time | by-time |
| past the end | -1 | the end | the end |

Below the start there is nothing to guess at: everything that still exists is something we never consumed. Past the end the offsets diverged and by-time is the best guess at where we were. In range after an OOOR means the end moved up past us, which is loss then regrowth, so by-time re-reads the replacements rather than skipping them. And by-time is never allowed ahead of N, because everything from N on is unread.

### The change

One round of three listings: start, end, by-time. `buildListReq` returns a third request when a load carries `ooorMilli`, `listOffsetsForBrokerLoad` aligns the three responses pairwise, and the first resolution arm applies the rule. Three listings in one round everywhere, against two today.

### Second commit

The bounding arm always took the leader epoch from the first response even when it took the offset from the second. A cursor at the end offset with the start's epoch gets validated on the next leader change, OffsetForLeaderEpoch answers the end of that old epoch, and we raise `ErrDataLoss` and rewind to it. The epoch now comes from whichever listing supplied the offset; an offset we kept rather than listed gets -1.

### Third commit

An OffsetForLeaderEpoch validation that answers UNDEFINED_EPOCH_OFFSET used to pin the cursor at -1 so the next fetch would fail and reset. That fetch existed only to fail, it sent a negative offset, and it discarded the offset we were at, which the rule above needs. The validation now issues the reset load directly, and epoch loads carry the cursor's last consumed timestamp for it.

### Tests

`TestOutOfRangeBelowStartResetsToStart`: five records stamped an hour ahead, five normal, consume five, delete through 7, require 7/8/9. Before the fix it consumes none of the three.

`TestOutOfRangePastEndResetsByTime`: one fetch past the end answered OOOR; require resuming at the last consumed record.

`TestOutOfRangeInRangeNeverSkipsForward`: N in range with the end above it; a by-time answer above N and a by-time answer of -1 both resume at N.

Note that kfake fills in `LogStartOffset` on an OOOR fetch response, which Kafka does not: `handleOffsetOutOfRangeError` falls to the error-only `LogReadResult` for non-tiered topics, so the field is -1. Nothing here relies on it.

### DIFF
--- a/pkg/kgo/config.go
+++ b/pkg/kgo/config.go
@@ -1616,13 +1616,13 @@ func ConsumeStartOffset(offset Offset) ConsumerOpt {
 // earliest offset. If using this option, it is strongly recommended to also
 // set ConsumeStartOffset.
 //
-// This option is *only* used if a consumer seeds OffsetOutOfRange on the
-// *first* fetch of a partition. If the consumer has consumed the partition at
-// all and sees the error, it will automatically reset to the first offset
-// after the timestamp of the last successfully consumed offset. If data loss
-// occurred such that even the last successfully consumed offset is lost, the
-// client automatically resets to the new current end offset. If you want to
-// disable offset resetting entirely, you can use [NoResetOffset].
+// This option is *only* used if a consumer sees OffsetOutOfRange before it
+// has consumed anything from a partition. Once a partition has been consumed,
+// OffsetOutOfRange resets to the nearest offset that still exists: the log
+// start if the consumer fell below it, otherwise the first offset at or after
+// the last consumed record's timestamp, never ahead of where the consumer
+// was, and never past the log end. If you want to disable offset resetting
+// entirely, you can use [NoResetOffset].
 //
 // If you use an exact or relative offsets and the offset ends up out of range,
 // the client chooses the nearest of either the log start offset or the log end
--- a/pkg/kgo/consumer.go
+++ b/pkg/kgo/consumer.go
@@ -2490,7 +2490,11 @@ func (cl *Client) listOffsetsForBrokerLoad(ctx context.Context, broker *broker, 
 			}
 
 			offset := poffset(&rPartition)
-			end := func() int64 { return poffset(&resp2.Topics[i].Partitions[j]) }
+			epoch := rPartition.LeaderEpoch
+			end := func() (int64, int32) {
+				p := &resp2.Topics[i].Partitions[j]
+				return poffset(p), p.LeaderEpoch
+			}
 
 			// We ensured the resp2 shape is as we want and has no
 			// error, so resp2 lookups are safe.
@@ -2500,35 +2504,35 @@ func (cl *Client) listOffsetsForBrokerLoad(ctx context.Context, broker *broker, 
 				// our end offset request: anything after the
 				// end offset *now* is after our milli.
 				if offset == -1 {
-					offset = end()
+					offset, epoch = end()
 				}
 			} else if loadPart.at >= 0 {
 				// If an exact offset, we listed start and end.
 				// We validate the offset is within bounds.
-				end := end()
+				end, endEpoch := end()
 				want := loadPart.at + loadPart.relative
 				if want >= offset {
-					offset = want
+					offset, epoch = want, -1
 				}
 				if want >= end {
-					offset = end
+					offset, epoch = end, endEpoch
 				}
 			} else if loadPart.at == -2 && loadPart.relative > 0 {
 				// Relative to the start: both start & end were
 				// issued, and we bound to the end.
-				offset += loadPart.relative
-				if end := end(); offset >= end {
-					offset = end
+				offset, epoch = offset+loadPart.relative, -1
+				if end, endEpoch := end(); offset >= end {
+					offset, epoch = end, endEpoch
 				}
 			} else if loadPart.at == -1 && loadPart.relative < 0 {
 				// Relative to the end: both start & end were
 				// issued, offset is currently the start, so we
 				// set to the end and then bound to the start.
-				start := offset
-				offset = end()
-				offset += loadPart.relative
+				start, startEpoch := offset, epoch
+				end, _ := end()
+				offset, epoch = end+loadPart.relative, -1
 				if offset <= start {
-					offset = start
+					offset, epoch = start, startEpoch
 				}
 			}
 			// Every arm above yields a non-negative offset from a
@@ -2555,7 +2559,7 @@ func (cl *Client) listOffsetsForBrokerLoad(ctx context.Context, broker *broker, 
 				partition:   partition,
 				cursor:      topicPartition.cursor,
 				offset:      offset,
-				leaderEpoch: rPartition.LeaderEpoch,
+				leaderEpoch: epoch,
 				request:     loadPart,
 			})
 		}
--- a/pkg/kgo/consumer.go
+++ b/pkg/kgo/consumer.go
@@ -1492,6 +1492,11 @@ type offsetLoadMap map[string]map[int32]offsetLoad
 // to directly use if a cursor had a preferred replica.
 type offsetLoad struct {
 	replica int32 // -1 means leader
+	// ooorMilli is non-zero when we are resetting a cursor that received
+	// OFFSET_OUT_OF_RANGE while consuming: the timestamp of the last
+	// record it consumed. The offset itself is the one that was out of
+	// range. See listOffsetsForBrokerLoad.
+	ooorMilli int64
 	Offset
 }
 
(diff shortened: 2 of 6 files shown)


pkg/kgo/consumer.go
```diff
@@ -1492,6 +1492,13 @@ type offsetLoadMap map[string]map[int32]offsetLoad
 // to directly use if a cursor had a preferred replica.
 type offsetLoad struct {
 	replica int32 // -1 means leader
+	// If we are resetting a cursor that received OFFSET_OUT_OF_RANGE
+	// while consuming, ooor is true and ooorMilli is the timestamp of
+	// the last record it consumed. The offset itself is the one that
+	// was out of range: listing bounds it within the log, and if it is
+	// at or past the end we follow up by listing after ooorMilli.
+	ooor      bool
+	ooorMilli int64
```
twmb, line 1499: Really we only need ooorMilli, if non-zero then it's from ooor.

pkg/kgo/source.go
```diff
@@ -1385,10 +1385,33 @@ func (s *source) handleReqResp(br *broker, req *fetchRequest, resp *kmsg.FetchRe
...
+						// policy. We first bound our own offset
+						// within the log: below the start, the
+						// broker deleted what we were consuming
+						// and the start is the only offset that
+						// skips nothing. At or past the end, the
+						// broker lost data, and we follow up by
+						// listing after the last consumed
+						// timestamp. A cursor pinned at -1 by an
+						// epoch validation has nothing to bound,
+						// so it lists by time directly.
```
twmb: Let's make this a bit wider than 80 characters.... like 120 characters for comment blocks that _begin_ deeply nested is fine.

pkg/kgo/consumer.go
```diff
@@ -2489,8 +2506,21 @@ func (cl *Client) listOffsetsForBrokerLoad(ctx context.Context, broker *broker,
...
 
+			// The epoch has to come from whichever listing the
+			// offset comes from. The epoch of the start paired
+			// with the offset of the end makes a later
+			// OffsetForLeaderEpoch validation compare our offset
+			// against the end of a much older epoch and report
+			// data loss. An offset we calculate ourselves belongs
+			// to no listing and so has no epoch: -1 skips
+			// validating until we consume a record and learn the
+			// real epoch from it.
```
twmb: Unnecessary doc comment, delete, it's obvious on reading the code

pkg/kgo/consumer.go
twmb: Dropped the bool; the load now carries only ooorMilli, zero meaning not a reset. A last consumed record stamped exactly 0 ms is treated as no timestamp and the reset only bounds.

pkg/kgo/source.go
twmb: Reflowed the deeply nested block to about 120 columns.

pkg/kgo/consumer.go
twmb: Deleted.

pkg/kgo/topics_and_partitions.go
```diff
@@ -685,7 +685,8 @@ func (old *topicPartition) migrateCursorTo( //nolint:revive // old/new naming ma
 		// properly.
 		old.cursor.use()
 		css.reloadOffsets.addLoad(old.cursor.topic, old.cursor.partition, loadTypeEpoch, offsetLoad{
-			replica: -1,
+			replica:   -1,
+			ooorMilli: old.cursor.lastConsumedMilli(),
```
twmb, line 692: Why do we need ooorMilli here?

pkg/kgo/source.go
```diff
@@ -1464,7 +1479,8 @@ func (s *source) handleReqResp(br *broker, req *fetchRequest, resp *kmsg.FetchRe
 				// not check KIP-320 support here.
 				if partOffset.lastConsumedEpoch >= 0 {
 					reloadOffsets.addLoad(topic, partition, loadTypeEpoch, offsetLoad{
-						replica: -1,
+						replica:   -1,
+						ooorMilli: c.lastConsumedMilli(),
```
twmb, line 1487: Why do we need ooorMilli here?

pkg/kgo/source.go
```diff
@@ -1438,7 +1452,8 @@ func (s *source) handleReqResp(br *broker, req *fetchRequest, resp *kmsg.FetchRe
 				default: // partOffset.offset > fp.HighWatermark, KIP-392 case 4
 					if kip320 {
 						reloadOffsets.addLoad(topic, partition, loadTypeEpoch, offsetLoad{
-							replica: -1,
+							replica:   -1,
+							ooorMilli: c.lastConsumedMilli(),
```
twmb, line 1458: Why do we need ooorMilli here?

pkg/kgo/config.go
```diff
@@ -1616,13 +1616,14 @@ func ConsumeStartOffset(offset Offset) ConsumerOpt {
...
-// This option is *only* used if a consumer seeds OffsetOutOfRange on the
-// *first* fetch of a partition. If the consumer has consumed the partition at
-// all and sees the error, it will automatically reset to the first offset
-// after the timestamp of the last successfully consumed offset. If data loss
-// occurred such that even the last successfully consumed offset is lost, the
-// client automatically resets to the new current end offset. If you want to
-// disable offset resetting entirely, you can use [NoResetOffset].
+// This option is *only* used if a consumer sees OffsetOutOfRange before it
+// has consumed anything from a partition. If the consumer has consumed the
+// partition at all and sees the error, it resumes at the first offset after
```
twmb: "and sees OOOR, it resumes"...

pkg/kgo/consumer.go
```diff
@@ -2724,7 +2810,29 @@ func (o offsetLoadMap) buildListReq(isolationLevel int8) (r1, r2 *kmsg.ListOffse
 		}
 	}
 
-	return r1, r2
+	// A reset after consuming also lists by the last consumed timestamp.
+	// Partitions in the same batch that are not resetting list the end
+	// again, which we ignore.
```
twmb: Add quick comment about how we still need to request those partitions so that they are not erroneously pruned in the align function.

pkg/kgo/consumer.go
```diff
@@ -2626,24 +2702,25 @@ func (*Client) loadEpochsForBrokerLoad(ctx context.Context, broker *broker, load
...
-				// resets per policy on this sentinel rather than comparing
-				// -1 against the validating position.
-				offset = rPartition.EndOffset
+				// KIP-320 UNDEFINED_EPOCH_OFFSET: the broker has no
+				// record of the epoch we asked about, so it cannot
+				// tell us a truncation point. That is not data
+				// loss. We do not trust our offset either, so we
+				// reset exactly as an out of range fetch after
+				// consuming does: by the last consumed timestamp,
+				// never ahead of where we were. Java resets per
```
twmb: Nix the java line.

pkg/kgo/consumer.go
```diff
@@ -2626,24 +2702,25 @@ func (*Client) loadEpochsForBrokerLoad(ctx context.Context, broker *broker, load
...
-				// minus the false alarm. Matches the Java client, which
-				// resets per policy on this sentinel rather than comparing
-				// -1 against the validating position.
-				offset = rPartition.EndOffset
+				// KIP-320 UNDEFINED_EPOCH_OFFSET: the broker has no
+				// record of the epoch we asked about, so it cannot
+				// tell us a truncation point. That is not data
+				// loss. We do not trust our offset either, so we
+				// reset exactly as an out of range fetch after
+				// consuming does: by the last consumed timestamp,
```
twmb: Is this the best option _in this case_? Why would we receive undefined_epoch_offset? Should we just... wait and retry?

pkg/kgo/consumer.go
```diff
@@ -2490,52 +2528,90 @@ func (cl *Client) listOffsetsForBrokerLoad(ctx context.Context, broker *broker,
...
+				// record left is one we never consumed, so the
+				// start it is. Past the end, the broker lost
+				// data, and the timestamp is where we were. In
+				// range, the log changed and regrew under us:
+				// by-time re-reads what replaced our records
+				// rather than skipping them. A by-time answer
+				// ahead of us means the records between are
+				// stamped older than ours; they are unread, so
+				// we keep our offset. No answer at all means the
+				// same for everything left.
```
twmb: Make this comment wider, up to 120chars on this row. The indentation makes this squashed.

Also the end of this comment is a bit confusing.

pkg/kgo/topics_and_partitions.go
twmb: Added an inline comment: if this validation answers UNDEFINED_EPOCH_OFFSET, the reset it issues is by the last consumed timestamp, so the epoch load carries the timestamp for it.

pkg/kgo/source.go
twmb: Added an inline comment: if this validation answers UNDEFINED_EPOCH_OFFSET, the reset it issues is by the last consumed timestamp, so the epoch load carries the timestamp for it.

pkg/kgo/source.go
twmb: Added an inline comment: if this validation answers UNDEFINED_EPOCH_OFFSET, the reset it issues is by the last consumed timestamp, so the epoch load carries the timestamp for it.

pkg/kgo/config.go
twmb: Reworded.

pkg/kgo/consumer.go
twmb: Added: the request keeps every partition in the batch because alignListResps prunes a partition that is missing from any one response.

pkg/kgo/consumer.go
twmb: Removed.

pkg/kgo/consumer.go
twmb: UNDEFINED comes from the epoch cache, not from timing. The cache answers -1 in two cases: it is empty, or every epoch it holds is older than the one we ask about. An epoch older than the cache gets a real answer (the start of the first cached epoch), so retention is not a source. Empty, or ending before our epoch, means this broker never saw the epoch we consumed at: an unclean election to a replica with no history, or a leader that diverged from the one we read from.

A retry returns the same -1. A leader's cache only gains the epochs it writes itself, so our old epoch never appears in it. Waiting costs time and learns nothing.

The reset is bounded: never ahead of where we were, never below the start, never past the end. The worst case is a re-read from the divergence point, which is exactly the case we want to re-read. Java resets per auto.offset.reset on this answer, with no retry. The comment now says where the answer comes from and why asking again does not help.

pkg/kgo/consumer.go
twmb: Widened to 120 and rewrote the ending: a by-time answer ahead of us, or no answer, both mean the records from our offset on are stamped older than our last record and are unread, so we keep our offset.

twmb:
Well the comments got a bit wide but whatever. lgtm

PR #1418, opened 2026-09-10
kgo: eliminate per-call closure and heap allocation in Decompress

scunningham:
(PR desc rewritten taking into account the extra commits before merge)

# kgo: eliminate per-call allocations in Decompress

## Summary

Refactors `Decompress` to remove the `rfn` closure that was allocated on every call and the `bytes.Buffer` wrapper on the user-pooled path. Follow-up commits go further: snappy and zstd no longer touch a `bytes.Buffer` on any path; xerial-framed snappy (what the Java client writes, and what a Redpanda broker writes when it recompresses a batch to snappy) decodes straight into its destination, sized from the chunk headers; the readers wrapped around the gzip and lz4 decoders are pooled with them; and a batch that fails to decode gives the user pool its slice back.

## Motivation

Every call to `Decompress` previously allocated a closure (`rfn`) to defer the clone-vs-return-direct decision, and when a user pool was in use, created a `bytes.NewBuffer(s[:0])` wrapping the pool's slice. Both escaped to the heap on every batch decompressed. For snappy and zstd, which decode directly into a `[]byte`, the `bytes.Buffer` wrapper was pure overhead. Without a user pool, every codec decoded into an 8K pooled buffer and then cloned the result: one extra copy of every batch over 8K.

## What changed

**Snappy and zstd** take a `[]byte` destination: the user pool's slice, or nil. `s2.Decode`, zstd `DecodeAll`, and `xerialDecode` decode in place when the destination has the capacity and otherwise allocate their own exactly sized output, which is what the caller gets. No `bytes.Buffer` and no clone on either path.

**Xerial-framed snappy** walks the chunk headers first (four bytes and a varint per chunk, skipping the data) to size the output once, then decodes each chunk straight into the destination's spare capacity. Previously each chunk went through a 32K scratch slice and an `append`, growing the destination per chunk.

**Gzip and lz4** still stream through a `bytes.Buffer`: the user pool's slice wrapped on the stack, or the internal pooled buffer followed by a clone. `Buffer.ReadFrom` is called directly rather than through `io.Copy` so the stack buffer does not escape, and the `bytes.Reader` and `io.LimitedReader` around each codec now live in the pooled decoder object.

**User pool on error**: a batch that fails to decode yields no records, so nothing would ever Recycle the slice taken from `PoolDecompressBytes`. `Decompress` now puts it back itself, with the full capacity zeroed. The `PutDecompressBytes` doc now says the decompressed bytes (the slice's length) are what is zeroed on the success path.

**Benchmark**: `BenchmarkDecompress` runs every codec with and without a user pool, including a xerial-framed snappy input.

## Benchmark results (300 KB decompressed payload)

allocs/op and B/op per `Decompress` call, master vs this PR.

### With a user pool

| codec | allocs master | allocs PR | B/op master | B/op PR |
|---|---|---|---|---|
| snappy | 2 | **0** | 64 | 0 |
| snappy (xerial) | 3 | **0** | 32,832 | 0 |
| zstd | 2 | **0** | 95 | 31 |
| gzip | 4 | **0** | 150 | 14 |
| lz4 | 5 | **1** | 1,647 | 1,511 |

### Without a user pool

| codec | allocs master | allocs PR | B/op master | B/op PR |
|---|---|---|---|---|
| snappy | 2 | **1** | 623,070 | 311,303 |
| snappy (xerial) | 10 | **1** | 1,475,748 | 311,298 |
| zstd | 4 | **1** | 624,171 | 311,884 |
| gzip | 3 | **1** | 312,989 | 312,518 |
| lz4 | 4 | **2** | 349,427 | 347,952 |

Without a user pool the remaining allocation is the output slice itself. gzip and lz4 keep the clone out of the pooled buffer, since their output size is unknown until the stream ends; lz4's second allocation is a per-stream struct inside the lz4 library. Xerial without a pool takes about half the time it did: the per-chunk growth was zeroing and copying about 1.4 MB per 300 KB batch.

## Behavioral note

The snappy and zstd paths without a user pool now return the decoder's own allocation rather than a clone of it. For a zstd frame with no content size in its header (Java's streaming writer), the decoder starts at min(2x input, 1 MB) and grows, so the returned slice's capacity can exceed its length.


twmb:
While you're working in the decompress code, want to take on https://github.com/twmb/franz-go/pull/1366 (review / kill it / rewrite it / anything)? I'm trying to get a release out within 1-2 weeks. with effectively all PRs merged.
It's entirely LLM generated at the moment (it's something I've always wanted done, but never wanted to do), it will require changes once I finally get #1416 into a shape I'm happy with, and it's complicated.

No obligation though. I'll review this soonish.

scunningham:
> While you're working in the decompress code, want to take on #1366 (review / kill it / rewrite it / anything)? I'm trying to get a release out within 1-2 weeks. with effectively all PRs merged. It's entirely LLM generated at the moment (it's something I've always wanted done, but never wanted to do), it will require changes once I finally get #1416 into a shape I'm happy with, and it's complicated.
> 
> No obligation though. I'll review this soonish.

I took a quick look at #1366 .  That's pretty complicated.  Tough because you don't know how much data to add to the compressor until it's done, and if you are over the threshold you have to restart.

Might be less complex to merge into one merged block, push it into the compressed buffer and then do surgery on the buffer and split into N payloads.  I know this is possible in lz4; pretty easy actually, but I'm biased because I wrote a [wrapper library](https://github.com/prequel-dev/plz4).  Unclear if that approach would be possible using the other codecs; but probably.

scunningham:
> Might be less complex to merge into one merged block, push it into the compressed buffer and then do surgery on the buffer and split into N payloads. I know this is possible in lz4; pretty easy actually, but I'm biased because I wrote a [wrapper library](https://github.com/prequel-dev/plz4). Unclear if that approach would be possible using the other codecs; but probably.

Yeah that won't work either because of the record boundaries.  You'd have to break on the frame size and flush. Not ideal.

twmb:
The theory of claude is that there is a maximum fixed overhead size for the compression, and (my own theory) - we have access to the underlying slice, so if adding a record causes us to exceed the maximum size, we back out of the change and cut the prior batch.

scunningham:
> The theory of claude is that there is a maximum fixed overhead size for the compression, and (my own theory) - we have access to the underlying slice, so if adding a record causes us to exceed the maximum size, we back out of the change and cut the prior batch.

I refactored the PR a bit for better or worse.  Appears logically the same; and perhaps slightly easier to understand.  Cannot attest to the logic validity.   The compression logic seems sound, but I don't pretend to fully understand the commit sequence.

twmb:
Pushed a few extra commits, squashed yours and one of mine into one single commit; should be much fewer allocs now. I'll merge, it'll be out with whatever the next release is.

scunningham:
> Pushed a few extra commits, squashed yours and one of mine into one single commit; should be much fewer allocs now. I'll merge, it'll be out with whatever the next release is.

Thanks!

PR #1419, opened 2026-09-11
kgo: add Offset.Lookback

twmb:
The API half of the out-of-range reset work, on top of #1428. #1428 ships the thirty second rewind, the `ConsumeResetOffset` policy on unlocatable loss, the bounding, and `ErrDataLoss`, all without new API so it can go in a patch. This adds the one exported method, for a minor.

### `Offset.Lookback(d)`

The first offset at or after `d` before a reference point, with the timestamp computed when the client builds the request rather than when you build the offset.

- As a reset offset, the reference is the last record the client consumed. `Lookback(30*time.Second)` is the default #1428 already ships; this lets you widen it for producers whose delivery latency spreads further, or narrow it to zero, which is exactly what #1416 merged.
- As a start offset, the reference is now, so `Lookback(time.Hour)` is the last hour of a partition as of when that partition is resolved. That is Kafka's `auto.offset.reset=by_duration`, which we could not express before.

`AfterMilli` keeps its role. `AfterMilli(time.Now().UnixMilli())` pins an instant, so a partition created later is read from its start; `Lookback(0)` resolves against the clock at discovery and would start that partition at its end. The two docs point at each other.

A negative duration clamps to zero. The millisecond floor and wrap guard are in #1428; this only adds the start-role resolution in `buildListReq` on top of them.

### Tests

`TestLookbackStartOffset`: records a minute apart over ten minutes, `ConsumeStartOffset(NewOffset().Lookback(5*time.Minute))`, requires starting about five records in. `TestLookbackClampsNegative` covers the clamp.


twmb:
Folded into #1428.

The two were cut at the wrong seam: #1428 carried the lookback mechanism, the `lookback`/`hasLookback` fields and the 30s `ConsumeResetOffset` default, while this PR added the only way to control it. Merging #1428 alone would have changed the reset default with no exported API to name, tune or disable it.

`Offset.Lookback` and the mechanism now live in the same commit in #1428, so that intermediate state cannot exist in history. `TestLookbackClampsNegative` was collapsed into the `TestOffsetLookback` table there, and `TestLookbackStartOffset` now uses `produceSpaced` instead of its own produce loop.

PR #1425, opened 2026-09-13
kgo: assign a regex topic added back while the broker still assigns it

twmb:
Purging a regex-consumed topic under KIP-848 leaves it unconsumed forever. Found while building the topic recreation hard-fail PR (#1426), where "purge and let the regex rediscover it" is the recovery.

The broker resolves the regex itself, so the subscription cannot leave the purged topic out and the broker keeps assigning it. When metadata adds the topic back, the assignment is the same by name, no session restarts, and the fresh cursors are never assigned. `findNewAssignments` now notes a topic it starts using while `nowAssigned` still holds it; the heartbeat restarts the session when its assignment names such a topic; `diffAssigned` treats it as added in full.

The broker also sends the assignment only to a full heartbeat or when the assignment changes. If the regex rediscovers the topic before the next heartbeat goes out, that heartbeat subscribes to and owns exactly what the last one did, so it is a keepalive, whose response carries nothing to restart the session with. While a topic waits in `reassign`, the heartbeat is sent in full.

Measured on kfake, each subtest run 30x with anchored `-run` names:

| | master | this PR |
|---|---|---|
| `TestPurgeRegexTopic/848` | 30/30 fail | 0/30 fail |
| `TestPurgeRegexTopic/classic` | 0/30 fail | 0/30 fail |
| `TestPurgeRegexTopicRediscoveredFirst` (848, keepalive) | n/a | 0/30 fail |

Classic passes on master because the client resolves the regex there: a purged topic leaves the subscription, so rediscovering it is an ordinary add. The fix is not 848 only, though. `reassign` lives in `consumer_group.go` and is populated in shared code, because `diffAssigned` is where a topic has to become "added in full"; only the two reads in `consumer_group_848.go` are 848 specific. Any consumer that has metadata add a topic back while the assignment still names it takes the same path. KIP-848 is where that is forced rather than incidental, which is why it is the case with a failing repro. `-race` clean; golangci-lint clean in both modules.

A note on what is deliberately *not* here: an earlier commit made the purge itself trigger a metadata refresh, so a regex consumer rediscovered the topic within `MetadataMinAge` instead of up to `MetadataMaxAge`. That is dropped. Purging is the user saying "stop consuming this"; the client should not immediately work to undo it. The tests call `ForceMetadataRefresh` themselves, which is what a user who wants the rediscovery now would do.


pkg/kgo/consumer_group.go
```diff
@@ -50,6 +50,14 @@ type groupConsumer struct {
 
 	rejoinCh chan string // cap 1; sent to if subscription changes (regex)
 
+	// reassign holds topics that metadata added back to using while the
+	// broker still assigned them, so their cursors are new but the
+	// assignment did not change by name: a regex topic purged and then
+	// rediscovered under KIP-848, where the broker resolves the regex
```
twmb: under kip-848 is a bit stale of a doc, since this applies to classic as well now.

twmb: Right, the KIP-848 framing was wrong. `reassign` lives here and is populated in shared code; only the two reads in `consumer_group_848.go` are 848 specific, and classic hits the same shape whenever metadata adds a topic back while the assignment still names it. Dropped the clause:

```
// reassign holds topics that metadata added back to using while the
// broker still assigned them, so their cursors are new but the
// assignment did not change by name. A regex topic that is purged
// and then rediscovered does this. The next session treats them as
// added; see diffAssigned. Guarded by mu.
```

Folded into the commit and rebased onto master.

PR #1426, opened 2026-09-13
kgo: make the topic recreation hard failure hold everywhere

twmb:
A deleted and recreated topic keeps its name and gets a new topic ID. kgo already refuses to move a name to a new ID -- cursors and record buffers keep the ID they were created with -- but that failure was not reliable, and in three places it was silent. This makes it hold, and makes the documented recovery (purge, then add the topic back) work.

Alternative to #1364: no ID is ever adopted, nothing heals itself, no new option.

### What was broken

- **Offsets from the old topic were committed to the new one.** OffsetCommit v10 carried whatever ID the client held, and the client re-adopted the new ID on the merge, so a Kafka 4.2 broker stored the old topic's offsets under the recreated topic. A consumer that restarted then skipped everything below that offset. Below v10 the commit goes by name and the broker cannot refuse it at all -- that is every Redpanda and every Kafka below 4.2.
- **A recreated topic could stall with no error at all.** A broker resolves a fetch session partition's name once, when the partition enters the session, so after a recreation it answers that entry INCONSISTENT_TOPIC_ID on every fetch. kgo strips that error. With any other partition in the same session, the fetch loop spun at round-trip speed forever and the user's poll never returned anything.
- **Purge and add the topic back did not work.** A classic cooperative member that self-revoked a purged topic kept it in `lastAssigned`, so when the re-add's metadata landed after that revoke the next session found the topic in both last and now and never assigned its fresh cursors. Share consumers never sent a full heartbeat, so the broker never re-sent the assignment.
- Partitions the recreated topic had and the old one did not took the new ID and consumed the new topic while the rest stalled. A fetch carries one ID per topic, so whichever cursor was added first decided which topic the whole request read.
- A KIP-848 or share heartbeat assigns by ID, and id2t refused the new one, so the client refreshed metadata on every heartbeat forever.

### What it does

- **Never adopts the new ID.** New partitions of a recreated topic take the old one. Produce buffers are abandoned with UNKNOWN_TOPIC_ID, the same path a purge uses.
- **Acts on the first metadata response that reports a new ID**, as the Java client does when it adopts one. A broker that missed a recreation can report the prior ID for a topic whose current ID the client holds; the client then fails that live topic until it is purged and re-added. Waiting for a second broker to agree was worse: while the brokers disagreed the buffers either waited or drained, and a drain below produce v13 went by name to the new topic.
- **Commits carry the ID the offsets were fetched under**, so a broker that keys commits by ID refuses them itself. The client refuses them too, which is what covers by-name commits: Kafka below 4.2, every Redpanda, and TxnOffsetCommit at every version this client sends. Marking and setting offsets keep the entry's ID. An entry with no ID of its own (the group's committed offsets, or a fetch below Kafka 3.1, which carries no ID) takes the ID the topic's cursors hold, which they keep through a recreation; that covers the window where metadata reports the deleted topic with a zero ID and nothing has reported the new one yet. A recreated topic's offsets never enter the wire request; the client answers them UNKNOWN_TOPIC_ID and puts them back into the request the commit callback sees.
- **Detecting a recreation stops the consumer session**, so the next fetch is a full one carrying our ID, which the broker answers UNKNOWN_TOPIC_ID. A per-partition INCONSISTENT_TOPIC_ID also resets the session. Both are needed: Kafka answers INCONSISTENT_TOPIC_ID forever on the session, and Redpanda has no such error at all, so only the session stop breaks a name-keyed session there.
- **A transaction that produced to the topic cannot commit.** `EndTransaction` refuses TryCommit with an error wrapping TransactionAbortable and UnknownTopicID; `GroupTransactSession.End` aborts. A request still on the wire counts, because its response can no longer mark the partition added.
- **id2t moves a name to its new ID** rather than keeping every ID a name has had: every reader of that map wants the ID the broker uses now, and the map is cloned on every metadata merge, so it must not grow per recreation.
- **Purge and add back works** on all three protocols. KIP-848 already had it through `reassign` (#1425). Classic needs a self-revoked topic dropped from `lastAssigned` so the next session sees it as newly assigned; this PR had that fix, and #1442 reached the same line from a real `TestPurgeRegexTopic/classic` failure and is on master, so the commit here is dropped. A share purge forces the next heartbeat to be a full one.

### Windows that remain, by broker

Detection needs topic IDs in metadata, which is Kafka 2.8+. Below that nothing tells the two topics apart.

| | by ID from | by name below that |
|---|---|---|
| fetch | 3.1 | reads the new topic at old offsets until purged; offsets still refused |
| produce | 4.1 | a request already on the wire can land in the new topic |
| offset commit | 4.2 | refused by the client instead of the broker |

Transactional offset commits carry names only until TxnOffsetCommit v6 (KIP-1319): the client refuses them once metadata has seen the new ID, and one sent before that can land in the new topic.

Two cases key by name whatever the broker version: a transactional producer on a cluster without KIP-890 part 2, which caps produce below the version that carries an ID, and Redpanda, which does the same for produce and for offset commits.

The README has this under `## Topic recreation`.

### State

One `[16]byte` per topic (the ID it was recreated from), one per uncommitted entry, one `id2t` entry per name, and one small map per open transaction. Nothing grows per recreation, and no hot path gained work.

### Series

The first commit adds `noID`, the zero topic ID, and replaces every `[16]byte{}` and local zero in the package with it. The rest of the series is built on it. The new kfake tests are in `package kfake`, as #1441 made every other kfake test.

### Testing

kfake grew nothing for this; it matches Kafka after #1429. Tests are in `pkg/kfake`, and every one of them fails with the code it guards removed. The tests for the two-response confirmation and the produce hold went with that mechanism.

The tests had grown their own broker plumbing before kfake had any, and the last commit hands that work to what #1435 added: an observing fault counts requests, `GroupInfo` reads the group's commits, `PartitionInfo` reads a partition's end offset, `CreateTopic` and `DeleteTopic` recreate the topic, and a fault with `When` kills the first full heartbeat after a purge and re-add. What is left is what a fault cannot do: sleeping a request at the broker, and answering metadata with a canned response. The regex purge test had its own copy of the sleeping helper and now calls the shared one.

Race suite clean twice, plain suite clean, the recreation and purge tests clean at `-count=20` under `-race`, golangci-lint clean in both modules, and every commit builds and vets on its own.


pkg/kgo/client.go
```diff
@@ -5,6 +5,28 @@
...
+// the new topic in the meantime: below Kafka 3.1, fetches carry the name and
+// a consumer keeps reading the new topic at its old offsets until the topic
+// is purged, though its offsets stop being committed; below 4.1 a produce
+// request already sent when the client learns of the recreation can land in
+// the new topic, and below 4.2 so can an offset commit, as can a
+// transactional offset commit at the versions this client sends, since
+// those carry no topic ID. Two cases key by name whatever the broker
+// version: a transactional producer on a cluster without KIP-890 part 2,
+// which caps produce below the version that carries a topic ID, and
+// Redpanda, which does the same for produce and for offset commits.
```
twmb: This doc does not go here. We only want this in the readme. Bullet list form.

pkg/kgo/client.go
```diff
@@ -698,6 +724,9 @@ func (cl *Client) Ping(ctx context.Context) error {
 // still exists on the broker, this function will at most only temporarily
 // remove the topic from the client and the topic will be re-discovered.
 //
+// Purging a topic and adding it back is also how to resume after a topic was
+// deleted and recreated; see the package documentation.
```
twmb: see the readme for more details. (not package documentation)

pkg/kgo/client.go
```diff
@@ -721,28 +750,26 @@ func (cl *Client) PurgeTopicsFromClient(topics ...string) {
...
 		}
 		cl.metaCache.mu.Unlock()
 
-		if len(purgedIDs) > 0 {
-			old := cl.id2tMap()
-			merged := make(map[[16]byte]string, len(old))
-			maps.Copy(merged, old)
-			for _, id := range purgedIDs {
-				delete(merged, id)
+		// id2t holds every ID a name has had; a purge forgets them all.
```
twmb: errr what? Like every id, always, forever? so if I spin loop recreate topics, this keeps growing?

pkg/kgo/source.go
```diff
@@ -1146,6 +1146,21 @@ func (s *source) fetch(consumerSession *consumerSession, doneFetch chan<- bool)
...
+	// The broker resolves a session partition's name once, when the
+	// partition enters the session. If that name now belongs to a
+	// different topic ID (the topic was recreated), every fetch of the
+	// session answers the partition INCONSISTENT_TOPIC_ID. Only a full
+	// fetch sends our ID again, and that one is answered
+	// UNKNOWN_TOPIC_ID.
+	//
+	// We strip the error, so without this the fetch loop can spin: an
+	// errored partition completes the fetch at once unless the same
+	// fetch is also reading from tiered storage.
```
twmb: "and that one is answered unknown_topic_id" makes no sense.

Also, reword final paragraph:
// We strip inconsistent_topic_id and unknown_topic_id, so we need
// to reset the session: if we don't, fetching can spin loop (brokers return
// error responses immediately, no waiting for other data on other partitions).

pkg/kgo/consumer.go
```diff
@@ -848,6 +848,17 @@ func (c *consumer) purgeTopics(topics []string) {
 	if c.g != nil {
 		c.g.mu.Lock() // required when updating using
 		defer c.g.mu.Unlock()
+		// A recreated topic's uncommitted offsets belong to the old
+		// topic. The rebalance this purge causes commits what we have
+		// uncommitted, and by then the topic is gone from tps, so the
+		// commit could not tell; we drop the offsets now. See
+		// uncommittedFrom.
```
twmb: what is this word vomit trying to say?

pkg/kgo/consumer_group.go
```diff
@@ -2590,11 +2603,17 @@ func (g *groupConsumer) updateUncommitted(fetches Fetches) {
 					final.Offset + 1,
 				}
 				prior, ok := topicOffsets[partition.Partition]
-				if !ok {
+				// Records from a different topic ID than the entry's
+				// come from a recreated topic that was purged and
+				// added back; the old topic's offsets are gone with it.
+				if !ok || prior.topicID != ([16]byte{}) && topic.TopicID != ([16]byte{}) && prior.topicID != topic.TopicID {
```
twmb: introduce `var noID [16]byte` in topics_and_partitions.go and replace all usages of this [16]byte{} throughout the entire package with it, as the first commit.

pkg/kgo/consumer_group.go
```diff
@@ -2980,6 +2999,32 @@ func (cl *Client) CommittedOffsets() map[string]map[int32]EpochOffset {
...
+			break
+		}
+	}
+	td := tps.loadTopic(topic)
+	if from == ([16]byte{}) && td != nil {
+		// Nothing told us which topic these offsets came from: they
+		// were marked, set, or fetched from the group rather than
+		// polled, or the broker's fetch response carries no topic ID
+		// (below Kafka 3.1). If the topic was recreated, they are from
+		// before that, because we drop them when it is purged.
```
twmb: "they are from before that"?????????????

pkg/kgo/consumer_group.go
```diff
@@ -3563,6 +3610,49 @@ func (g *groupConsumer) commit(
 	req.InstanceID = g.cfg.instanceID
 	is848 := g.is848 // g.mu is held, per the function comment above
 
+	// We build the topics now, under g.mu, which a purge also holds: the
+	// purge of a recreated topic drops its offsets, and a commit built
+	// after the purge would no longer see the topic to refuse it.
```
twmb: "no longer see the topic to refuse it"?????

pkg/kgo/consumer_group.go
```diff
@@ -3657,17 +3725,42 @@ func (g *groupConsumer) commit(
...
+				if _, ok := recreated[rt.Topic]; !ok {
+					kept = append(kept, rt)
+					continue
+				}
+				d := droppedTopic{name: rt.Topic, id: rt.TopicID, err: kerr.UnknownTopicID}
+				for _, rp := range rt.Partitions {
+					d.partitions = append(d.partitions, rp.Partition)
+				}
+				dropped = append(dropped, d)
+			}
```
twmb: Why not just avoid adding them to the request ABOVE when it is built? And track them dropped there?

pkg/kgo/consumer_share.go
```diff
@@ -55,6 +55,14 @@ type (
 		lastSentSubscribedTopics []string
 		lastSentRack             bool
 
+		// resubscribe makes the next heartbeat carry
+		// SubscribedTopicNames even if they did not change. A purge
+		// sets it: the broker returns the assignment only to a full
+		// request or when it changes, and a topic purged and added
+		// back before the next heartbeat leaves both as they were,
+		// so the fresh cursors would never be assigned.
```
twmb: "leaves both as they were" - WHAT?

pkg/kgo/consumer_share.go
```diff
@@ -1096,6 +1104,12 @@ func (sc *shareConsumer) heartbeat() (time.Duration, error) {
 		req.RackID = &sc.cfg.rack
 	}
 
+	// Take resubscribe before the topics: a purge landing after this
+	// sets it again, and the next heartbeat is the one that carries the
+	// topics as that purge left them.
```
twmb: Honestly this doc is just confusing and not worth adding.

pkg/kgo/metadata.go
```diff
@@ -828,6 +814,65 @@ func (cl *Client) mergeTopicPartitions(
 		lv.writablePartitions = r.writablePartitions
 	}()
 
+	// Topic IDs are never reused, so metadata reporting a different ID
+	// for a topic we hold means the topic was deleted and recreated. We
+	// do not adopt the new ID: our cursors and record buffers keep the
+	// old one and fail with UNKNOWN_TOPIC_ID until the topic is purged
+	// and re-added (see cursor.topicID). We remember the old ID so that
+	// partitions added after the recreation fail the same way and so
+	// that offsets from the old topic are not committed.
```
twmb: "we remember the old id" sentence is confusing.

Also: this only remembers one ID back? What happens if a topic is deleted and recreated twice and we see both - or if the cluster is very unstable and we see three IDs for the same topic, only one of which is valid?

pkg/kgo/metadata.go
```diff
@@ -828,6 +814,65 @@ func (cl *Client) mergeTopicPartitions(
...
+	//
+	// We act on the first response that reports a new ID. A broker that
+	// missed a recreation can report the prior ID for a topic whose
+	// current ID we hold, and we then fail a live topic until it is
+	// purged and re-added. Waiting for a second broker to agree would
+	// have the buffers wait or drain while the brokers disagree, and a
+	// drain below produce v13 goes by name to the new topic.
```
twmb: This paragraph isn't useful.

pkg/kgo/metadata.go
```diff
@@ -828,6 +814,65 @@ func (cl *Client) mergeTopicPartitions(
...
+		default:
+			oldID = lv.partitions[0].cursor.topicID
+		}
+		if oldID != noID && oldID != r.id {
+			lv.recreatedFrom = oldID
+			// A fetch session serves a partition under the name it
+			// was added with, and the broker resolves the name once:
+			// the new topic answers as long as the session lasts.
+			// Only a full fetch sends our ID, which the broker fails.
+			// Stopping the session starts every source over.
```
twmb: Delete this paragraph.

pkg/kgo/metadata.go
```diff
@@ -828,6 +814,65 @@ func (cl *Client) mergeTopicPartitions(
...
+			if kind == partitionKindConsume {
+				css.stop()
+			}
+			// Cursors keep fetching under the old ID and the broker
+			// fails them. Produce requests below v13 carry the topic
+			// name, which the new topic answers to, so we fail the
+			// buffers ourselves: everything buffered fails now, and
+			// anything buffered later fails at once.
```
twmb: Delete this paragraph.

pkg/kgo/metadata.go
```diff
@@ -828,6 +814,65 @@ func (cl *Client) mergeTopicPartitions(
...
+			if isProduce {
+				for _, tp := range lv.partitions {
+					tp.records.abandon(kerr.UnknownTopicID)
+				}
+				// A transaction that already produced to the topic
+				// cannot commit; see EndTransaction.
+				if cl.cfg.txnID != nil {
```
twmb: Delete the doc

pkg/kgo/metadata.go
```diff
@@ -828,6 +814,65 @@ func (cl *Client) mergeTopicPartitions(
 		lv.writablePartitions = r.writablePartitions
 	}()
 
+	// Topic IDs are never reused, so metadata reporting a different ID
```
twmb: The docs on this entire block need to say what it does up top, succinctly:
// We never adopt new topic IDs from recreated topics. Users must purge and re-add the topic if they want them to work.
// When we detect recreation, we permanently save the original ID we saw into recreatedFrom. This allows us to fail producing/consuming going forward.
//  * For consumers, stopping our consumer session and restarting it removes the topic from any broker fetch session state.
// * For producers, we fail anything client side ourselves. It is possible that a few records will get through to the new topic due to automatic OOOSN recovery before metadata catches it and locks.
// *  Any active transaction will be failed.

pkg/kgo/metadata.go
```diff
@@ -878,6 +923,13 @@ func (cl *Client) mergeTopicPartitions(
 		}
 		newTP := r.partitions[part]
 
+		// A recreated topic's buffers are abandoned and out of every
+		// sink; there is nothing to migrate, re-add, or clear.
```
twmb: Drop the doc

pkg/kgo/metadata.go
```diff
@@ -1056,9 +1107,27 @@ func (cl *Client) mergeTopicPartitions(
 			}
 			retryWhy.add(topic, newTP.partition(), newTP.loadErr)
 		}
+		// A partition that only the recreated topic has takes the old
+		// ID, so that it fails like the rest of the topic rather than
+		// consuming or producing the new topic. A fetch request carries
+		// one ID per topic, from whichever cursor is added first, so a
+		// partition holding the new ID would also have the old
+		// partitions' offsets fetched against the new topic.
```
twmb: Drop the doc.

pkg/kgo/producer.go
```diff
@@ -45,6 +47,14 @@ type producer struct {
 	id           atomic.Value
 	producingTxn atomic.Bool
 
+	// recreatedInTxn holds topics the current transaction produced to
+	// before they were deleted and recreated. The records the old topic
+	// acknowledged are gone with it, so the transaction cannot commit;
+	// see EndTransaction. Cleared when the next transaction begins, so a
+	// retry that puts this one back cannot commit it either.
```
twmb: "retry that puts this one back"????

pkg/kgo/producer.go
```diff
@@ -277,39 +287,49 @@ func (p *producer) purgeTopics(topics []string) {
...
 		}
 	}
 }
 
 func (p *producer) isAborting() bool { return p.aborting.Load() > 0 }
 
+// noteRecreatedInTxn records that the current transaction produced to a
+// topic that has since been recreated, if any partition of the topic was
+// added to the transaction or may still be. A request on the wire may have
+// been written: its response can no longer mark the partition added, because
```
twmb: The final sentence in this is confusing. In general that's the case for all final sentences. Wtf?

pkg/kgo/topics_and_partitions.go
```diff
@@ -657,9 +662,15 @@ func (old *topicPartition) migrateProductionTo(new *topicPartition) { //nolint:r
...
 // have fewer concurrency issues to worry about.
+//
+// A recreated topic's cursor is not validated: its offset and epoch belong to
+// the old topic, and the cursor fails with UNKNOWN_TOPIC_ID rather than
+// consume the new one, so validating against the new topic could only
+// report a data loss that never happened.
 func (old *topicPartition) migrateCursorTo( //nolint:revive // old/new naming makes this clearer
 	new *topicPartition,
 	css *consumerSessionStopper,
+	validate bool,
```
twmb: should honestly be named 'wasRecreated' because that's the only reason this is set

pkg/kgo/client.go
twmb: Moved to the README as a `## Topic recreation` bullet list, one bullet per fact: the purge-and-re-add recovery, commits under the fetched ID, detection from metadata at 2.8, and the by-name windows below 3.1, 4.1, and 4.2, for TxnOffsetCommit, and for Redpanda. The package doc is back to what master has.

pkg/kgo/client.go
twmb: Now reads: "Purging a topic and adding it back is also how to resume after a topic was deleted and recreated; see the README for more details."

pkg/kgo/client.go
twmb: It does not grow. The merge in metadata.go moves a name to its new ID and drops the old one, so id2t holds one entry per name however many times a topic is recreated, and this loop only drops the purged names. The comment described an earlier version of the commit; it now says "Drop the purged names from id2t."

pkg/kgo/source.go
twmb: Reworded. The first paragraph now says the recreated topic's session partition reaches the new topic, so every fetch of the session answers INCONSISTENT_TOPIC_ID, and a full fetch carries our ID and is answered UNKNOWN_TOPIC_ID. The final paragraph is your text verbatim.

pkg/kgo/consumer.go
twmb: Now: "The rebalance this purge causes commits our uncommitted offsets. By then the topic is gone from tps and uncommittedFrom cannot tell it was recreated, so we drop its offsets now."

pkg/kgo/consumer_group.go
twmb: Done as the first commit of the series (kgo: add noID, the zero topic ID): `var noID [16]byte` in topics_and_partitions.go, and every `[16]byte{}` and local `var zeroID` / `var noID` in pkg/kgo, tests included, now uses it. The rest of the series is rebased on top and uses it as well.

pkg/kgo/consumer_group.go
twmb: That fallback is gone. An entry with no ID of its own (the group's committed offsets, or a fetch below Kafka 3.1) now takes the ID the topic's cursors hold, which they keep through a recreation; the comment says that in three lines. This also closes a gap: while metadata reports the deleted topic with a zero ID, recreatedFrom is still zero too, so such a commit went by name at v9. SetOffsets also used to wipe the entry's ID; it keeps it now.

pkg/kgo/consumer_group.go
twmb: Now: "Build the request under g.mu: a purge, which also holds g.mu, removes the topic from tps, and a request built after it would commit a recreated topic's offsets by name."

pkg/kgo/consumer_group.go
twmb: Done. The build loop appends a recreated topic to `refused` instead of `req.Topics`, so it never enters the wire request, and the `recreated` map, the second pass, and the `err` field on droppedTopic are gone. What stays after the request is answering them: updateCommitted skips the whole response when the request and response topic counts differ, so the fill step synthesizes UNKNOWN_TOPIC_ID for each refused topic and appends those topics back onto the request that onDone sees.

pkg/kgo/consumer_share.go
twmb: Now: "resubscribe makes the next heartbeat send the subscribed topics even if they did not change. A purge sets it: the broker returns the assignment only to a full heartbeat, and a topic purged and added back before the next heartbeat needs its new cursors assigned."

pkg/kgo/consumer_share.go
twmb: Deleted. A purge that lands after the Swap sets the flag again and the next heartbeat is a full one, so nothing depends on the order.

pkg/kgo/metadata.go
twmb: That sentence went with its paragraph: the block's doc is now your text (never adopt, save the original ID into recreatedFrom, the three bullets).

pkg/kgo/metadata.go
twmb: Deleted.

pkg/kgo/metadata.go
twmb: Deleted.

pkg/kgo/metadata.go
twmb: Deleted.

pkg/kgo/metadata.go
twmb: Deleted.

pkg/kgo/metadata.go
twmb: Done: your text, wrapped to the file's width with the bullets under a blank comment line. Each bullet matches the code: css.stop() restarts the consumer session, abandon fails the buffers (an OOOSN resend can still land by name if the metadata refresh before it has not reported the new ID yet), and noteRecreatedInTxn makes EndTransaction refuse the commit.

pkg/kgo/metadata.go
twmb: Dropped.

pkg/kgo/metadata.go
twmb: Dropped.

pkg/kgo/producer.go
twmb: Now: "recreatedInTxn holds topics the current transaction produced to before they were deleted and recreated. The transaction cannot commit; see EndTransaction. BeginTransaction clears it."

pkg/kgo/producer.go
twmb: Now: "noteRecreatedInTxn records that the current transaction produced to a topic that was then recreated. A partition counts if it was added to the transaction or has a request in flight: the buffer is abandoned, so the response can no longer mark it added." I also went through every comment this branch adds and cut the closing sentence wherever it only restated the one before it: the id2t merge, uncommit.topicID, the commit request build, the refused fill, commitTxn, sink.abandon in two places, recreatedFrom, and migrateCursorTo.

pkg/kgo/topics_and_partitions.go
twmb: Renamed to wasRecreated. Both callers pass the fact itself (`recreated` in mergeTopicPartitions, `recreatedFrom != noID` in doMove) and the KIP-320 check reads `!wasRecreated`.

README.md
```diff
@@ -215,6 +215,32 @@ Kafka 0.10.0 introduced the ApiVersions request; if you are working with
...
+  the recreation can land in the new topic.
+- Below Kafka 4.2, an offset commit already sent when the client learns of the
+  recreation can land in the new topic.
+- Transactional offset commits carry names only until TxnOffsetCommit v6
+  (KIP-1319), so the client refuses them only once metadata has seen the new
+  ID, and one sent before that can land in the new topic.
+- Redpanda caps produce and offset commits below the versions that carry a
+  topic ID, so both go by name at every Redpanda version. A transactional
+  producer on a cluster without KIP-890 part 2 produces by name as well.
+
```
twmb: The client always produces to or consumes from the _first_ instance of a topic. If you delete and recreate a topic, and if the client can detect it (i.e. the broker is modern enough and is using topic IDs), the client will refuse to produce to or consume from the new instance of the topic.

You can handle this in your application by using `PurgeTopicsFromClient`, which purges all information about a topic from within the client, and then (for consumers) calling `AddConsumeTopics`. Producers will auto recover after purging once you produce again.

Depending on your broker version, you may experience different side effects from trying to keep a client alive across topic recreation:
* Producing pre-4.1 while recreating a topic may result in batch 2 being before batch 1 on the new topic, and may result in some duplicates.
* Consumign pre-3.1 could result in the new topic's messages being skipped, since the consumer will stay at the offset it was at in the old topic.
* Committing pre-4.2 could result in commits for the old topic landing for the new topic, and consumers picking up these stale commits and skipping messages
* Transactional commits pre-4.4 have the same problem.

It is not recommended to try to keep a client alive across a topic recreation, _but_ if you want to do so: always upgrade to the latest Kafka. Doing so will make this client hard-lock more reliably and avoid side effects; you can then purge and add as needed.

twmb: Applied your text verbatim, wrapped to the file's width, with one typo fixed: "Consumign" -> "Consuming". Folded into `kgo: document topic recreation in the README`; the section is now yours end to end and the old bullet list is gone.

PR #1427, opened 2026-09-13
kgo: FollowRecreatedTopics purges and re-adds a recreated topic itself

twmb:
Rebased onto master now that #1426 is merged. The eight commits this branch carried from #1426's earlier drafts are gone; the kfake fetch session change is on master as `958d2292`. What is left is this PR's own two commits, rebuilt on the merged code: the recreation is acted on from the first metadata response that reports a new ID, so there is no confirmation step, no produce hold from the base, and no `suspectID`.

`FollowRecreatedTopics()` opts into the client purging a deleted and recreated topic and adding it back itself, instead of failing the topic until you do. When the merge sees the new ID it dispatches a goroutine (purging runs a blocking metadata fn on the metadata loop) that purges that side of the topic and re-adds it: `AddConsumeTopics`, `AddConsumePartitions` at the reset offset for a direct consumer of specific partitions, nothing for regex (rediscovered), the same for share consumers. Consumers resume from the new topic's committed offsets or per the reset policy.

The second commit keeps a producer's buffered records: the buffers are held from the moment the merge sees the new ID, the follow waits for produce requests in flight (so a batch the broker accepted by name is not produced twice and one it rejected is carried), removes the topic like a purge but takes the records instead of failing them, and produces them anew, in order, once the topic loads again. A partitioner that loaded the topic before the carry produces its record anew; after the carry a produce waits on the topic as an unknown one.

Known costs: purge + re-add is a rebalance per heal (revoke, then assign), and the heal takes seconds rather than sub-second. A transaction that produced to the topic before the recreation still cannot commit (#1426's rule), so the carried records are aborted with it and produced again by your retry. Since #1426 acts on the first response, a broker that lags behind a recreation makes this option purge and re-add a live topic; the topic reloads from the next response.

Tests: direct, `ConsumePartitions`, classic group with a commit landing for the new topic, 848, and a producer by ID and by name asserting the carried records land exactly once and nothing is failed. The README's `## Topic recreation` section names the option.


twmb:
Not liking this, closing. It's easy enough to do in an application. Following in the client is NOT easy and is riddled with edge case downsides (duplicates, out of order produce requests, etc). See also #1364 for a longer more comprehensive attempt.

PR #1428, opened 2026-09-13
kgo: add RewindOffset and LookbackOffset, use ConsumeResetOffset on unlocatable data loss

twmb:
Follow-up to #1416. This folds in #1419, which added an exported lookback offset on top of this branch: the rewind mechanism and the offsets that name it are one change, and shipping either alone leaves the other half broken - a default nobody can tune, or a setter for behaviour that does not exist yet.

### API changes

- `RewindOffset(d)` and `LookbackOffset(d)` are added.
- `ConsumeResetOffset`'s default changes from `NewOffset().AtStart()` to `RewindOffset(time.Minute)`.

### Behavior changes

Against v1.21.7 (#1416 is unreleased):

- After a partition has consumed, an OFFSET_OUT_OF_RANGE the client cannot locate follows `ConsumeResetOffset`. The default rewinds one minute from the last consumed record and reports `ErrDataLoss`; v1.21.7 reset per the option too, but the default `AtStart` re-read the whole log and nothing was reported.
- Below the log start after consuming, the client resumes at the log start regardless of `ConsumeResetOffset`. v1.21.7 applied the option, so `AtEnd` skipped the retained backlog.
- Setting only `ConsumeStartOffset` no longer sets `ConsumeResetOffset`; only the `noReset` flag carries over. A partition that has consumed nothing and gets OFFSET_OUT_OF_RANGE, a consumer resuming a commit that fell below the log start, follows the reset default, the log start. `ConsumeStartOffset(AtEnd())` alone used to skip to the end; `ConsumeResetOffset(NewOffset().AtEnd())` still does.
- Setting only `ConsumeResetOffset` still sets the start offset, unless it is a `RewindOffset`, in which case the start offset stays at the start.
- An OffsetForLeaderEpoch validation answered UNDEFINED_EPOCH_OFFSET after consuming resets per `ConsumeResetOffset` and reports `ErrDataLoss`. v1.21.7 pinned the cursor at -1 so the next fetch failed and reset per the option, reporting nothing. Before consuming, or with `NoResetOffset`, the cursor now stays where it was.
- `AfterMilli` as a reset offset never resumes ahead of where the client was.

### Why timestamps do not rise

A producer stamps a record when you call Produce, before linger, batching, retries, and the network. So a record's timestamp is when the application handed it over, not when it landed, and a log's timestamp order is only as good as the spread in delivery latency across its producers. A producer five seconds further away writes records stamped five seconds behind a local one that wrote later; a retry widens that to however long the retry took. On a multi-region topic that is the normal shape of the log.

The #1416 reset lists by the last consumed record's timestamp. If the records that replaced ours after a loss are stamped older than that, the listing answers ahead of us or not at all, and we keep our offset, skipping them.

### The principle

Policy applies only where no exact answer exists.

| situation | resolution |
|---|---|
| `NoResetOffset()` configured, any OOOR | the error, partition fatal (unchanged) |
| OOOR, nothing consumed yet | `ConsumeResetOffset`, as it always has; a `RewindOffset` resumes at the log start |
| OOOR after consuming, below the log start | the log start (unchanged) |
| OOOR after consuming, past the end or back in range | `ConsumeResetOffset`, plus `ErrDataLoss` |
| epoch validation answered `UNDEFINED_EPOCH_OFFSET` after consuming | same row |
| epoch validation answered `UNDEFINED_EPOCH_OFFSET` before consuming, or under `NoResetOffset` | stay put; the next fetch decides |
| epoch validation located the divergence | the divergence point, `ErrDataLoss` (unchanged) |

Below the start there is nothing to decide: every record left is one we never consumed. The other rows are loss the client cannot locate, and that is where `ConsumeResetOffset` now applies. Java applies `auto.offset.reset` to `UNDEFINED_EPOCH_OFFSET` too, with `latest` as its default; we keep the offset we were validating when there is nothing to look back from, and let the fetch path bound or reset it if it turns out to be out of the log.

### `ConsumeResetOffset` decides on unlocatable loss

The default, `RewindOffset(time.Minute)`, resumes at the first record stamped at or after one minute before the last consumed record's timestamp, never ahead of where we were, never below the start, never past the end. A minute covers the delivery-latency spread on a worldwide topic including a retrying producer, and the price is a minute of topic throughput re-read on a broker bug. `RewindOffset(0)` is exactly what #1416 merged.

`AtStart()` never skips and re-reads the log. `AtEnd()` skips. Relative offsets keep their relative part: `AtEnd().Relative(-100)` resumes a hundred before the end. `NoResetOffset()` still makes the partition fatal on any OOOR.

`ErrDataLoss` is raised on every reset through this row, with `ConsumedTo` the offset we were at and `ResetTo` where we resumed. Its doc now says that when the client could not locate the divergence, records below `ResetTo` may also have been replaced.

### `RewindOffset(d)` and `LookbackOffset(d)`

- `RewindOffset(d)` resumes `d` before the last consumed record. It is only valid as the reset offset; `NewClient` rejects it as a start offset or in `ConsumePartitions`. With nothing consumed there is nothing to rewind from, so it resumes at the log start, and it is not copied into the start offset.
- `LookbackOffset(d)` begins `d` before the newest record in a partition, Kafka's `auto.offset.reset=by_duration`, KIP-1106; the README row for it moves from the not-relevant list to Supported. The client lists the newest record's timestamp with ListOffsets -3 (v7, Kafka 3.0, KIP-734), then lists by that minus `d`. A cluster without v7 answers -3 as an ordinary timestamp; Redpanda advertises v6 and would return its earliest record. So the client only asks when the cluster advertises v7, sends the request under the consumer's isolation level, and ignores an answer negotiated below v7, which a version-capped client or an older broker mid-roll produces; there, and when a broker answers -3 with an error code, it uses the current time minus `d`. As a reset offset it keeps the last `d` of the log after a loss, ahead of where we were if need be.
- Each is a position like `AfterMilli`, held in `Offset.at` next to `atCommitted`, so `At`, `AtStart`, and `AtEnd` overwrite it and nothing needs clearing. `Relative` and `WithEpoch` have no effect on either. Combining a lookback with an exact offset was considered and dropped: it needs a Fetch to learn that record's timestamp, and `AfterMilli(ts-d)` already covers a fixed instant.
- The subtraction is done in milliseconds and floors rather than wraps: a record stamped near the unix epoch would otherwise produce a timestamp of -1 or -2, which a ListOffsets request reads as the end or the start.

### `ConsumeStartOffset` no longer sets where `ConsumeResetOffset` resumes

That copy would make `ConsumeStartOffset(NewOffset().AtStart())`, the most common consumer config, re-read the whole log on every unlocatable loss. Only the `noReset` flag still carries over, because `AtCommitted` documents that it opts into `NoResetOffset` and a group consumer using it as the start offset relies on OOOR being fatal.

The one visible change is a partition that has consumed nothing, which in practice is a group consumer restarting from a commit that fell below the log start. It now follows the reset offset, the log start by default, rather than the start offset, so `ConsumeStartOffset(AtEnd())` alone re-reads the retained backlog where it used to skip it. Skipping is what `ConsumeResetOffset(NewOffset().AtEnd())` is for, and the default skips less.

Setting only `ConsumeResetOffset` still sets the start offset as well, unless it is a `RewindOffset`, which cannot start a partition; the start offset then stays `AtStart`. `ConsumeResetOffset(LookbackOffset(time.Hour))` alone starts new partitions an hour before their newest record and keeps the last hour after a loss.

`offsetLoad.ooorMilli` is renamed `lastConsumedMilli`: a non-zero value says the load is resetting a cursor rather than positioning a fresh one, which `AtStart` and `AtEnd` need even though neither lists a timestamp. `cursorOffset.lastConsumedMilli` floors at 1 so a record stamped at or before the unix epoch does not read as nothing consumed. An `UNDEFINED_EPOCH_OFFSET` validation after consuming routes through the same reset, so the load also carries the epoch we consumed at for the report.

### Tests

kfake, as tables over shared setup (`consumedThenOOR`, `consumeThenPause`, `pollUntil`):

- `TestOutOfRangeReset` covers the unlocatable rows: the default rewind at ten second spacing (six records back) and at two minutes (the record itself), `AtStart`, `AtEnd`, `AtEnd().Relative(-3)`, and a `LookbackOffset` that keeps the tail ahead of where we were. Every row asserts `ErrDataLoss.ConsumedTo`, `ResetTo`, and the offset the records actually resume at. The base client sets only `ConsumeStartOffset(AtStart())`, so the default rows exercise the derived reset.
- `TestOutOfRangeBelowStart` requires the log start whatever the policy says, including `AtEnd`, which would otherwise skip the whole retained backlog.
- `TestOutOfRangeNothingConsumed` covers the fresh partition: the default rewind resumes at the log start, `ConsumeStartOffset(AtEnd())` alone resolves at the start too, and `NoResetOffset` errors whether set as the reset offset or carried from the start offset.
- `TestOutOfRangeInRangeNeverSkipsForward` answers the by-time listing ahead of the cursor, and with nothing at all, and requires the cursor to stay where it was and the loss to be reported from and to that offset.
- `TestUndefinedEpochStays` covers the two epoch-validation gates that stay put: `NoResetOffset` after consuming, and nothing consumed.
- `TestLookbackStartOffset`: records a minute apart over ten minutes, `ConsumeStartOffset(LookbackOffset(5*time.Minute))` requires starting at record 4, five minutes before the newest record; the same offset set only as the reset offset requires the same; with ListOffsets pinned to v6 it requires record 5, the current time minus five minutes.

One kgo unit test in `consumer_test.go`, `TestLookbackMilli`, covers the millisecond floor including a subtraction that wraps.

`TestAuditOFLEUndefinedEpochOffsetNotDataLoss` is renamed `...ResetsInBounds`: it now asserts the loss is reported from a real offset, which is what the original bug was about.

The pre-existing out of range tests move onto the kfake test APIs from #1435. The fetch answered OFFSET_OUT_OF_RANGE is a fault with a `When` on the fetch offset rather than a control that builds the response; the watermarks the control set were never read. `Cluster.DeleteRecords` replaces the hand-built request and `produceBatches` the per-record `ProduceSync` loops. The `OffsetForLeaderEpoch` and by-time `ListOffsets` controls stay: both answer with a response body that is what the test is about.

Java and librdkafka apply `auto.offset.reset` to every OOOR with no regard to what was consumed and default to the end, so a slow consumer that falls below the log start skips the retained backlog in both. Neither has a nearest or lookback reset.

### Changelog

- `kgo`: add `RewindOffset` and `LookbackOffset`, and use `ConsumeResetOffset` when the broker loses data the client cannot locate. The reset default is now `RewindOffset(time.Minute)`, one minute before the last consumed record rather than its timestamp; `AtStart`, `AtEnd`, relative, and `NoResetOffset` resets now apply after consuming as well. Setting only `ConsumeStartOffset` no longer sets `ConsumeResetOffset`: an OFFSET_OUT_OF_RANGE on a partition that has consumed nothing follows the reset default, the log start, rather than the start offset; set `ConsumeResetOffset(NewOffset().AtEnd())` to skip. Every such reset reports `ErrDataLoss`.


pkg/kgo/config.go
```diff
@@ -1625,62 +1623,66 @@ func MaxConcurrentFetches(n int) ConsumerOpt {
...
 //	NewOffset().AtEnd()       == auto.offset.reset "latest"
 //	NewOffset().AtCommitted() == auto.offset.reset "none"
 //
-// Be sure to check the documentation for [ConsumeResetOffset], especially if
-// you rely on this option as the reset offset as well.
+// This option sets the start offset only. [ConsumeResetOffset] decides every
+// OffsetOutOfRange, including one on a partition that has consumed nothing,
+// and setting this option does not change it. One flag does carry over: a
+// start offset that opts into [NoResetOffset], such as [Offset.AtCommitted],
+// makes OffsetOutOfRange fatal as well.
```
twmb: Reword the end here, confusing.

pkg/kgo/config.go
```diff
@@ -1625,62 +1623,66 @@ func MaxConcurrentFetches(n int) ConsumerOpt {
...
-// start if the consumer fell below it, otherwise the first offset at or after
-// the last consumed record's timestamp, never ahead of where the consumer
-// was, and never past the log end. If you want to disable offset resetting
-// entirely, you can use [NoResetOffset].
+// There are three such points: our offset is past the log end, our offset is
+// back within the log after an OffsetOutOfRange, or an OffsetForLeaderEpoch
+// validation replied that it has no record of the epoch we consumed at. Each
+// one also reports an [ErrDataLoss]. If we instead fell below the log start
+// after consuming, we resume at the log start: every record that still exists
+// is one we never consumed, so there is nothing to decide.
```
twmb: There are three times this is used:
* Our offset is past the log end (<how does this happen - via SetOffsets? - isn't this just the next paragraph?>)
* Our offset is back within the log after OffsetOutOfRange, meaning the broker lost data, we detected it, and then a producer produced enough for us to be back in range
* Whatever else
Each of these cases reports an ErrDataLoss. If we were consuming before the log start, the client resets to the log start.

Also the lookback line - very poorly worded and confusing!

pkg/kgo/config.go
```diff
@@ -1625,62 +1623,66 @@ func MaxConcurrentFetches(n int) ConsumerOpt {
...
-// first time, or reset while fetching:
-//
-//	at start?                         => reset to the log start offset
-//	at end?                           => reset to the log end offset
-//	at exact?                         => reset to an exact offset (3 means offset 3)
-//	relative?                         => reset to the above, + / - the relative amount
-//	exact/relative are out of bounds? => reset to the nearest boundary (start or end)
-//	after millisec?                   => reset to the first offset after millisec if one exists, else the log end offset
+// The default rewinds a minute from the last record we consumed and resumes
+// at the first record stamped at or after that, never ahead of where we were
```
twmb: The default is to resume consuming at the first record with a timestamp one minute before our last consumed record. The client cannot guarantee where exactly to begin consuming from to avoid missing any data on data loss. So, a larger lookback reduces your chance of skipping data but increases the amount of re-processing you may do. See Offset.Lookback for more details.

pkg/kgo/config.go
```diff
@@ -1625,62 +1623,66 @@ func MaxConcurrentFetches(n int) ConsumerOpt {
...
+// partition starts at 8 resumes at offset 8.
 //
-// Be sure to check the documentation for [ConsumeStartOffset], especially if
-// you rely on this option as the start offset as well.
+// If you do not set [ConsumeStartOffset], this option's position sets it as
+// well; a lookback stays here, so a partition consumed for the first time
+// starts at that position rather than by duration. The reverse does not
+// hold: setting ConsumeStartOffset leaves this option at its default, and
+// only a start offset that opts into [NoResetOffset] makes OffsetOutOfRange
+// fatal here as well.
```
twmb: Word vomit. Improve. Make clearer.

pkg/kgo/consumer.go
```diff
@@ -106,14 +118,43 @@ func (o Offset) AfterMilli(millisec int64) Offset {
 	o.at = millisec
 	o.relative = 0
 	o.epoch = -1
-	o.afterMilli = true
+	o.afterMilli, o.hasLookback = true, false
+	return o
+}
+
+// Lookback returns an offset that consumes from the first offset at or after
+// a duration before a reference point in time. Unlike [Offset.AfterMilli],
```
twmb: Lookback delays computing the offset to use until it is needed. At that point,
the client uses the first offset at/after `now-duration`.

For start offsets,
AtEnd().Lookback(time.Minute) means you start consuming one minute before
the most recently produced offset. This is equivalent to Kafka's ...

For resetting, NewOffset().Lookback(time.Minute)
means you consume one minute before the most recently consumed offset.

(nix the line about "if no record is that new", as well as the whole "as a reset offset" paragraph since it's poorly worded and confusing, and skip the final line about looking forward clamping)

pkg/kgo/consumer.go
```diff
@@ -172,14 +213,38 @@ func (o Offset) WithEpoch(e int32) Offset {
...
+	return max(o.at, 1)
+}
+
+// lookbackMilli returns the millisecond timestamp d before milli, floored so
+// that a by-time listing never asks for a timestamp a ListOffsets request
+// reserves: -1 is the end, -2 the start, -3 the max timestamp. Lookback clamps
+// d to be non-negative, so only a record carrying a timestamp near the bottom
+// of an int64 can subtract past it, wrapping around to a huge future timestamp
+// that would list the end and skip records. Any target at or below zero
+// precedes every record in the log, so 1 answers the same as all of them.
```
twmb: Honestly just delete this doc. The code is obvious

pkg/kgo/consumer.go
```diff
@@ -172,14 +213,38 @@ func (o Offset) WithEpoch(e int32) Offset {
...
+func lookbackMilli(milli int64, d time.Duration) int64 {
+	back := d.Milliseconds()
+	if milli-back > milli {
```
twmb: Inline comment: // avoid wraparound on massive durations

pkg/kgo/consumer.go
```diff
@@ -1501,11 +1566,12 @@ type offsetLoadMap map[string]map[int32]offsetLoad
...
-	// ooorMilli is non-zero when we are resetting a cursor that received
-	// OFFSET_OUT_OF_RANGE while consuming: the timestamp of the last
-	// record it consumed. The offset itself is the one that was out of
-	// range. See listOffsetsForBrokerLoad.
-	ooorMilli int64
+	// lastConsumedMilli is non-zero when we consumed from this partition
+	// and can no longer trust where we are: a list load resets, an epoch
+	// load resets if the validation cannot answer. The Offset carries the
+	// offset we could not use, and Offset.epoch the epoch we consumed at,
+	// which a list load otherwise ignores. See listOffsetsForBrokerLoad.
```
twmb: just delete this entire comment and add an inline comment on the field, // for handling OffsetOutOfRange

pkg/kgo/source.go
```diff
@@ -204,14 +204,15 @@ type cursorOffset struct {
 	hwm int64
 }
 
-// lastConsumedMilli returns the millisecond of the last record we consumed,
-// or zero if we have consumed nothing. Resetting after an out of range fetch
-// resumes by this millisecond.
+// lastConsumedMilli returns the millisecond of the last record we consumed, or
+// zero if we consumed nothing. A load carries this to say that it is resetting
+// a cursor rather than positioning a fresh one, so a record stamped at or
+// before the unix epoch reports 1 rather than colliding with that meaning.
```
twmb: Undo the rewording here, instead add inline comment below: ``// ensure we lookup a time after 0, not a reserved offset below 0`

pkg/kgo/source.go
```diff
@@ -1407,32 +1408,42 @@ func (s *source) handleReqResp(br *broker, req *fetchRequest, resp *kmsg.FetchRe
...
 						keep = true
-					} else if !c.lastConsumedTime.IsZero() {
-						// We were consuming and the log changed under us, so rather than follow the reset policy
-						// we resume by the last consumed timestamp, bounded within the log and never ahead of
-						// where we were; see listOffsetsForBrokerLoad.
+
+					case c.lastConsumedTime.IsZero():
+						// A reset offset that looks back does so from the last
+						// record we consumed, and we have consumed none, so we
+						// drop the lookback and take the offset's position.
```
twmb: Why aren't we instead looking back from now? i.e., just respecting the user's requested offset?

pkg/kgo/source.go
```diff
@@ -1407,32 +1408,42 @@ func (s *source) handleReqResp(br *broker, req *fetchRequest, resp *kmsg.FetchRe
...
 							)
 						}
-					} else {
+
+					default:
+						// We were consuming and the log changed under us. If we fell below the log start, the
+						// start is exact and we resume there. Otherwise the broker lost data at a point we
+						// cannot determine, and ConsumeResetOffset decides; see listOffsetsForBrokerLoad. The
+						// epoch never reaches the wire on a list load; we carry the epoch we consumed at so
+						// the reset can report it in ErrDataLoss.
```
twmb, line 1439: reformat this... it's really wide.

pkg/kgo/consumer.go
```diff
@@ -2549,28 +2616,49 @@ func (cl *Client) listOffsetsForBrokerLoad(ctx context.Context, broker *broker,
...
+				} else {
+					// The log shrank below us, at a point we cannot determine: the epoch path either
+					// did not run or could not answer. ConsumeResetOffset decides where we resume,
+					// and we never resume ahead of where we were.
+					r := cl.cfg.resetOffset
+					switch {
+					case r.afterMilli:
+						// If the by-time answer is ahead of us, the records between are
+						// stamped older than our last record and we never read them, so
+						// we keep our offset; no answer at all means the same thing.
```
twmb, line 2630: also needs a few words about why we use -1 epoch

pkg/kgo/consumer.go
```diff
@@ -2708,20 +2797,27 @@ func (*Client) loadEpochsForBrokerLoad(ctx context.Context, broker *broker, load
...
-						ooorMilli: loadPart.ooorMilli,
-						Offset:    NewOffset().At(offset),
-					},
-				})
-				continue
+				// the same answer, since a leader's cache only gains the epochs it writes itself. We cannot
+				// locate where the log diverged, so we reset exactly as an out of range fetch after consuming
+				// does; see the addList closure in source.go.
+				//
+				// With nothing consumed there is nothing to look back from, and with NoResetOffset we do not
```
twmb: why is this referring to look back here? Why are we gating on last consumed milli?

pkg/kgo/config.go
twmb: Reworded: this option only sets the start offset; ConsumeResetOffset handles every OffsetOutOfRange; the one exception is a NoResetOffset start offset, which those options document.

pkg/kgo/config.go
twmb: Rewrote as three bullets. Past the log end after consuming is the broker losing its tail; with an epoch to validate against, OffsetForLeaderEpoch catches that first, so that bullet is the no-epoch case, and the doc now says the validation runs first when it can. The lookback line now reads: a lookback has no last consumed record there, so the client uses the log start.

pkg/kgo/config.go
twmb: Took your wording.

pkg/kgo/config.go
twmb: Cut to three sentences: this option also sets the start offset; a lookback is not copied since it only applies after consuming; new partitions start at the log start.

pkg/kgo/consumer.go
twmb: Rewrote per your draft. One correction: as a start offset it is one minute before the current time, not before the most recently produced record; a log last written to hours ago resolves to its end, which is what by_duration does.

pkg/kgo/consumer.go
twmb: Deleted.

pkg/kgo/consumer.go
twmb: Inlined, plus one word on the floor: never a reserved timestamp.

pkg/kgo/consumer.go
twmb: Done.

pkg/kgo/source.go
twmb: Restored the original doc, inline comment added.

pkg/kgo/source.go
twmb: Looking back from now finds nothing in a log older than the lookback: ListOffsets answers -1, the client takes the end, and every record still in the log is skipped. That is the stale-commit case, a consumer restarting after being down longer than retention, which is the most common OffsetOutOfRange there is. The lookback's position is the log start, so dropping the lookback resumes there and skips nothing. The comment now says this. If you would rather resolve against now, it is one line to change and the default would skip on every stale commit.

pkg/kgo/source.go
twmb: Split across lines.

pkg/kgo/consumer.go
twmb: Added: an offset we did not list, our own or one moved by a relative amount, has no epoch to validate against, so it gets -1.

pkg/kgo/consumer.go
twmb: Reworded without the lookback wording: lastConsumedMilli is non-zero only if we consumed something; with nothing consumed there is no loss to reset from, so we keep the offset we were validating, and if it is out of the log the next fetch returns OffsetOutOfRange and the fetch path handles it.

pkg/kgo/config.go
```diff
@@ -1655,62 +1659,70 @@ func MaxConcurrentFetches(n int) ConsumerOpt {
 //	relative?                         => start at the above, + / - the relative amount
 //	exact/relative are out of bounds? => start at the nearest boundary (start or end)
 //	after millisec?                   => start at first offset after millisec if one exists, else log end offset
+//	LookbackOffset(d)?                => start at the first offset at or after the newest record's timestamp minus d, else log end offset
```
twmb: "else log end offset" - ?
also - shorten doc if possible

pkg/kgo/config.go
```diff
@@ -1655,62 +1659,70 @@ func MaxConcurrentFetches(n int) ConsumerOpt {
...
+// there. This option is used in the three cases where the client cannot
+// locate the loss:
 //
-// If you use an exact or relative offsets and the offset ends up out of range,
-// the client chooses the nearest of either the log start offset or the log end
-// offset. For example, using At(3) when the partition starts at 8 results in
-// the partition being consumed from offset 8.
+//   - The broker replied OffsetOutOfRange for an offset past the log end: the
+//     broker lost data. With an epoch to validate against, OffsetForLeaderEpoch
+//     catches this first, so this is the case where the client had none.
```
twmb: "so this is the case where the client had none" what?

pkg/kgo/config.go
```diff
@@ -1655,62 +1659,70 @@ func MaxConcurrentFetches(n int) ConsumerOpt {
...
+//   - OffsetForLeaderEpoch replied that the broker has no record of the epoch
+//     the client consumed at, so the broker cannot say where the log diverged.
 //
-// The following determines the offset for when a partition is seen for the
-// first time, or reset while fetching:
+// Each of these cases reports an [ErrDataLoss]. If instead the client fell
+// below the log start, it resumes at the log start: every record that still
+// exists is one it never consumed.
+//
+// This option also decides an OffsetOutOfRange for a partition that has not
```
twmb: "decides an offsetoutofrange" - no, decides how an OffsetOutOfRange error is handled for a partition that has not consumed anything yet. (period, end of sentence, no "as it always has")

"A RewindOffset has nothing to rewind from there" - ugly wording.  Instead: In this case, a RewindOffset resumes from the log start.

Also explain how this case can happen - how can we receive OOOR when not consuming anything. Brief sentence.

pkg/kgo/config.go
```diff
@@ -1655,62 +1659,70 @@ func MaxConcurrentFetches(n int) ConsumerOpt {
...
-// consuming when you encounter OffsetOutOfRange. It is highly recommended
-// to read the docs for all Offset methods.
+// An exact or relative offset is bounded within the log: At(3) when the
+// partition starts at 8 resumes at offset 8.
 //
-// Be sure to check the documentation for [ConsumeStartOffset], especially if
-// you rely on this option as the start offset as well.
+// If you do not set [ConsumeStartOffset], this option also sets it, unless it
+// is a RewindOffset, which cannot start a partition; the start offset then
+// stays at its default.
```
twmb: reword final line:
For historical compatibility, this option also sets [ConsumeStartOffset] if you do not set it yourself (unless you use RewindOffset, in which case the start offset stays at the start).

pkg/kgo/consumer.go
```diff
@@ -110,6 +121,27 @@ func (o Offset) AfterMilli(millisec int64) Offset {
 	return o
 }
 
+// LookbackOffset returns an offset that begins a duration before the newest
+// record in a partition: the first record stamped at or after that record's
+// timestamp minus d. This is Kafka's auto.offset.reset "by_duration".
+//
+// The newest record's timestamp comes from ListOffsets v7, Kafka 3.0
+// (KIP-734). On brokers without it, including Redpanda, the client uses the
```
twmb: nix ", including Redpanda,"

pkg/kgo/config.go
twmb: Dropped the else clause and shortened: start d before the newest record.

pkg/kgo/config.go
twmb: Reworded: the broker lost data, and the client had no leader epoch to validate against (with one, OffsetForLeaderEpoch catches this first).

pkg/kgo/config.go
twmb: Took your wording, and added how it happens: a committed or requested offset that is no longer in the log, most often a consumer resuming a commit that fell below the log start.

pkg/kgo/config.go
twmb: Took your wording.

pkg/kgo/consumer.go
twmb: Removed.

PR #1432, opened 2026-09-13
kfake: stop a ListOffsets timestamp scan at the segment boundary

twmb:
Follow-up to #1431, for the segment boundary divergence found afterward.

`UnifiedLog.searchOffsetInLocalLog` commits to the first segment whose `largestTimestamp()` reaches the target and returns whatever `findOffsetByTimestamp` finds in it, empty included; it never moves on to a later segment. `largestTimestamp()` is the segment's `maxTimestampSoFar`, which still counts records deleted from below the log start offset because their bytes stay in the file until the whole segment is deleted. So on a real broker, if DeleteRecords removes a segment's only records at or after the target, a lookup at that target answers -1 even when a later segment has a qualifying record. After #1431 kfake answered the later record.

Each `segmentInfo` now carries a file-level `maxTimestamp`: set on append, on compaction rewrite, and on load from the parsed batches; `trimLeft` leaves it alone. `offsetForTimestamp` picks the first segment whose max reaches the target and scans only that segment (binary search on the running max to the first candidate batch, then the broker's batch and record loop). `eachBatchMetaFrom` from #1431 is gone since nothing else needed it.

`TestIssue1422` gains the case: batch 1 is now `{10_060, 10_050}`, a second DeleteRecords removes offset 4 (the 10_060 record), and a query at 10_055 expects -1 with one batch per segment and offset 8 with every batch in one segment, which is also where the broker's answer differs. The restart subtests cover the per-segment max computed on load.

Not changed: ListOffsets -3 still answers from the index, so after DeleteRecords removes the max timestamp batch it answers the max over the remaining batches, where the broker (`readMaxTimestampAndOffsetSoFar`, no log start offset check) can answer an offset below the log start offset. Matching that would mean keeping index entries for deleted batches.

kfake package passes with `-race`, lint is clean, and the kgo integration tests pass against a standalone kfake with `--data-dir` (the CI `integration-test-kfake` persist configuration).


twmb:
Pushed a second commit after measuring the scan's I/O. After DeleteRecords and a restart from snapshot, the index still lists batches entirely below the log start offset, and the scan read and decompressed each one before finding nothing in it: one ListOffsets over a 500 batch segment with 499 deleted did 500 file reads. Those batches cannot hold the answer, so the scan now skips them from the index alone, the way `LogSegment.findOffsetByTimestamp` starts its scan no earlier than the log start offset. The same lookup now does one read. `TestListOffsetsTimestampSkipsDeletedBatches` counts the reads through a wrapped fs and fails with 200 reads without the skip.

twmb:
Third commit: `segmentInfo.maxEarlierTimestamp` is the running max over this and every earlier segment's file-level max, so picking the segment is a binary search again (the second commit had made it a linear scan of `pd.segments`, as the broker does). `pushBatch` sets it on append and `rebuildMaxTimestampMeta` recomputes it whenever segments are dropped or loaded, so it is exact and the lookup needs no forward pass over segments. `TestIssue1422` checks the sequence after producing, after reload, and after the delete that drops segment 0.

Also, `forEachBatchRecord`'s callback now returns whether to keep going, so the lookup and the -3 handler stop decoding at the first matching record instead of collecting the whole batch through `BatchRecords` first.

PR #1435, opened 2026-09-14
kfake: test APIs for faults, topics and groups, and move the tests onto them

twmb:
Adds kfake test APIs for faults, topics and groups, and moves the existing kfake tests onto them.

The point is the test corpus: 34 files change, +1521 / -2392, net **-871 lines**, with the new production code and its tests already in that count.

Rebased on #1437, which moved every group request onto the single cluster loop. Two of these APIs were split in two to work around the per-group goroutines; both are now one call.

### What is new

**`Fault.When` and `Fault.Observe`.** `When` filters a fault by the request itself, for what the selectors cannot reach: the member epoch on a heartbeat, or the topics the heartbeat subscribes to. `Observe` counts matching requests rather than faulting them, so a test that only wants to know how many of a request arrived installs a fault instead of a keep-forever control. Both are checked where a fault is already checked, so neither needs a new place in the pipeline: `check`, which handlers reach through `deny`, and `topLevel`, which answers the entityless requests.

Two bugs in that path are fixed here, because both are reachable only through the API this PR adds:

* An observing fault answers nothing, so it no longer counts towards `anyHit`. ShareFetch asks `anyHit` whether to long-poll, and an observing fault that matched at the group check ended every empty ShareFetch at once instead of waiting out `MaxWait`.
* A parked fetch is one request, not two. `run()` handed a woken request back through the same path that builds its fault check, which threw away what the request had already hit, so an observing fault counted it twice. The check now survives the park, the way a control already survives it through `lastReq`.

**`CreateTopic`, `DeleteTopic`, `DeleteRecords`.** A cluster could seed topics when it started and read them back, but could not create or delete one while running. Each of the three calls the same function its request handler calls, so the two cannot drift: `createTopic` validates the name, the counts and any manual assignment and then creates the topic, `deleteTopic` wakes the topic's watching fetchers and tears the topic down, and `deleteRecords` truncates a partition. The handlers keep their ACL checks, their response building and their per-request batching, so they behave exactly as they did, and their own tests are untouched. The three APIs answer with the error a client would see rather than a hand-written string.

**`GroupInfo`, `WaitGroupStable`, `WaitGroupInfo`, `SetGroupConfigs`.** `GroupInfo` snapshots a group's state, epoch, members and committed offsets for classic, KIP-848 and share groups. Group state lives on the cluster loop after #1437, so the whole snapshot is one read under `admin`. Neither wait holds `admin` while it waits: `run()` is what would have to make the condition true. `WaitGroupInfo` calls `fn` with nil while the group does not exist, so you can wait for a group to appear and you can wait for one to go away; `WaitGroupStable` takes -1 for any member count. `SetGroupConfigs` goes through the same function `IncrementalAlterConfigs` calls, so the two cannot drift either: the handler refreshes the share sweep after writing configs, and a second copy of that write would not.

### What the tests do with them

Call sites the PR adds, net:

| what | sites |
|---|---|
| `newCluster`, replacing `NewCluster` plus its error check plus a deferred `Close` | 165 |
| `waitStable` / `WaitGroupStable` / `WaitGroupInfo`, replacing a describe loop | 44 |
| `groupCommits`, replacing a `kadm.FetchOffsets` read-back | 32 |
| `SetGroupConfigs`, replacing a hand-built `IncrementalAlterConfigs` | 23 |
| `CreateTopic` | 21 |
| faults with `Observe` | 20 |
| faults with `When` | 13 |
| `DeleteRecords` | 10 |
| `produceN`, replacing a throwaway producer in an inline func | 10 |
| `DeleteTopic` | 6 |

And what they replace, net: 53 fewer `kadm` clients (95 to 42), 31 fewer `ControlKey` controls, 29 fewer `kadm.FetchOffsets` read-backs, 6 fewer `DescribeConsumerGroups` polls.

Deleted outright: `waitForStableGroup` and `waitForStableClassicGroup` (both copies of each), `totalAssignedPartitions` (both copies), `setShareAutoOffsetReset` (both copies), `waitShareGroupEmpty`'s describe loop, `describe848`, `deleteRecordsToEnd`, `recreateTopicRaw`, `injectOffsetCommitError`, `newCoverCluster`'s copy of `newCluster` and `newCoverClient`'s copy of `newPlainClient`, and the counting controls behind `countHandshakes` and `observeEndTxns`.

Two rate tests counted requests with a `ControlKey`, a mutex and a slice of `time.Now()`; both count with an observing fault and a `Hits` delta now. The heartbeat pair that answers FENCED_MEMBER_EPOCH or UNKNOWN_MEMBER_ID by epoch is the example `Fault.When`'s own doc gives. A 500ms sleep before a describe is `waitStable`, which is what the sleep was waiting for.

The persist tests moved onto `newCluster` too, 115 sites across `persist_test.go` and `persist_fuzz_test.go`. Those tests close a cluster mid-test and reopen the same `DataDir` to check what survives a restart, and every one of those `Close` calls stays: `Close` is idempotent, so the `t.Cleanup(c.Close)` the helper registers is a no-op at test end.

### What stays on the wire

Anything whose subject is the handler: the `CreateTopics` error tables, `ValidateOnly`, duplicate topics, `NotController`, a manual `ReplicaAssignment`, delete-by-ID, `fault_test.go`'s `DeleteRecords` and `CreatePartitions` cases, `TestDeleteTopicDropsGroupCommits` (it exists to assert that the `DeleteTopics` *handler* drops them), the `ShareGroupDescribe` and `ConsumerGroupDescribe` tests, the kadm commit and fetch-offset round trips, `behavior_test.go`'s `OffsetFetch` controls fabricating STALE_MEMBER_EPOCH and UNSTABLE_OFFSET_COMMIT, and `broker_sweep_test.go`'s ApiVersions-at-v0 control.

### Still on the wire that could move later

Not done here, listed so it is not lost:

* `issues_test.go` counts Metadata requests with two keep-forever controls (around lines 1931 and 2245). An observing fault counts requests, not entities, so the count itself would match, but a fault only counts once the handler reaches a fault check: a Metadata request naming an unknown topic ID answers `UNKNOWN_TOPIC_ID` before any check runs and would go uncounted. Thirteen of those subtests also reset the counter with `Store(0)`, which a `FaultHandle` has no equivalent for. Left alone rather than change what those cache tests measure.
* Six hand-rolled `RecordBatch` builders with a Castagnoli CRC. `kmsg.RecordBatch.AppendTo` writes `Length` and `CRC` straight from the struct fields, so every builder computes both itself. That wants a finalizing append in `kmsg`, not a kfake API.
* `kafka_tests/static_membership_test.go` still builds `CreateTopics` by hand in two places, and the kafka_tests `adm.CommitOffsets` sites stay as they are.

### Verification

* `gofmt -l`: clean.
* `go vet ./...`: clean, `pkg/kfake` and `pkg/kfake/kafka_tests`.
* `go test -count=1 ./...`: pass, both packages.
* `golangci-lint run --timeout=5m ./...`: 0 issues, both packages.
* `go test -race -count=1 -timeout 1500s ./...`: pass twice, 0 failures and 0 races.
* `go test -race -count=200` on `TestIssue1422` and on `TestListOffsetsMaxTimestampFirstRecord`: pass, 0 failures. Both were test bugs fixed in #1436; the rebase keeps those fixes.
* The `CreateTopics`, `DeleteTopics` and `DeleteRecords` handler tests pass untouched.
* `TestFaultObserveLongPoll` covers both fault fixes: it fails without them (the share fetch returns in ~283us instead of waiting out its 2s `MaxWait`) and passes with them, counting the parked share fetch and the parked classic fetch once each.
* Each of the 12 commits builds and vets on its own, via `git rebase --exec`.


pkg/kfake/topic_lifecycle.go
```diff
@@ -0,0 +1,134 @@
...
+// DeleteTopic deletes a topic, and returns once a group no longer answers
+// with the topic's commits.
+//
+// Deleting is two phases: the data and files go under admin, and the group
+// commits are dropped after that, off the run goroutine. Waiting for a group
+// from inside admin would block run(), which is what services adminCh. The
+// request path does not wait at all: it hands the same work to the group
+// goroutines and answers.
+func (c *Cluster) DeleteTopic(topic string) error {
+	var (
```
twmb: This should just craft a DeleteTopics request and send it through the main loop, rather than recreating the topic deletion logic here. Same with CreateTopic above, and DeleteRecords below.

twmb: Agreed, and deferring this deliberately rather than doing it in this PR.

The mechanism is a `clientReq.internal` bool: push a synthetic request into `c.reqCh` with a `clientConn` that has no `net.Conn`, and guard control dispatch, SASL and `faultsFor` on `!creq.internal` so the real handler runs but a test's faults and controls do not intercept setup calls. Faults need no new nil handling, since `check` and `topLevel` both already return nil for a nil `faultCheck`. `allowedACL` gets one internal exemption.

The reason to wait: `DeleteTopic` here promises it returns once a group no longer answers with the topic's commits, and the `DeleteTopics` handler does not provide that. `dropGroupCommits` (groups.go:203) sends to each group's `controlCh` and returns without waiting. Routing through the handler today would drop that guarantee and make every converted call site racy.

The group refactor that moves group handling onto the run loop makes `dropGroupCommits` synchronous by construction, and then a handler-routed `DeleteTopic` keeps its contract for free and the two-phase comment goes away. So: refactor, rebase this on top, then swap all three.

pkg/kfake/group_info.go
```diff
@@ -0,0 +1,246 @@
...
+	for i := range g.Members {
+		n += g.Members[i].NumAssigned()
+	}
+	return n
+}
+
+// GroupInfo returns information about a group if it exists. This covers
+// classic, KIP-848, and share groups. Everything in the returned info is
+// copied, as TopicInfo copies configs: the members, each member's topics and
+// assignment, and the commits.
```
twmb: Delete everything from `Everything in the` onward.

pkg/kfake/group_info.go
```diff
@@ -0,0 +1,246 @@
...
+func (c *Cluster) WaitForGroup(ctx context.Context, group string, fn func(*GroupInfo) bool) (*GroupInfo, error) {
+	for {
+		g := c.GroupInfo(group)
+		if fn(g) {
+			return g, nil
+		}
+		select {
+		case <-ctx.Done():
+			return g, ctx.Err()
+		case <-time.After(10 * time.Millisecond):
```
twmb: 50 or 100ms; 10 is a lot.

pkg/kfake/group_info.go
```diff
@@ -0,0 +1,246 @@
...
+		return g != nil && g.State == groupStable.String() && len(g.Members) == members
+	})
+}
+
+// WaitForGroup waits until fn returns true for the group and returns it. fn
+// is called with nil while the group does not exist, so you can wait for one
+// to appear.
+//
+// Neither wait holds admin while it waits: run() is what would have to make
+// the condition true.
```
twmb: Delete this block of lines; also this is referring to internal logic.

pkg/kfake/group_info.go
```diff
@@ -0,0 +1,246 @@
...
+// of members and returns it.
+func (c *Cluster) WaitForGroupStable(ctx context.Context, group string, members int) (*GroupInfo, error) {
+	return c.WaitForGroup(ctx, group, func(g *GroupInfo) bool {
+		return g != nil && g.State == groupStable.String() && len(g.Members) == members
+	})
+}
+
+// WaitForGroup waits until fn returns true for the group and returns it. fn
+// is called with nil while the group does not exist, so you can wait for one
+// to appear.
```
twmb: fn should not be called with nil. We should wait ourselves. Is there benefit to letting the callback see a nil group?

pkg/kfake/group_info.go
```diff
@@ -0,0 +1,246 @@
...
+
+// WaitForGroupStable waits for the group to be stable with the given number
+// of members and returns it.
+func (c *Cluster) WaitForGroupStable(ctx context.Context, group string, members int) (*GroupInfo, error) {
+	return c.WaitForGroup(ctx, group, func(g *GroupInfo) bool {
+		return g != nil && g.State == groupStable.String() && len(g.Members) == members
+	})
+}
+
+// WaitForGroup waits until fn returns true for the group and returns it. fn
```
twmb: Maybe WaitGroupInfo? since it's waiting on a specific condition from GroupInfo?

pkg/kfake/group_info.go
```diff
@@ -0,0 +1,246 @@
...
+	return out
+}
+
+// Members come out of a map, so we sort them: a test that indexes Members
+// wants the same member every run.
+func sortMembers(ms []GroupMember) {
+	slices.SortFunc(ms, func(l, r GroupMember) int { return cmp.Compare(l.MemberID, r.MemberID) })
+}
+
+// WaitForGroupStable waits for the group to be stable with the given number
```
twmb: I'm still not sure this function really carries its weight. If we don't allow nil groupinfo to be passed, callsites could just look like

```
c.WaitGroupInfo(ctx, "mygroup", func(i *kfake.GroupInfo) bool { return i.State == "stable" })
```
I doubt most tests really care about member count? maybe....

pkg/kfake/group_info.go
```diff
@@ -0,0 +1,246 @@
...
+// SetGroupConfigs sets group level configs, as an IncrementalAlterConfigs
+// request would.
+//
+// A group config exists independently of the group, here and on a real
+// cluster: setting one does not create a group, and the group picks it up if
+// it is created later, because nothing caches a config at group creation.
+// Setting one on a live group works for the same reason, with the caveat that
+// a share group chooses a partition's start offset once, when the partition is
+// first initialized, so share.auto.offset.reset only reaches partitions that
+// have not started yet.
```
twmb: Reword to:
```
// Group configs exist independently from a group (same as real Kafka):
// setting configs does not create the group; the group picks them up if
// it is created later. Setting configs on a live group works, though some
// group configs are applied at group creation time and would be missed.
// Notably, share.auto.offset.reset only applies to partitions that have
// not yet been read.
```

pkg/kfake/group_info.go
```diff
@@ -0,0 +1,246 @@
...
+		}
+		select {
+		case <-ctx.Done():
+			return g, ctx.Err()
+		case <-time.After(10 * time.Millisecond):
+		}
+	}
+}
+
+// SetGroupConfigs sets group level configs, as an IncrementalAlterConfigs
```
twmb: // SetGroupConfigs sets group level configs, the same way IncrementalAlterConfigs does.

pkg/kfake/topic_lifecycle.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/twmb/franz-go/pkg/kerr"
+	"github.com/twmb/franz-go/pkg/kmsg"
+)
+
+// These run their work through admin, as TopicInfo and MoveTopicPartition
+// do, so that c.data is only touched on the run goroutine. Each calls the
+// same function the matching request handler calls, so the two cannot
+// drift, and each answers with the error a client would have seen. A test
+// checking a handler's own error codes, its ValidateOnly, or a manual
+// replica assignment still sends the request.
```
twmb: Useless comment delete

pkg/kfake/topic_lifecycle.go
```diff
@@ -0,0 +1,83 @@
...
+
+// CreateTopic creates a topic with the given configs, nil for none. A topic
+// deleted and created again under the same name gets a new topic ID, as on a
+// real cluster.
+func (c *Cluster) CreateTopic(topic string, partitions int32, configs map[string]string) error {
```
twmb: Delete "as on a real cluster" from all docs.

pkg/kfake/topic_lifecycle.go
```diff
@@ -0,0 +1,83 @@
...
+		c.persistTopicsState()
+	})
+	if e != nil {
+		return e
+	}
+	return nil
+}
+
+// DeleteTopic deletes a topic. The topic's commits are dropped from every
+// group before we return.
```
twmb: Delete the line about topic commits. Or, reword to: Same as Kafka, the topic's commits are dropped from every group before the delete finishes.

pkg/kfake/faults.go
```diff
@@ -48,6 +48,25 @@ type Fault struct {
...
+	//
+	// When runs on the cluster goroutine. It must not block and must not
+	// call back into the cluster. Returning false means the request is not
+	// faulted; you can use this for observing.
+	When func(kmsg.Request) bool
+
+	// Observe counts matching requests rather than faulting them: the
+	// request proceeds as it would have while the fault is counted for
+	// Hits and Wait. Count applies as it does to any fault, so counting
+	// every request needs -1.
```
twmb: Delete the line about "Count applies as it does to any fault", we shouldn't need to reiterate how count works, especially if it's not changing anything, in a different field.

pkg/kfake/topic_lifecycle.go
twmb: Done in `kfake: share the topic mutations between handler and test API`: `createTopic`, `deleteTopic` and `deleteRecords` are each one function that both the request handler and the test API call, so there is no second copy of the logic here any more, and the handlers keep their ACL checks, response building and per-request batching, with their own tests passing untouched.

The synthetic-request routing described above stays deferred; this gets the no-drift half of it without the `clientReq.internal` plumbing.

pkg/kfake/group_info.go
twmb: Cut from `Everything in the` onward.

pkg/kfake/group_info.go
twmb: Now 50ms.

pkg/kfake/group_info.go
twmb: Deleted.

pkg/kfake/group_info.go
twmb: Kept, per your follow-up. Three of the four call sites use nil to wait for a group to be gone or empty (`api_test.go`, `behavior_test.go`, `rebalance_churn_test.go`), so waiting internally would make that inexpressible. The doc now says plainly that fn is called with nil while the group does not exist, so you can wait for one to appear or to go away.

pkg/kfake/group_info.go
twmb: Renamed to `WaitGroupInfo`, and `WaitForGroupStable` to `WaitGroupStable` so the pair reads together.

pkg/kfake/group_info.go
twmb: Kept. Of 32 call sites, 20 want 1 member, 9 want 2 and 3 want 3, so a third do care about the count and the rest still assert "one member, stable". Added `-1` for any member count, per your follow-up.

pkg/kfake/group_info.go
twmb: Applied verbatim.

pkg/kfake/group_info.go
twmb: Applied verbatim.

pkg/kfake/topic_lifecycle.go
twmb: Deleted.

pkg/kfake/topic_lifecycle.go
twmb: Deleted from every doc in the file.

pkg/kfake/topic_lifecycle.go
twmb: Took the reword; the guarantee is real now that the handler's `dropGroupCommits` is synchronous.

pkg/kfake/faults.go
twmb: Deleted.

PR #1445, opened 2026-09-15
Kafka 4.4 support

twmb:
Behavior half of Kafka 4.4 support; the protocol half was #1446, released as kmsg v1.14.0, which the first commit here depends on. Verified against the 4.4.0-rc1 broker and controller (image built from Kafka's own Dockerfile with the staged RC tarball; ApiVersions captured from both listeners and diffed against 4.3.1). The staged rc2 source is byte-identical to rc1 on every protocol file.

## CI

The first commit teaches both workflows to use a staged release candidate while its version is newer than the newest release tag: "validate DSL against Kafka JSON" validates the RC's source tarball (green against 4.4.0-rc2 now, instead of after the 4.4.0 tag), and the Kafka integration job gains an "integration test kafka 4.4.0-rc2" matrix entry built from Kafka's own Dockerfile and the staged binary tarball. Detection reads https://dist.apache.org/repos/dist/dev/kafka/; RC tags reach GitHub late (rc2 is still untagged there). Both fall back to the release on their own once 4.4.0 is tagged, and if dist.apache.org is unreachable only the release runs.

Locally: every commit builds and vets against the published kmsg without a workspace; the full kfake package passes plain and with `-race`; the new tests pass `-race -count=30`; lint.sh is clean. The 8-minute kgo integration suite was not run here.

## Protocol (merged in #1446)

| key | change | KIP |
|---|---|---|
| 18 ApiVersions | v5: ClusterID and NodeID; REBOOTSTRAP_REQUIRED on a mismatch | KIP-1242 |
| 28 TxnOffsetCommit | v6: TopicID replaces Topic; GROUP_ID_NOT_FOUND, STALE_MEMBER_EPOCH, UNKNOWN_TOPIC_ID | KIP-1319 |
| 42 DeleteGroups | v3: per-group ErrorMessage | KIP-1331 |
| 88, 89, 93 streams group | v1 heartbeat and describe (plus AssignorName, KIP-1357), new StreamsGroupTopologyDescriptionUpdate | KIP-1331 |
| 94 UnregisterController | new, broker and controller listeners | KAFKA-20395 |

Errors 134 GROUP_DELETION_FAILED, 135 STREAMS_TOPOLOGY_DESCRIPTION_UPDATE_FAILED, 136 CONTROLLER_ID_NOT_REGISTERED. No min version changed. The DSL validator passes against both the 4.3.1 and 4.4.0-rc1 message JSON.

## Behavior

- **KIP-1242 (kgo)**: every connection to a broker learned from metadata now sends the cluster ID and node ID we expect; seed connections send neither. A broker answering REBOOTSTRAP_REQUIRED means the address we hold for that node is some other broker or cluster: kgo drops all discovered brokers, refreshes metadata from the seeds (calling `OnRebootstrapRequired` first if set), and the request that opened the connection retries. A stopped broker no longer dials. Brokers before 4.4 answer v5 with UNSUPPORTED_VERSION and their max and we retry at that version, one extra round trip per new connection to them, the same as when v4 was adopted. To opt out, cap ApiVersions at 4 with `MaxVersions`.
- **KIP-1319 (kgo)**: TxnOffsetCommit fills topic IDs like OffsetCommit v10, pins to v5 if a `PreTxnCommitFnContext` fn adds an id-less topic, resolves names on v6 responses, treats GROUP_ID_NOT_FOUND as abortable, and commits under the ID the offsets were fetched under (so a recreated topic is refused by the broker rather than committed to by name).
- **OffsetCommit v10 fix (kgo)**: the v9 pin was decided before the `PreCommitFnContext` fn ran, so a fn-added id-less topic went out at v10 with a zero id, committing to no topic. It is now decided from the final request.
- **kfake**: ApiVersions v5 checks (INVALID_REQUEST for one field only, REBOOTSTRAP_REQUIRED for a mismatch), a KIP-511 reply for a version above what the cluster serves (only the ApiVersions range, so a client retries and still learns features; before, a capped cluster accepted any version), TxnOffsetCommit v6 id resolution and the v6 error mapping, DeleteGroups v3.
- **kadm**: DeleteGroups surfaces the per-group error message.
- **kversion**: `V4_4_0`, `Stable` is 4.4.

## Raw requests

A raw `TxnOffsetCommitRequest` issued by name now goes out at v6, where the name is not on the wire; set `TopicID` as raw `OffsetCommitRequest`s have needed since v10. kgo's own commit paths fill it. The existing kfake tests that did this by name were updated.

## Not in this PR

- KIP-1191 (share group dead letter queues), KIP-1323, KIP-1276 (CIDR ACL hosts), KIP-1170: broker side, no client change. KIP-909: kgo already resolves seeds at dial time.
- Starting ApiVersions from the version cached from a broker's previous connection would avoid the extra round trip on pre-4.4 brokers; left for a follow-up.

## Changelog

- kgo now supports Kafka 4.4: KIP-1242 misrouted connection detection (ApiVersions v5 with the expected cluster and node; a REBOOTSTRAP_REQUIRED reply rebootstraps from the seeds), KIP-1319 TxnOffsetCommit v6 by topic ID, and the KIP-1331 and KIP-1357 streams group protocol changes (proto only). `kversion.Stable` is now 4.4.
- kgo: an OffsetCommit topic added without a TopicID by a `PreCommitFnContext` fn was sent at v10 with a zero id, committing to no topic; the request now pins to v9.
- kadm: DeleteGroups results carry the per-group error message (DeleteGroups v3).
- kfake: ApiVersions v5, TxnOffsetCommit v6, DeleteGroups v3; a request version above what the cluster serves is answered with UNSUPPORTED_VERSION per KIP-511.


pkg/kgo/client.go
```diff
@@ -1187,6 +1188,57 @@ func (cl *Client) updateMetadataBrokers(resp *kmsg.MetadataResponse) {
...
+		return
+	}
+	cl.cfg.logger.Log(LogLevelWarn, "broker rejected our connection with REBOOTSTRAP_REQUIRED: the address we have for this node belongs to a different broker or cluster; dropping all discovered brokers and rebootstrapping from the seeds",
+		"broker", logID(b.meta.NodeID),
+		"addr", b.addr,
+		"cluster_id", clusterID,
+	)
+	if fn := cl.cfg.onRebootstrapRequired; fn != nil {
+		seeds, err := fn()
+		if err != nil {
```
twmb: || len(seeds) == 0

pkg/kgo/broker.go
```diff
@@ -1008,6 +1015,17 @@ start:
 	req.Version = maxVersion
 	req.ClientSoftwareName = cxn.cl.cfg.softwareName
 	req.ClientSoftwareVersion = cxn.cl.cfg.softwareVersion
+	// KIP-1242: tell a discovered broker which broker of which cluster we
+	// expect it to be, so that it can reject a connection that DNS or a
+	// load balancer routed elsewhere. Seeds have no node ID, and before
+	// the first metadata response we have no cluster ID; both fields stay
+	// unset then and the broker skips the check.
+	if maxVersion >= 5 && cxn.b.meta.NodeID >= 0 {
```
twmb: We should also only do this if we have the onrebootstrap cfg, otherwise we wont be able to do anything with the error code.

pkg/kgo/client.go
```diff
@@ -1187,6 +1188,57 @@ func (cl *Client) updateMetadataBrokers(resp *kmsg.MetadataResponse) {
...
+// loadClusterID returns a copy of the cluster ID from the last metadata
+// response, or nil if no response has carried one.
+func (cl *Client) loadClusterID() *string {
+	cl.controllerIDMu.Lock()
+	defer cl.controllerIDMu.Unlock()
+	if cl.clusterID == nil {
+		return nil
+	}
+	s := *cl.clusterID
+	return &s
```
twmb: Should clusterid be an atomic.Pointer[string]?

pkg/kgo/broker.go
```diff
@@ -1008,6 +1015,17 @@ start:
 	req.Version = maxVersion
 	req.ClientSoftwareName = cxn.cl.cfg.softwareName
 	req.ClientSoftwareVersion = cxn.cl.cfg.softwareVersion
+	// KIP-1242: tell a discovered broker which broker of which cluster we
+	// expect it to be, so that it can reject a connection that DNS or a
+	// load balancer routed elsewhere. Seeds have no node ID, and before
+	// the first metadata response we have no cluster ID; both fields stay
+	// unset then and the broker skips the check.
```
twmb: Shorten this doc. It doesn't need so much exposition.

pkg/kgo/client.go
```diff
@@ -1187,6 +1188,57 @@ func (cl *Client) updateMetadataBrokers(resp *kmsg.MetadataResponse) {
...
+}
+
+// rebootstrapMisrouted is called when broker b answered REBOOTSTRAP_REQUIRED
+// to an ApiVersions request that named the cluster and node we expected
+// (KIP-1242): the address our metadata has for b belongs to a different
+// broker or cluster, so every discovered broker is suspect. We drop them all
+// so that the next metadata request goes to a seed, after giving
+// OnRebootstrapRequired a chance to replace the seeds, and refresh metadata
+// now. Requests in flight to discovered brokers fail with errChosenBrokerDead
+// and retry once metadata rediscovers the cluster.
```
twmb, line 1199: Do we really need to drop all brokers, vs just the one that is known misrouted? What does the Java client do?

pkg/kgo/config.go
```diff
@@ -1060,6 +1060,14 @@ func ConsiderMissingTopicDeletedAfter(t time.Duration) Opt {
 // call UpdateSeedBrokers with the seeds you return. All other live connections
 // to brokers are stopped and active requests are failed.
 //
+// The same function is called when a broker answers REBOOTSTRAP_REQUIRED to
+// the ApiVersions request that opens a connection (KIP-1242, Kafka 4.4). The
```
twmb: drop "that opens a connection".

pkg/kgo/config.go
```diff
@@ -1060,6 +1060,14 @@ func ConsiderMissingTopicDeletedAfter(t time.Duration) Opt {
...
+// client names the cluster and broker it expects on every connection to a
```
twmb: change "on every connection to a broker it learned from metadata" "on every connection,"

pkg/kgo/config.go
```diff
@@ -1060,6 +1060,14 @@ func ConsiderMissingTopicDeletedAfter(t time.Duration) Opt {
...
+// broker it learned from metadata, and a broker that is not that node answers
```
twmb: "is not the node the client says it should be answers REBOOTSTRAP_REQUIRED."

pkg/kgo/txn.go
```diff
@@ -231,9 +231,10 @@ func (s *GroupTransactSession) failed() bool {
 // No returned error is retryable. Either the transactional ID has entered a
 // failed state, or the client retried so much that the retry limit was hit,
 // and odds are you should not continue. The exception is a transaction that
-// produced to a topic which was then deleted and recreated: this aborts and
-// returns an error wrapping kerr.UnknownTopicID, and you can continue once
-// you purge the topic and add it back; see EndTransaction. While a context is
+// produced to or consumed from a topic which was then deleted and recreated:
+// this aborts and returns an error wrapping kerr.UnknownTopicID, and you can
+// continue once you purge the topic and add it back; see EndTransaction.
+// While a context is
```
twmb: fix comment wrapping here.

pkg/kgo/txn.go
```diff
@@ -1360,12 +1375,57 @@ func (g *groupConsumer) commitTxn(ctx context.Context, tx890p2 bool, req *kmsg.T
...
 			onDone(req, nil, err)
 			return
 		}
+
+		// v6 responses carry TopicID with no topic name. Resolve names
+		// from the request (which always carries both) so the response
+		// handed to onDone -- and the per-partition error messages End
+		// builds from it -- name the topics.
+		if resp.Version >= 6 {
+			id2t := make(map[[16]byte]string, len(req.Topics))
```
twmb: Can't we use the client's id2t? and only fallback here if needed?

pkg/kgo/client.go
twmb: Done: an empty seed list is logged and keeps the current seeds, the same as an error.

pkg/kgo/client.go
twmb: Done: clusterID is an atomic.Pointer[string]. The store still clones the response's string, RequestCachedMetadata loads it outside controllerIDMu, and ApiVersions reads it directly, so loadClusterID is gone.

pkg/kgo/broker.go
twmb: Done, two lines.

pkg/kgo/client.go
twmb: Java drops everything: on REBOOTSTRAP_REQUIRED from ApiVersions, NetworkClient closes every node it knows and calls the metadata updater's rebootstrap, which replaces the cluster with the bootstrap addresses. Same here. Dropping only the one broker would refresh metadata from a sibling that came from the same metadata response as the bad address; a wrong cluster id in particular means the whole map came from elsewhere. Kept as is unless you want the lighter version.

pkg/kgo/config.go
twmb: Done.

pkg/kgo/config.go
twmb: Done.

pkg/kgo/config.go
twmb: Done.

pkg/kgo/txn.go
twmb: Done.

pkg/kgo/txn.go
twmb: Done: names come from the client's id2t first, and the request only fills what id2t lacks (a topic the fn added, or the old id of a recreated topic, since id2t follows the new id). updateCommitted still builds its map from the request for OffsetCommit v10; want that changed the same way?

PR #1447, opened 2026-09-15
ci: test against a Kafka release candidate when one is staged

twmb:
Both workflows look at https://dist.apache.org/repos/dist/dev/kafka/ for the newest staged release candidate and use it while its version is newer than the newest release tag on GitHub:

- **validate DSL against Kafka JSON**: validates the RC's source tarball instead of the release tag. Today that is 4.4.0-rc2, so this goes green for the 4.4 definitions now instead of after the 4.4.0 tag. RC tags reach GitHub late (rc2 is still untagged), so the tarball is the source.
- **integration test kafka**: a second matrix entry, "integration test kafka 4.4.0-rc2", builds the image from Kafka's `docker/jvm` Dockerfile and the staged binary tarball (about three minutes; the build-jsa stage boots Kafka once) and runs the same compose and tests. The existing entry keeps its name, so required checks are unchanged.

Once 4.4.0 is released the RC is no longer newer than a release and both fall back to the release without a change here. If dist.apache.org is unreachable, detection yields no RC and only the release runs.

This PR's own checks are the first run of both paths.


twmb:
Folded into #1445 as its first commit.

PR #1461, opened 2026-09-17
kgo: add MaxDecompressedBatchBytes

twmb:
Fetch limits bound only the compressed bytes on the wire, and a batch is decompressed into one contiguous allocation. The decompressor already capped a batch at math.MaxInt32; this makes the cap an option, applies it to the batches StreamingCompression merges, and changes what happens when a batch is over it.

Before, a batch over the cap was kept as a fetch error with no offset advance: the next fetch re-requested the same offset and failed the same way, every fetch. Now the cursor goes fatal. PollFetches returns an `ErrDecompressTooLarge` once, naming the batch and the offset after it, an error log says the same, and the partition is not fetched again until `SetOffsets` moves it or the partition is unassigned.

The bound is a property of the decompressor, fixed when it is built: the client builds its default decompressor with the option, and a default decompressor the user built and passed to `WithDecompressor` is rebuilt with it. Every codec rejects before allocating where the format declares its output size (zstd frame content size, snappy decoded length, xerial chunk sums) and at the bound otherwise (gzip and lz4 through a LimitedReader). The client does not bound any other decompressor; one that rejects a batch with an error wrapping the new `ErrMaxDecompressed` stops the partition the same way.

StreamingCompression merges batches up to a cap on their uncompressed size, so that a consumer can decompress a merged batch. The cap was a 256 MiB constant. It is now min(1 GiB, MaxDecompressedBatchBytes).

Related: apache/kafka@4d781f5 bounds per-record decompressed size on the broker.

## Changelog

- `kgo`: new `MaxDecompressedBatchBytes` option bounds how large a fetched batch may decompress to and caps the uncompressed size of batches `StreamingCompression` merges. A batch over the bound stops the partition: `PollFetches` returns `ErrDecompressTooLarge` and the partition resumes after `SetOffsets`. A decompressor supplied with `WithDecompressor` opts in by returning an error wrapping `ErrMaxDecompressed`.


pkg/kgo/client.go
```diff
@@ -182,8 +182,14 @@ func validateCfg(opts ...Opt) (cfg, []hostport, error) {
 			return cfg, nil, err
 		}
 	}
-	if cfg.decompressor == nil {
-		cfg.decompressor = DefaultDecompressor(cfg.pools...)
+	// Our own default is bounded at MaxDecompressedBatchBytes. A default
+	// decompressor the user built is bounded at math.MaxInt32, which the
+	// zstd decoder honors while decoding; rebuild it with our bound.
```
twmb: "is bounded at math.MaxInt32; rebuild it with our bound". No reason to mention zstd here.

pkg/kgo/compression.go
```diff
@@ -18,23 +19,11 @@ import (
...
-// so a batch claiming more is corrupt or hostile and is rejected like any
-// other corrupt batch: a loud, repeated fetch error with no offset advance.
-// A var only so tests can shrink it.
-var maxDecompressedSize = int64(math.MaxInt32)
-
-var errDecompressedTooLarge = errors.New("decompressed data exceeds the maximum allowed decompressed batch size (corrupt or malicious batch)")
+// ErrMaxDecompressed is returned, wrapped, by [DefaultDecompressor] and
+// should be returned, wrapped, by any [BoundedDecompressor] when a batch
+// would decompress to more than the bound. The client treats this error as
+// fatal for the partition; see [MaxDecompressedBatchBytes].
```
twmb: ErrMaxDecompressed is returned when a batch we consumed would decompress larger than [MaxDecompressedBytes]. The client treats this error as fatal for the partition and it can only be recovered via SetOffsets or by you restarting your client with a higher limit.

Then, update error wording: "decompressed data exceeds MaxDecompressedBatchBytes", nothing more. We could also make this a typed error so that we can add the number to the option?

pkg/kgo/compression.go
```diff
@@ -136,6 +125,19 @@ type Decompressor interface {
 	Decompress(src []byte, codecType CompressionCodecType) ([]byte, error)
 }
 
+// BoundedDecompressor is an optional interface a [Decompressor] can
+// implement to honor [MaxDecompressedBatchBytes]. If your decompressor does
+// not implement this interface, the option does not apply to it. The
+// [DefaultDecompressor] implements this interface.
```
twmb: Reword final sentence to "All decompressors provided by kgo implement this interface."

pkg/kgo/compression.go
```diff
@@ -136,6 +125,19 @@ type Decompressor interface {
...
+type BoundedDecompressor interface {
+	Decompressor
+
+	// DecompressBounded is Decompress with a bound: if src would
+	// decompress to more than maxBytes, this returns an error wrapping
+	// [ErrMaxDecompressed], ideally before allocating the output.
```
twmb: What do you mean ideally? It needs to be guaranteed. Do not allocate more. Drop everything after the comma. If this is referring to zstd where we can only control individual frames, clarify that here: "The zstd decompressor can control only the size of individual decompressed frames. This option ensures no frame can exceed the maxBytes, and then has a second client-side step that errors as soon as an additional batch's frame exceeds maxBytes." Actually I think my wording here is wrong? What if a frame declares a small size but then decompresses huge?

pkg/kgo/compression.go
```diff
@@ -439,13 +441,29 @@ type decompressor struct {
...
 
 // DefaultDecompressor returns the default decompressor used by clients.
 // The first pool provided that implements PoolDecompressBytes will be
 // used where possible.
+//
+// The decompressor bounds a batch at math.MaxInt32 decompressed, which no
+// legitimate batch can exceed: every known producer serializes a batch's
+// records into an int32-indexed buffer before compressing. A client bounds
+// its own default decompressor at [MaxDecompressedBatchBytes] instead. The
+// decompressor also implements [BoundedDecompressor].
```
twmb: Drop this paragraph.

pkg/kgo/compression.go
```diff
@@ -439,13 +441,29 @@ type decompressor struct {
...
 func DefaultDecompressor(pools ...Pool) Decompressor {
+	return newDecompressor(math.MaxInt32, pools...)
+}
+
+// newDecompressor returns a decompressor bounded at max. The zstd decoder
+// is built with max as its ceiling, so a zstd frame that omits its content
+// size is rejected once the decoder has produced max bytes; DecompressBounded
+// with a smaller bound can only reject such a frame after decoding.
```
twmb: What's this final line mean?

pkg/kgo/compression.go
```diff
@@ -573,52 +598,64 @@ func (d *decompressor) Decompress(src []byte, codecType CompressionCodecType) (_
...
 	return s2.Decode(dst, src)
 }
 
-func (d *decompressor) decompressZstd(dst, src []byte) ([]byte, error) {
+// decompressZstd rejects a frame that declares a content size over max
+// before decoding. A frame without one is bounded by the decoder's ceiling
+// (d.max) while decoding and checked against max after.
+func (d *decompressor) decompressZstd(dst, src []byte, max int) ([]byte, error) {
+	var h zstd.Header
+	if h.Decode(src) == nil && h.HasFCS && h.FrameContentSize > uint64(max) {
```
twmb: What's the purpose of this? The header can be lying can't it? Does this only control one frame?

pkg/kgo/config.go
```diff
@@ -982,6 +992,31 @@ func BrokerMaxReadBytes(v int32) Opt {
...
+// [SetOffsets] past the batch. Use a bound comfortably above the largest
+// batch your producers write; a compressed batch is at most the broker's
+// max.message.bytes, and a batch usually decompresses to a few times that.
+//
+// The bound applies to the [DefaultDecompressor] and to any decompressor you
+// pass to [WithDecompressor] that implements [BoundedDecompressor]; other
+// decompressors are not bounded.
+//
+// The value must be at least [ProducerBatchMaxBytes], which bounds the
+// client's own batches before compression.
```
twmb: This is not the wording we agreed to.

pkg/kgo/config_test.go
```diff
@@ -134,3 +134,14 @@ func TestConfigRejectsInt32MillisOverflow(t *testing.T) {
 		}
 	}
 }
+
+func TestConfigMaxDecompressedBatchBytes(t *testing.T) {
```
twmb: Useless test, delete.

pkg/kgo/source.go
```diff
@@ -1310,6 +1323,22 @@ func (s *source) handleReqResp(br *broker, req *fetchRequest, resp *kmsg.FetchRe
...
+				// unusable after the fetch is drained.
+				if tooLarge, ok := errors.AsType[*ErrDecompressedTooLarge](fp.Err); ok {
+					partOffset.from.fatal.Store(true)
+					s.cl.cfg.logger.Log(LogLevelError, "batch decompresses larger than MaxDecompressedBatchBytes, stopping consuming the partition; skip the batch with SetOffsets",
+						"broker", logID(s.nodeID),
+						"topic", topic,
+						"partition", partition,
+						"offset", tooLarge.Offset,
+						"max_bytes", tooLarge.MaxBytes,
+						"set_offsets", fmt.Sprintf("map[string]map[int32]kgo.EpochOffset{%q: {%d: {Epoch: %d, Offset: %d}}}", topic, partition, tooLarge.Epoch, tooLarge.NextOffset),
```
twmb: nix this log line, bit ugly tbh.

pkg/kgo/source.go
```diff
@@ -1600,6 +1630,11 @@ type ProcessFetchPartitionOpts struct {
 	// Pools contain potential pools to use for memory pooling.
 	Pools []Pool
 
+	// MaxDecompressedBytes bounds how large a batch may decompress to, if
+	// the decompressor implements [BoundedDecompressor]. This field is the
+	// same as [MaxDecompressedBatchBytes]; zero means math.MaxInt32.
```
twmb: Delete "This field is the same as MaxDecompressedBatchBytes". At most, change to "See MaxDecompressedBatchBytes" (similar to the isolationlevel doc above)

pkg/kgo/errors.go
```diff
@@ -429,6 +429,37 @@ func (e *ErrGroupSession) Error() string {
...
+// consuming the partition: it is not fetched again until you [SetOffsets]
+// it past the batch, to NextOffset. Alternatively, create a new client with
+// a larger bound.
+type ErrDecompressedTooLarge struct {
+	// Topic is the topic the batch is in.
+	Topic string
+	// Partition is the partition the batch is in.
+	Partition int32
+	// Offset is the batch's first offset, or the wrapper message's offset
+	// for a legacy (v0 or v1) message set.
```
twmb: skip the "or the wrapper" bit; almost nobody uses message sets.

pkg/kgo/errors.go
```diff
@@ -429,6 +429,37 @@ func (e *ErrGroupSession) Error() string {
...
+	Offset int64
+	// NextOffset is the offset after the batch's last record, i.e., where
+	// to SetOffsets to skip the batch.
+	NextOffset int64
+	// Epoch is the batch's partition leader epoch, or -1 for a legacy
+	// message set.
```
twmb: Is this used in SetOffsets or not? Is this the epoch of the failed record? I think so. I think we need to note: This is the epoch of the record that failed, it should not be used in SetOffsets because the client does not know the epoch of the next offset.

Also this should go before NextOffset in the field order, since it belongs to the offset above.

pkg/kgo/errors.go
```diff
@@ -429,6 +429,37 @@ func (e *ErrGroupSession) Error() string {
...
+type ErrDecompressedTooLarge struct {
+	// Topic is the topic the batch is in.
+	Topic string
+	// Partition is the partition the batch is in.
+	Partition int32
+	// Offset is the batch's first offset, or the wrapper message's offset
+	// for a legacy (v0 or v1) message set.
+	Offset int64
+	// NextOffset is the offset after the batch's last record, i.e., where
+	// to SetOffsets to skip the batch.
```
twmb, line 448: where tto SetOffsets (with epoch -1) to skip the batch.

pkg/kgo/client.go
twmb: Done.

pkg/kgo/compression.go
twmb: Done, doc and error text as written. On a typed error: ErrDecompressedTooLarge already carries MaxBytes (and topic, partition, offsets) and wraps this sentinel, so PollFetches callers get the number. I kept the sentinel untyped so a custom BoundedDecompressor has one value to return.

pkg/kgo/compression.go
twmb: Done.

pkg/kgo/compression.go
twmb: Dropped everything after the comma, and made it a guarantee. The zstd decoder is built with the bound as its ceiling (WithDecoderMaxMemory). Checked in klauspost v1.18.4: DecodeAll rejects a frame whose declared content size is over the remaining ceiling before allocating (decoder.go DecodeAll), errors while decoding when a frame outgrows its declared size or the ceiling (framedec.go runDecoder), and counts every frame of the input against the ceiling. So a lying header cannot get past the bound in either direction. The one hole was DecompressBounded with a bound under the pooled decoder ceiling: that now builds its own decoder at that bound instead of checking after decoding.

pkg/kgo/compression.go
twmb: Done.

pkg/kgo/compression.go
twmb: It meant a per-call bound below the decoder ceiling was only checked after DecodeAll. Gone: that case gets its own decoder now, and the comment is one line.

pkg/kgo/compression.go
twmb: Removed. It was an early reject for the per-call case only; the decoder does the same check itself before allocating, and it does not trust the header: a frame that outgrows its declared size errors, and every frame counts against the ceiling.

pkg/kgo/config.go
twmb: Reworded to the reason from the proposal: "The value must be at least ProducerBatchMaxBytes: an uncompressed batch that size is returned as is, so a smaller bound cannot be enforced." If you meant a different sentence, give me the line.

pkg/kgo/config_test.go
twmb: Deleted.

pkg/kgo/source.go
twmb: Removed the set_offsets line; the log keeps topic, partition, offset, max_bytes.

pkg/kgo/source.go
twmb: Done, "See [MaxDecompressedBatchBytes]."

pkg/kgo/errors.go
twmb: Done.

pkg/kgo/errors.go
twmb: Moved Epoch above NextOffset. On the note: SetOffsets does not want the epoch of the next offset. assignSetMatching stores the pair as cursor{offset: at, lastConsumedEpoch: epoch}, and after normally consuming a batch the cursor holds exactly {offset: lastOffset+1, lastConsumedEpoch: lastRecord.LeaderEpoch} (source.go, processRespPartition). So {Epoch: batch epoch, Offset: NextOffset} is the same state as having consumed the batch, the same pair a commit carries, and it keeps KIP-320 truncation detection; -1 skips without it. I documented it that way. If you would rather steer everyone to -1, say so and I will switch both docs.

pkg/kgo/errors.go
twmb: See the Epoch thread: the doc says "where to SetOffsets to skip the batch" and the Epoch field doc says what -1 versus the batch epoch does. Will change to "(with epoch -1)" if you want that steer.

pkg/kgo/errors.go
twmb: Done: "Epoch is the leader epoch of the batch. Use SetOffsets with {Epoch, NextOffset} to skip this batch."

pkg/kgo/errors.go
twmb: Done: "Epoch is the leader epoch of the batch. Use SetOffsets with {Epoch, NextOffset} to skip this batch."

pkg/kgo/source.go
```diff
@@ -1777,6 +1803,26 @@ func ProcessFetchPartition(o ProcessFetchPartitionOpts, rp *kmsg.FetchResponseTo
 	return fp, o.Offset
 }
 
+// decompress decompresses one batch (or legacy wrapper message) at offset
+// through nextOffset. A batch the decompressor rejects as too large yields
+// an ErrDecompressedTooLarge.
```
twmb: Useless doc

twmb: Removed.

PR #1468, opened 2026-09-18
kgo: decode UTF-8 in regex record readers

ajavanma:
Regex record fields can split valid UTF-8 characters because their RuneReader returns raw bytes with width one. For example, `%v{re[.]}\n` reads `a\n` but fails on `é\n` when the delimiter parser encounters a continuation byte.

Decode whole UTF-8 runes while preserving non-consuming lookahead and byte-based match offsets. Track the returned position separately so truncated input drains correctly, and retain `ErrBufferFull` when the buffer ends inside a valid rune. Public API regressions cover Unicode dot/literal/class matches, adjacent fields and records, fragmented reads, malformed input, and EOF. Direct reader tests cover rune widths, reader errors, retained input, and buffer exhaustion.

Related to #1467, which fixes failed-match error handling in `readRe`. This change fixes rune decoding and byte widths in `reReader`. Both fixes pass the focused tests together, and the separate test file lets the patches merge without conflicts.

Validation:

- New regressions fail on base `85ea1416bff369c2e896e08acfed466d25d05004` and pass with the fix.
- `GOPROXY=off go test ./pkg/kgo -run '^(TestRecord|TestNewRecord|TestReReader)' -count=1` passes.
- `go build ./...`, `go vet ./...`, and `./lint.sh -p .` pass; lint reports zero issues.
- Race tests with Go 1.26.1 and 1.27.1 cannot start on the local ARM64 host: `ThreadSanitizer: unsupported VMA range; Found 47 - Supported 48`.
- Broker integration tests were not run.


twmb:
Good finds, I merged the other PR and pushed one minor change to this one (will squash merge this one)
These will likely wait a week or month or two while I wait for more issues to come in, mostly since essentially nobody uses the record reader so I consider these fairly minor bugs.

PR #1470, opened 2026-09-19
kgo: preserve millisecond timestamps outside the nanosecond range

jakezwang:
Record formatter `%d` and the corresponding record reader convert millisecond timestamps through nanoseconds. Dates outside the `UnixNano` range overflow: formatting 2500-01-01 emits `-1721518473709` instead of `16725225600000`, and reading the correct millisecond value produces a date in 1915.

Use `Time.UnixMilli` and `time.UnixMilli` directly. Tests cover 1600 and 2500, epoch and modern-date controls, default/explicit ASCII and both 64-bit binary byte orders, with literal millisecond/date expectations. ASCII reader behavior for negative numbers is unchanged.

Validation: all `TestRecordFormatter*`, `TestRecordReader*`, and new timestamp tests pass with `-race` on Go 1.26.6 and 1.27.1; `go build ./...`, `go vet ./...`, and `./lint.sh -p .` (golangci-lint 2.13.2) pass. The full `go test -race ./...` run could not complete its Kafka integration tests: `TestPauseIssueOct2023` fails connecting to the required local broker at `127.0.0.1:9092`. No broker was available for that run.


twmb:
Thanks! I'm pushing an update to the test that folds your new test into the existing table driven tests.

